PR #1, opened 2021-12-21
Implement for_each and use it in the example

tpadioleau:


include/for_each.hpp
```diff
@@ -0,0 +1,113 @@
...
+    DiscreteDomainIterator<DDim2> const it2_e = select<DDim2>(domain).end();
+#pragma omp parallel for collapse(2) default(none) shared(it1_b, it1_e, it2_b, it2_e, f)
+    for (DiscreteDomainIterator<DDim1> it1 = it1_b; it1 != it1_e; ++it1) {
+        for (DiscreteDomainIterator<DDim2> it2 = it2_b; it2 != it2_e; ++it2) {
+            f(DiscreteCoordinate<DDim1, DDim2>(*it1, *it2));
+        }
+    }
+}
+
+using default_policy = serial_policy;
```
jbigot, line 85: you could create a constexpr instance for the users to use

include/for_each.hpp
```diff
@@ -0,0 +1,113 @@
+#pragma once
+
+#include <utility>
+
+#include "ddc/chunk_span.hpp"
+#include "ddc/discrete_coordinate.hpp"
+#include "ddc/discrete_domain.hpp"
+
+/// Serial execution
+struct serial_policy
```
jbigot, line 10: you could create a constexpr instance for the users to use

include/for_each.hpp
```diff
@@ -0,0 +1,113 @@
...
+            f(DiscreteCoordinate<DDim1, DDim2>(i1, i2));
+        }
+    }
+}
+
+/** OpenMP parallel execution
+ * - default scheduling
+ * - collapsing in case of tightly nested loops
+ */
+struct omp_policy
```
jbigot, line 47: you could create a constexpr instance for the users to use

jbigot (approved):

include/for_each.hpp
tpadioleau: Done

include/for_each.hpp
tpadioleau: Done

include/for_each.hpp
tpadioleau: I would not expose it for now.

docs/first_steps/heat_equation.cpp
```diff
@@ -1,6 +1,7 @@
 #include <ddc/Chunk>
 #include <ddc/DiscreteCoordinate>
 #include <ddc/DiscreteDomain>
+#include <ddc/ForEach>
```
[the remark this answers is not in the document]
jbigot: `ddc/for_each`

jbigot:
@tpadioleau : we're good, aren't we?

tpadioleau:
I think so, let's merge.

PR #3, opened 2022-02-17
Add some domain related traits

tpadioleau:


include/ddc/traits.hpp
```diff
@@ -0,0 +1,40 @@
...
+#include "non_uniform_discretization.hpp"
+#include "uniform_discretization.hpp"
+
+template <class T>
+struct is_uniform_domain : std::false_type
+{
+};
+
+template <class... DDims>
+struct is_uniform_domain<DiscreteDomain<DDims...>>
```
jbigot, line 14: shouldn't you include something for `DiscreteDomain`?

tpadioleau: I think a declaration is missing in this case.

jbigot: Not sure to understand

tpadioleau: I mean this:
```suggestion

template <class... DDims>
class DiscreteDomain<DDims...>;

struct is_uniform_domain<DiscreteDomain<DDims...>>
```

jbigot (approved):

PR #4, opened 2022-02-17
Update example

tpadioleau:


include/ddc/uniform_discretization.hpp
```diff
@@ -124,6 +124,18 @@ Coordinate<CDim> to_real(DiscreteCoordinate<UniformDiscretization<CDim>> const&
...
 }
 
+template <class CDim>
+Coordinate<CDim> distance_at_left(DiscreteCoordinate<UniformDiscretization<CDim>> i)
+{
+    return discretization<UniformDiscretization<CDim>>().step();
+}
+
+template <class CDim>
+Coordinate<CDim> distance_at_right(DiscreteCoordinate<UniformDiscretization<CDim>> i)
```
jbigot, line 134: maybe add a `max_step_between()`

docs/first_steps/heat_equation.cpp
```diff
@@ -1,6 +1,9 @@
+#include <cmath>
+
 #include <ddc/Chunk>
 #include <ddc/DiscreteCoordinate>
 #include <ddc/DiscreteDomain>
+#include <ddc/NonUniformDiscretization>
```
jbigot, line 6: is it used?

jbigot:
is it still a draft?

include/ddc/uniform_discretization.hpp
tpadioleau: I am not against, but I would wait for the need for it. We could also provide a `distance_between()` algorithm.

tpadioleau: I added these two functions because the PR is about updating the example, not to provide new features on DDC. I can move the functions to the example if needed.

jbigot: works for me like that :+1: no need to change

jbigot (approved):

PR #9, opened 2022-03-09
Add an allocator parameter

tpadioleau:


include/ddc/chunk.hpp
```diff
@@ -60,23 +61,29 @@ class Chunk<ElementType, DiscreteDomain<DDims...>>
...
 public:
     /// Empty Chunk
     Chunk() = default;
 
     /// Construct a Chunk on a domain with uninitialized values
-    explicit Chunk(mdomain_type const& domain)
-        : base_type(new (std::align_val_t(64)) value_type[domain.size()], domain)
+    explicit Chunk(mdomain_type const& domain, Allocator const& allocator = Allocator())
+        : m_allocator(allocator)
+        , base_type(std::allocator_traits<Allocator>::allocate(m_allocator, domain.size()), domain)
```
tpadioleau, line 77: I think we have UB here, the base_type should be initialized before m_allocator but `allocate` captures by lvalue& the Allocator.

include/ddc/chunk.hpp
```diff
@@ -60,23 +61,32 @@ class Chunk<ElementType, DiscreteDomain<DDims...>>
...
     /// Construct a Chunk on a domain with uninitialized values
-    explicit Chunk(mdomain_type const& domain)
-        : base_type(new (std::align_val_t(64)) value_type[domain.size()], domain)
+    explicit Chunk(mdomain_type const& domain, Allocator const& allocator = Allocator())
+        : base_type()
+        , m_allocator(allocator)
     {
+        static_cast<base_type&>(*this) = base_type(
+                std::allocator_traits<Allocator>::allocate(m_allocator, domain.size()),
+                domain);
```
jbigot, line 81: Why not do it in the initializer list like before?

include/ddc/chunk.hpp
```diff
@@ -87,15 +97,17 @@ class Chunk<ElementType, DiscreteDomain<DDims...>>
...
+        , m_allocator(std::move(other.m_allocator))
     {
         other.m_internal_mdspan = internal_mdspan_type();
     }
 
     ~Chunk()
     {
         if (this->m_internal_mdspan.data()) {
-            operator delete[](this->data(), std::align_val_t(64));
+            std::allocator_traits<Allocator>::deallocate(m_allocator, this->data(), this->size());
```
jbigot, line 110: This is a behaviour change, the default allocator doesn't provide aligned memory, does it?

include/ddc/chunk.hpp
```diff
@@ -108,7 +120,9 @@ class Chunk<ElementType, DiscreteDomain<DDims...>>
      */
     Chunk& operator=(Chunk&& other)
```
jbigot, line 121: we should check for self-assignment, shouldn't we?

include/ddc/chunk_span.hpp
```diff
@@ -11,8 +11,9 @@
 
 #include "ddc/chunk_common.hpp"
 #include "ddc/discrete_domain.hpp"
+#include "ddc/memory.hpp"
 
-template <class, class>
+template <class ElementType, class, class Allocator = HostAllocator<ElementType>>
```
jbigot, line 16: Why specify the default allocator here and not in `include/ddc/chunk.hpp` ?

Also, I would rely on `std::allocator` in this CPU-only branch

include/ddc/memory.hpp
```diff
@@ -0,0 +1,53 @@
+#pragma once
```
jbigot, line 1: Don't you think the introduction of the notion of memory-space would benefit from its own branch?

tpadioleau: Why not, we can add a step

include/ddc/chunk_span.hpp
tpadioleau: 1) In fact I thought it had to be specified on the first declaration but it does not seem to be required. I will change update this.

2) Why using `std::allocator` ? We were not relying on this allocator before.

include/ddc/chunk.hpp
tpadioleau: I don't know, what should we expect from `a = std::move(a)` ?

include/ddc/chunk.hpp
tpadioleau: I guess it depends on the default allocator we choose.

include/ddc/chunk.hpp
tpadioleau: Because the `base_type` constructor will be called first, using the uninitialised member `m_allocator`, and then initialize `m_allocator`. I agree it is not pretty.

include/ddc/chunk.hpp
jbigot: a noop, no ?

jbigot: https://cplusplus.github.io/LWG/issue2839 not that clear actually

include/ddc/chunk.hpp
jbigot: Why not just
```
: base_type(std::allocator_traits<Allocator>::allocate(allocator, domain.size()), domain)
, m_allocator(allocator)
```

is it an issue to use the allocator that will be copied instead of the actual member?

jbigot: otherwise, maybe add a protected function dedicated to the allocation in `ChunkCommon`, this could be more explicit than this strange use of the move assignment

jbigot (dismissed):

include/ddc/chunk.hpp
```diff
@@ -20,18 +20,36 @@ class AlignedAllocator
 public:
     using value_type = T;
 
+    using is_always_equal = std::true_type;
```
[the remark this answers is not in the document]
tpadioleau: Indeed it can deduced because the class is empty, I will change that.

include/ddc/chunk.hpp
```diff
@@ -5,14 +5,76 @@
...
+template <class ElementType, class, class Allocator = AlignedAllocator<ElementType, 64>>
 class Chunk;
 
-template <class ElementType, class SupportType>
-inline constexpr bool enable_chunk<Chunk<ElementType, SupportType>> = true;
+template <class ElementType, class SupportType, class Allocator>
+inline constexpr bool enable_chunk<Chunk<ElementType, SupportType, Allocator>> = true;
+
+template <class T, std::size_t N>
+class AlignedAllocator
```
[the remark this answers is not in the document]
tpadioleau: We can expose it if it is useful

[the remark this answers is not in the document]
jbigot: In fact, I wonder if we shouldn't use the system default allocator as default value for `Chunk` and just provide this as an option

PR #12, opened 2022-03-17
Avoid forwarding at each iteration loop

tpadioleau:


jbigot:
Why the change?

jbigot (approved):

tpadioleau:
`mcoords` is being forwarded multiple times, even though it is not critical, it could be captured by value from the functor. The functor is not captured by value anywhere but I don't consider a good pattern to forward multiple times inside a loop.

jbigot:
You're fully right, let's merge right now!

PR #18, opened 2022-03-19
Updated cmake stuff & moved heat_equation to examples/

jbigot:


CMakeLists.txt
```diff
@@ -5,35 +5,119 @@ project(DDC VERSION 0.0.0 LANGUAGES CXX)
...
+
+find_package(OpenMP REQUIRED)
 
-if(DDC_mdspan_DEPENDENCY_POLICY STREQUAL "AUTO")
+
+## mdspan
+
+set(DDC_mdspan_DEPENDENCY_POLICY "AUTO" CACHE STRING "Policy to find the `mdspan` package. Options: ${DDC_DEPENDENCY_POLICIES}")
+set_property(CACHE DDC_mdspan_DEPENDENCY_POLICY PROPERTY STRINGS "${DDC_DEPENDENCY_POLICIES}")
+if("${DDC_mdspan_DEPENDENCY_POLICY}" STREQUAL "AUTO")
```
tpadioleau, line 34: Why using `"${DDC_mdspan_DEPENDENCY_POLICY}"` instead of `DDC_mdspan_DEPENDENCY_POLICY` ?

CMakeLists.txt
```diff
@@ -51,81 +135,60 @@ install(
...
-if(DDC_BUILD_PDI_WRAPPER)
-	if(NOT TARGET PDI::PDI_C)
-		find_package(PDI REQUIRED COMPONENTS C)
-	endif()
 
-	add_library(PDI INTERFACE)
-	target_compile_features(PDI INTERFACE cxx_std_17)
-	target_include_directories(PDI
+if(DDC_BUILD_PDI_WRAPPER)
+	add_library(PDI_wrapper INTERFACE)
```
tpadioleau, line 142: Projects will use `DDC::PDI_wrapper` right ?

Should we find a naming convention ?

CMakeLists.txt
```diff
@@ -51,81 +135,60 @@ install(
...
 		SYSTEM INTERFACE
 			"$<INSTALL_INTERFACE:include>"
 	)
-	target_link_libraries(PDI
+	target_link_libraries(PDI_wrapper
 		INTERFACE
 			DDC::DDC
 			PDI::PDI_C)
-	add_library(DDC::PDI ALIAS PDI)
+	add_library(DDC::PDI ALIAS PDI_wrapper)
```
tpadioleau, line 156:
```suggestion
	add_library(DDC::PDI_wrapper ALIAS PDI_wrapper)
```

CMakeLists.txt
jbigot: the second one is a leftover from before variables in CMake and depending on the name of the variable, can be an issue

tpadioleau: It could be an issue using `${DDC_mdspan_DEPENDENCY_POLICY}` but I guess `DDC_mdspan_DEPENDENCY_POLICY` is okay ? Reading the documentation it says if the variable does not exists it will be equivalent to `"DDC_mdspan_DEPENDENCY_POLICY" STREQUAL "AUTO"`, which seems as wrong as `"" STREQUAL "AUTO"`.

jbigot: I would recommend to never rely on the fact that some command will auto deref variables and always quote var usage so to avoid any ambiguity

CMakeLists.txt
jbigot: I have no issue with `DDC::PDI`, however, just `PDI` can clash with a PDI library if including both

CMakeLists.txt
jbigot: See the other comment

CMakeLists.txt
tpadioleau: Really ? Is there a `PDI` target that is exported ?

jbigot: The one I renamed `PDI_wrapper` used to be named `PDI`, no?

tpadioleau: I mean in the PDI library. Clearly if a project is embedding both PDI and DDC this will break. That is a limitation of embedding libraries as submodules I guess.

tpadioleau: Yes we should recommend to install libraries instead of embedding the world :)

jbigot: yes, use spack!

tpadioleau: I thought there was an agreement on keeping DDC::PDI then ;)

jbigot: Yes, but then I realized that when you find package, the target name will be `DDC::PDI_wrapper` anyway and i throught that was what you were telling me first

jbigot, line 142:
```suggestion
	add_library(PDI_Wrapper INTERFACE)
```

PR #20, opened 2022-03-19
Two minor fixes in tests

jbigot:
* use the correct dim name,
* add missing include

### DIFF
--- a/include/ddc/is_uniform_domain_v
+++ b/include/ddc/is_uniform_domain_v
@@ -1,0 +1,3 @@
+// SPDX-License-Identifier: MIT
+
+#include "ddc/uniform_domain.hpp"
--- a/tests/uniform_domain.cpp
+++ b/tests/uniform_domain.cpp
@@ -1,13 +1,11 @@
 // SPDX-License-Identifier: MIT
 
-#include <type_traits>
+#include <cstddef>
 
-#include <ddc/Coordinate>
-#include <ddc/DiscreteCoordinate>
 #include <ddc/DiscreteDomain>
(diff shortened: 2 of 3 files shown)


include/ddc/is_uniform_domain_v
```diff
@@ -0,0 +1,3 @@
+// SPDX-License-Identifier: MIT
```
tpadioleau, line 1: There is the header `UniformDomain`, if we change that, we also need to change `RectilinearDomain`

tests/uniform_domain.cpp
```diff
@@ -1,13 +1,11 @@
...
 
-#include <type_traits>
+#include <cstddef>
 
-#include <ddc/Coordinate>
-#include <ddc/DiscreteCoordinate>
 #include <ddc/DiscreteDomain>
 #include <ddc/NonUniformDiscretization>
 #include <ddc/UniformDiscretization>
-#include <ddc/UniformDomain>
```
tpadioleau, line 10: I would not change that in this PR then.

include/ddc/is_uniform_domain_v
jbigot: I think we should and we should clarify that the file names should match symbol names

tpadioleau: But what symbols ? some of our public symbols don't have headers, i am afraid it will be difficult to respect in practice.

jbigot: I believe we should do it for all "public" symbols, those we expect our users to manipulate

tpadioleau (approved):

PR #21, opened 2022-03-19
Improved Doxygen doc

jbigot:


include/ddc/chunk.hpp
```diff
@@ -7,12 +7,16 @@
 #include "ddc/chunk_common.hpp"
 #include "ddc/chunk_span.hpp"
 
+/** A chunck
```
tpadioleau, line 10:
```suggestion
/** A chunk
```

include/ddc/chunk.hpp
```diff
@@ -7,12 +7,16 @@
...
+ */
 template <class ElementType, class, class Allocator = std::allocator<ElementType>>
 class Chunk;
 
 template <class ElementType, class SupportType, class Allocator>
 inline constexpr bool enable_chunk<Chunk<ElementType, SupportType, Allocator>> = true;
 
+/** A chunck
```
tpadioleau, line 18:
```suggestion
/** A chunk
```

include/ddc/for_each.hpp
```diff
@@ -13,8 +13,8 @@ struct serial_policy
 {
 };
 
-/** iterates over a 1d domain
- * @param[in] domain the 1d domain to iterate
+/** iterates over a nD domain
```
tpadioleau, line 16: I guess we don't need to change that in this PR.

include/ddc/transform_reduce.hpp
```diff
@@ -29,18 +33,22 @@ inline T transform_reduce(
...
+ * @param[in] reduce a binary FunctionObject that will be applied in unspecified order to the
+ *            results of transform, the results of other reduce and init. 
+ * @param[in] transform a unary FunctionObject that will be applied to each element of the input
+ *            range. The return type must be acceptable as input to reduce
  */
-template <class DDim1, class DDim2, class T, class BinaryReductionOp, class BinaryTransformOp>
+template <class DDim1, class DDim2, class T, class BinaryReductionOp, class UnaryTransformOp>
 inline T transform_reduce(
-        serial_policy,
+        serial_policy policy,
```
tpadioleau, line 47: We will have an "unused warning", shall we move the documentation to a declaration without definition ?

include/ddc/chunk.hpp
jbigot: That was not intended to be here at all, I even think another PR removes this comment

include/ddc/chunk.hpp
```diff
@@ -7,12 +7,16 @@
...
+ */
```
jbigot, line 19:
```suggestion
```

include/ddc/transform_reduce.hpp
jbigot: or use [`maybe_unused`](https://en.cppreference.com/w/cpp/language/attributes/maybe_unused)

include/ddc/for_each.hpp
jbigot: true

jbigot:
```suggestion
/** iterates over a 1D domain
```

include/ddc/transform_reduce.hpp
jbigot:
```suggestion
        [[maybe_unused]] serial_policy policy,
```

tpadioleau: Why not if it does not appear on the documentation

tpadioleau (approved):

PR #23, opened 2022-03-19
Implement for_each recursively and use it in deepcopy

jbigot:

PR #24, opened 2022-03-19
Support iterating over extents with for_each

jbigot:
This should come after #23

### DIFF
--- a/include/ddc/for_each.hpp
+++ b/include/ddc/for_each.hpp
@@ -118,6 +118,12 @@ inline void for_each(DiscreteDomain<DDims...> const& domain, Functor&& f) noexce
     for_each(default_policy(), domain, std::forward<Functor>(f));
 }
 
+template <class... DDims, class Functor>
+inline void for_each(DiscreteVector<DDims...> const& domain, Functor&& f) noexcept
+{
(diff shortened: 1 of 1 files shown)


include/ddc/for_each.hpp
```diff
@@ -120,6 +120,12 @@ inline void for_each(DiscreteDomain<DDims...> const& domain, Functor&& f) noexce
     for_each(default_policy(), domain, std::forward<Functor>(f));
 }
 
+template <class... DDims, class Functor>
+inline void for_each(DiscreteVector<DDims...> const& domain, Functor&& f) noexcept
```
tpadioleau, line 124: Why do we need this `for_each` version ?

tpadioleau: I am skeptical, is it for the periodic boundary conditions ? Could we imagine a remapping instead ?

tpadioleau: Maybe call it differently then ? I think the matching of the domains can help debugging.

jbigot: We could, but I figured that the use of a different type (`DiscreteVector<>` instead of `DiscreteDomain<>`) was enough

jbigot:
merged

PR #27, opened 2022-03-19
Do not use a dedicated uniform domain file

jbigot:
This must come after #20


tpadioleau:
In fact I was thinking of moving some functions of `uniform_discretization.hpp` into `uniform_domain.hpp`. The latter would define functions that rely on a `DiscreteDomain` like `rmin`, `rmax`, `rlength`...

jbigot:
If you prefer, that works for me

tpadioleau:
The drawback is that it breaks the rule of one symbol per header then.

PR #28, opened 2022-03-19
Expose vector & domain types in discretization

jbigot:


include/ddc/uniform_discretization.hpp
```diff
@@ -51,6 +55,8 @@ class UniformDiscretization
      * @param a the coordinate of a first real point (will have mesh coordinate 0)
      * @param b the coordinate of the second real point (will have mesh coordinate `n-1`)
      * @param n the number of points to map the segment \f$[a, b]\f$ including a & b
+     * 
+     * @deprecated use the version accepting a vector for n instead
      */
     constexpr UniformDiscretization(rcoord_type a, rcoord_type b, std::size_t n)
```
tpadioleau, line 61: Should we mark it `[[deprecated]]` ?

tpadioleau (dismissed):

include/ddc/uniform_discretization.hpp
jbigot: As long as our only user is Voice, I even think we should remove it ASAP

include/ddc/discrete_domain.hpp
```diff
@@ -22,6 +22,8 @@ class DiscreteDomain
     template <class...>
     friend class DiscreteDomain;
 
+    friend std::tuple_element_t<0, std::tuple<DDims...>>;
+
```
jbigot, line 26:
```suggestion
```

include/ddc/discrete_domain.hpp
```diff
@@ -59,6 +61,8 @@ class DiscreteDomain
 
     /** Construct a DiscreteDomain starting from (0, ..., 0) with size points.
      * @param size the number of points in each dimension
+     * 
+     * @deprecated use the version with explicit lower bound instead
      */
     constexpr DiscreteDomain(mlength_type const& size)
```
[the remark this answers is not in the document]
jbigot:
```suggestion
    [[deprecated]] constexpr DiscreteDomain(mlength_type const& size)
```

PR #29, opened 2022-03-19
Offer initialization function generating base domains for uniform dis…

jbigot:
…cretization

This must come after #28


include/ddc/discrete_domain.hpp
```diff
@@ -22,6 +22,8 @@ class DiscreteDomain
     template <class...>
     friend class DiscreteDomain;
 
+    friend std::tuple_element_t<0, std::tuple<DDims...>>;
+
```
jbigot, line 26:
```suggestion
```

PR #36, opened 2022-03-23
  Implements a recursive for_each with OpenMP policy

tpadioleau:


include/ddc/for_each.hpp
```diff
@@ -38,69 +57,55 @@ struct serial_policy
...
-    DiscreteDomainIterator<DDim2> const it2_b = select<DDim2>(domain).begin();
-    DiscreteDomainIterator<DDim2> const it2_e = select<DDim2>(domain).end();
-#pragma omp parallel for collapse(2) default(none) shared(it1_b, it1_e, it2_b, it2_e, f)
-    for (DiscreteDomainIterator<DDim1> it1 = it1_b; it1 != it1_e; ++it1) {
-        for (DiscreteDomainIterator<DDim2> it2 = it2_b; it2 != it2_e; ++it2) {
-            f(DiscreteCoordinate<DDim1, DDim2>(*it1, *it2));
-        }
-    }
+    detail::for_each_omp(
+            DiscreteVector<DDims...> {detail::type_constant_v<DDims, std::ptrdiff_t, 0>...},
```
jbigot, line 106: I committed that without asking you. Do you see any better way to implement it?

jbigot (approved):

include/ddc/for_each.hpp
tpadioleau: You mean the null vector ?

jbigot: yep, or more exactly repeat a constant n times

jbigot: there's no `std::repeat_v<sizeof...(DDims), 0>` or equivalent ?

tpadioleau: I have an other way, better will be quite subjective:
```
{std::integral_constant<std::conditional_t<true, std::ptrdiff_t, std::ptrdiff_t>, 0>::value...}
```

tpadioleau: I don't think so

PR #52, opened 2022-03-29
Use a single header

tpadioleau:


tpadioleau:
Shall we have an other header for PDI, a ddc/pdi.hpp ? Or just use a ifdef inside ddc/ddc.hpp ?

PR #53, opened 2022-03-29
Introduce DiscreteElement class

tpadioleau:


tpadioleau:
@jbigot what do you think of renaming TaggedVector into DiscreteVector (thus removing the element parameter template) ? And replacing DiscreteCoordinate by DiscreteElement (no element template either ?) ?

jbigot:
Isn't there a lot of implementation in common that could be shared?

tpadioleau:
What implementation are you thinking of ?

tpadioleau:
Let `a, b` be a discrete element, `u, v` discrete vectors and `x` an integral, we need to define the following operators

- internal operations `+u`, `-u`, `x * u`, `u+v`, `u-v`,
- right external operations `a+u`, `a-u`
- external operation `a-b`

I agree that it would make sense to allow implicit conversion of an integral to a 1-d discrete vector

jbigot (approved):

PR #61, opened 2022-06-07
Make some ctors implicit

jbigot:
it's nice to be able to use `{vx,vy}` where a `vxy` is expected and I don't see any danger with it


tpadioleau (approved):
I do not see any harm either.

PR #67, opened 2022-06-28
Use a single location for local space definition

jbigot:
@tpadioleau : How can I test that easily with an nvcc & hip compiler, could we add at least nvcc to our test infra?


jbigot:
@tpadioleau : I'd wait for PR !68 to be merged before this, but then this is ready from my point of view and your can merge if you feel like it

tpadioleau (approved):

PR #70, opened 2022-07-05
Some cleaning regarding the Kokkos layer

tpadioleau:


include/ddc/detail/kokkos.hpp
```diff
@@ -8,55 +8,101 @@
...
+struct type_holder
+{
+    using type = T;
+};
 
 template <class KokkosLP>
-using mdspan_layout_t = typename mdspan_layout<KokkosLP>::type;
+struct kokkos_to_mdspan_layout
+{
+    static_assert(
```
jbigot, line 20: is this better than just not defining the struct?

jbigot (approved):

include/ddc/detail/kokkos.hpp
tpadioleau: It just helps to diagnose the problem with an error message.

jbigot: :+1:

PR #72, opened 2022-07-07
Implement possibility to have empty discrete domains

tpadioleau:


include/ddc/discrete_domain.hpp
```diff
@@ -29,9 +29,9 @@ class DiscreteDomain
     using mlength_type = DiscreteVector<DDims...>;
 
 private:
-    DiscreteElement<DDims...> m_lbound;
+    DiscreteElement<DDims...> m_element_begin;
 
-    DiscreteElement<DDims...> m_ubound;
+    DiscreteElement<DDims...> m_element_end;
```
jbigot, line 34: This adds the constraint that such a past-the-end element must be valid to construct

tpadioleau: In my opinion a user can construct any valid `DiscreteElement<DDims...>` like one would do for a raw integer, it is the responsibility of the user to know if it makes sense to use it or not.

PR #81, opened 2022-10-25
Remove rank in discrete spaces

tpadioleau:


jbigot:
Ok, why not, but why?

jbigot (approved):
Why not

tpadioleau:
Well it is not used and not well defined. I would advocate to remove it. If we find it useful in the future we can totally reintroduce it.

jbigot:
Not used, I agree. Not well defined, not so much. The rank is the number of discrete dimensions introduced by each discretization, *i.e.* whether an order is defined over the discrete elements (1D) or whether the discrete elements are a vector of discrete elements for which an order is defined

tpadioleau:
Ok why not but it was not defined until now. I understand the purpose of defining the rank for a chunk/chunkspan because it tells you how many indices you need to give. However in the case of discrete spaces the purpose is not clear. For example if we want to reintroduce the `UniformMesh` discretization, we would introduce the `node` and `cell` discrete spaces thus making a rank 2, even if some functions are of rank 0, 1 and 2.

PR #101, opened 2022-11-19
Remove line raising unused warning

EmilyBourne:
`uniform_point_sampling.hpp` is raising the following warning:
```
/home/emily/Code/VOICE/voicexx/vendor/ddc/include/ddc/uniform_point_sampling.hpp: In static member function ‘static std::tuple<ddc::UniformPointSampling<CDim>::Impl<Kokkos::HostSpace>, ddc::DiscreteDomain<ddc::UniformPointSampling<CDim> >, ddc::DiscreteDomain<ddc::UniformPointSampling<CDim> >, ddc::DiscreteDomain<ddc::UniformPointSampling<CDim> >, ddc::DiscreteDomain<ddc::UniformPointSampling<CDim> > > ddc::UniformPointSampling<CDim>::init_ghosted(ddc::UniformPointSampling<CDim>::continuous_element_type, ddc::UniformPointSampling<CDim>::continuous_element_type, ddc::UniformPointSampling<CDim>::discrete_vector_type, ddc::UniformPointSampling<CDim>::discrete_vector_type, ddc::UniformPointSampling<CDim>::discrete_vector_type)’:
/home/emily/Code/VOICE/voicexx/vendor/ddc/include/ddc/uniform_point_sampling.hpp:177:15: warning: typedef ‘using continuous_element_type = using continuous_element_type = ddc::Coordinate<CDim>’ locally defined but not used [-Wunused-local-typedefs]
  177 |         using continuous_element_type = continuous_element_type;
```
This PR removes the offending line

PR #103, opened 2022-11-21
Allow addition of non-exact matching objects

### LINKED ISSUE #102: Allow addition of non-exact matching objects
I think it would be useful to allow expressions such as:
```cpp
ddc::DiscreteElement<Dim1, Dim2> old_pos = ...;
ddc::DiscreteElement<Dim1, Dim2> new_pos = old_pos + ddc::DiscreteVector<Dim2>(3);
```
Currently it is only possible to do:
```cpp
ddc::DiscreteElement<Dim1, Dim2> old_pos = ...;
ddc::DiscreteElement<Dim1, Dim2> new_pos = old_pos + ddc::DiscreteVector<Dim1, Dim2>(3, 0);
```

This addition would be useful in cases where we know that `Dim2` is one of the dimensions but we don't want to specialise the function for all possible values of `Dim1` (present or not present)

EmilyBourne:
Fix #102 and add corresponding tests


EmilyBourne:
@jbigot Do you have any idea why docker doesn't like my fork?

jbigot:
> @jbigot Do you have any idea why docker doesn't like my fork?

The image is generated alright, but you are not allowed to push the generated image to the "organization package registry"

What's not clear to me is that it doesn't seem you changed anything to the image generation in the diff. Did I miss something?

jbigot:
> > The image is generated alright, but you are not allowed to push the generated image to the "organization package registry"
> > What's not clear to me is that it doesn't seem you changed anything to the image generation in the diff. Did I miss something?
> 
> I don't think so. The only difference to any other PR is that this is a fork (in my previous PR #101 I did not create a branch. On other repos I have seen this behave differently, maybe because the base branch is different in this case)

Indeed, since this is a different repo, it doesn't have the right to push to the `Maison-de-la-Simulation` organization. Do you think you could make a branch instead?

.github/workflows/pages.yml
```diff
@@ -20,7 +20,7 @@ jobs:
           docker/doxygen
         docker push ghcr.io/maison-de-la-simulation/ddc/doxygen:${GITHUB_SHA:0:7}
     - name: Publish
-      if: ${{ github.event_name == 'push' && github.ref_name == 'main' }}
+      if: ${{ github.event_name == 'push' && github.ref_name == 'main' && github.repository == 'Maison-de-la-Simulation/ddc' }}
       run: |
         docker push ghcr.io/maison-de-la-simulation/ddc/doxygen
```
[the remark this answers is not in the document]
jbigot: the first push with the commit sha1 as tag is used in the following steps, the one without tag (hence the `latest` auto tag) represents the top of `main` and is only pushed in case of merge into main

.github/workflows/tests.yml
```diff
@@ -30,7 +30,10 @@ jobs:
           -t ghcr.io/maison-de-la-simulation/ddc/test_env \
           -t ghcr.io/maison-de-la-simulation/ddc/test_env:${GITHUB_SHA:0:7} \
           docker/test_env
-        docker push ghcr.io/maison-de-la-simulation/ddc/test_env:${GITHUB_SHA:0:7}
+        if [ $(git diff main..HEAD docker/test_env | wc -l) -ne 0 ]
+        then
+          docker push ghcr.io/maison-de-la-simulation/ddc/test_env:${GITHUB_SHA:0:7}
```
jbigot, line 35: if you don't push it, the following jobs (`test-Debug`, `test-Release`, `test-hip`) won't be able to use it

.github/workflows/tests.yml
```diff
@@ -49,7 +52,10 @@ jobs:
           -t ghcr.io/maison-de-la-simulation/ddc/jammy_env \
           -t ghcr.io/maison-de-la-simulation/ddc/jammy_env:${GITHUB_SHA:0:7} \
           docker/jammy_env
-        docker push ghcr.io/maison-de-la-simulation/ddc/jammy_env:${GITHUB_SHA:0:7}
+        if [ $(git diff main..HEAD docker/jammy_env | wc -l) -ne 0 ]
+        then
+          docker push ghcr.io/maison-de-la-simulation/ddc/jammy_env:${GITHUB_SHA:0:7}
```
jbigot, line 57: if you don't push it, the following job (`test-cuda`) won't be able to use it

jbigot (changes requested):

.github/workflows/tests.yml
EmilyBourne: :+1: I spotted that now. I have opened a PR (#105 ) using a branch but I thought I'd play around and see if I could get this PR to work so you can accept forks in the future

jbigot: yep, I've seen your branch. Maybe you can look at artifacts to transfer data from one job to another. without requiring access on the mother organization. I'm sure there is a better way to do it than what we do now, but I haven't taken the time to look into it.

Maybe you can create an issue about this specific problem

EmilyBourne: I have created the issue (#107). I have some experience with artifacts so I have done that step (#108). But I am not familiar enough with docker to know how to use those artifacts

jbigot: that would be:
```
Usage:  docker export [OPTIONS] CONTAINER

Export a container's filesystem as a tar archive

Options:
  -o, --output string   Write to a file, instead of STDOUT
```

to create the artifact, then
```
Usage:  docker import [OPTIONS] file|URL|- [REPOSITORY[:TAG]]

Import the contents from a tarball to create a filesystem image

Options:
  -c, --change list       Apply Dockerfile instruction to the created image
  -m, --message string    Set commit message for imported image
      --platform string   Set platform if server is multi-platform capable
```

to use it I guess

PR #108, opened 2022-12-12
Use artifacts for docker images

EmilyBourne:
Use artifacts for docker images to ensure forks can contribute.


.github/workflows/pages.yml
```diff
@@ -17,6 +17,10 @@ jobs:
           -t ghcr.io/maison-de-la-simulation/ddc/doxygen \
           -t ghcr.io/maison-de-la-simulation/ddc/doxygen:${GITHUB_SHA:0:7} \
           docker/doxygen
+        docker export docker/doxygen
```
jbigot, line 20: I believe you need to specify `-o`

```
docker export -o doxygen.tar docker/doxygen
```

EmilyBourne: `Error response from daemon: No such container: docker/doxygen`
Maybe I need to run first? Or I don't have the right name?

jbigot: the generated image has two names (you can use either) `ghcr.io/maison-de-la-simulation/ddc/doxygen` or `ghcr.io/maison-de-la-simulation/ddc/doxygen:${GITHUB_SHA:0:7}`

jbigot: `docker/doxygen` is a directory

EmilyBourne: `import`/`export` handle containers not images. It seems we need `load`/`save` to handle images.

I now seem to have found the right commands to handle the docker images via artifacts, however `test_env` gives me the following error:
```
write /dev/stdout: no space left on device
```

jbigot: indeed, we're a bit tight on disk space in the VMs...
We already remove some non critical files of the VM to save space

jbigot: Cf. the use of action `jlumbroso/free-disk-space`

EmilyBourne: Hmm, so then what do you recommend? I'm not sure how we can get round this.

How is the VM set up? Would it help if we organised the tests so that `jammy_env` and `test_env` are not created simultaneously?

jbigot: one VM is created for each Job, so it's better to keep them in separate jobs

you can try to copy the `jlumbroso/free-disk-space` action from `docker-build-test_env` to `docker-build-jammy_env` so as to also make more space in this case

jbigot: but since your issue is with `test_env`, all I can think of, is to be more aggressive in the parameters of `jlumbroso/free-disk-space` to free more space or to setup our own runner with more space available than those offered by Microsoft

EmilyBourne: I can try that.

To be clear I wasn't suggesting that we put these into 1 job. I just wondered if the VMs are on the same physical machine? And if so then are they sharing memory? Because if so adding a dependency between the two jobs would prevent them running simultaneously which might ensure the RAM is available.

EmilyBourne: `test_hip` was passing when the test_env was failing but is now failing with no space left on device. This makes me think that running them in parallel on the same vm might be a problem.

jbigot: > I can try that.
> 
> To be clear I wasn't suggesting that we put these into 1 job. I just wondered if the VMs are on the same physical machine?  And if so then are they sharing memory? 

Sorry, I misunderstood. The answer is we have no control over that.

> Because if so adding a dependency between the two jobs would prevent them running simultaneously which might ensure the RAM is available.

Just to be clear, the memory issue is not in RAM, but on disk, so running in parallel or not won't change the space available in the virtual disk image.

jbigot: > `test_hip` was passing when the test_env was failing but is now failing with no space left on device. This makes me think that running them in parallel on the same vm might be a problem.

It's always random. We are ensured to get [14 GB of SSD space](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources) by github (which is not enough) and we pray to get more (which we usually get) but this might not always be the case.

jbigot: you can setup `jlumbroso/free-disk-space` in `test_hip` too

.github/workflows/pages.yml
```diff
@@ -1,7 +1,6 @@
 name: pages
 on:
   push:
-    branches: [ main ]
```
jbigot, line 4: I'd keep it

jbigot (approved):

.github/workflows/pages.yml
EmilyBourne: I think it must have been changed for debugging purposes. I've reverted the change

jbigot (dismissed):
perfect, thanks !

.github/workflows/pages.yml
jbigot: `docker-build-test_env` seems to still sometimes fail with lack of space

jbigot: I don't see a good solution that will work all the time.

My proposition is to introduce a condition in the code to test from which repository the source branch is:
* when on the original repository, use the initial solution using docker push,
* when on a different directory, use the version with artifact.

That way, we shouldn't have random failure in case that already worked, but new cases when pulling from external repo might work (maybe after a few tries due to size limitations)

tpadioleau:
Shall it be merged ?

.github/workflows/pages.yml
EmilyBourne: Fixed

jbigot:
Alrighty ! Thank you very much, @EmilyBourne !

jbigot (approved):

PR #110, opened 2022-12-20
Create multi-D `Coordinate` from multi-D `DiscreteElement`

EmilyBourne:


EmilyBourne:
I suspect that this file is not the correct location for this function, but I don't know where would be preferable.

tpadioleau:
then let's merge

tpadioleau:
we can move this free function into its own file `coordinate.hpp`

jbigot:
I didn't look at the implementation in detail on my side

If it's ok for you, let's go, indeed!

PR #118, opened 2023-04-20
Add tag support v1

tpadioleau:


jbigot (dismissed):
you like this better than the following ?
```
class DDimX {
  using discretization = ddc::UniformPointSampling<X>;
};
```

tpadioleau:
It seemed easier to implement but I will also implement the other solution and compare

tpadioleau:
If we have
```
class DDimX {
  using discretization = ddc::UniformPointSampling<X>;
};
```
I do not see how the discretization template by the contiguous dimension can implement functions depending on the name of the discretization like:
```
template <class CDim>
template <class MemorySpace>
struct UniformPointSampling<CDim>::Impl<MemorySpace>
{
    /// @brief Convert a mesh index into a position in `CDim`
    constexpr Coordinate<CDim> coordinate(DiscreteElement<DiscretizationName> const& icoord) const noexcept
    {
       return m_origin + Coordinate<CDim>(icoord.uid()) * m_step;
    }
}
```
I guess the name of the discretization must be injected into the discretization.

tpadioleau:
choosing #121

PR #123, opened 2023-04-25
Update Kokkos to version 4.0.00

tpadioleau:


.github/workflows/pages.yml
```diff
@@ -35,7 +35,8 @@ jobs:
     - name: Build site
       run: |
         cat<<-EOF > run.sh
-          set -xe 
+          set -xe
+          git config --global --add safe.directory /src/vendor/kokkos
```
tpadioleau, line 39: I don't know if this can be improved

PR #125, opened 2023-04-27
Simplify implementation

tpadioleau:


tpadioleau:
@EmilyBourne What do you think of this implementation ?

At some point I think we will an implementation where `DiscreteElement<X, Y> + DiscreteElement<Z, Y>` returns a `DiscreteElement<X, Y, Z>`

EmilyBourne:
> @EmilyBourne What do you think of this implementation ?
> 
> At some point I think we will an implementation where `DiscreteElement<X, Y> + DiscreteElement<Z, Y>` returns a `DiscreteElement<X, Y, Z>`

Am I right that it is the same, it is just relying on the incrementation operators to decrease duplication?

tpadioleau:
Yes the features remain the same

PR #138, opened 2023-06-20
add tests in C++ 17, 20, 23 & factorize test.yml

jbigot:


include/ddc/chunk_span.hpp
```diff
@@ -38,14 +38,11 @@ class ChunkSpan<ElementType, DiscreteDomain<DDims...>, LayoutStridedPolicy, Memo
...
-    /// the raw mdspan underlying this, with the same indexing (0 might no be dereferenceable)
-    using internal_mdspan_type = std::experimental::mdspan<
-            ElementType,
-            std::experimental::dextents<sizeof...(DDims)>,
-            std::experimental::layout_stride>;
-
     using base_type = ChunkCommon<ElementType, DiscreteDomain<DDims...>, LayoutStridedPolicy>;
 
+    /// the raw mdspan underlying this, with the same indexing (0 might no be dereferenceable)
+    using typename base_type::internal_mdspan_type;
```
tpadioleau, line 44: Was it needed ?

docker/test_env/Dockerfile
```diff
@@ -18,26 +18,26 @@ RUN chmod +x /bin/bash_run \
     ca-certificates \
     wget \
  && echo 'deb [arch=amd64] https://repo.radeon.com/amdgpu/latest/ubuntu focal main' > /etc/apt/sources.list.d/amdgpu.list \
- && echo 'deb [arch=amd64] https://repo.radeon.com/rocm/apt/debian/ ubuntu main' > /etc/apt/sources.list.d/rocm.list \
+ && echo 'deb [arch=amd64] https://repo.radeon.com/rocm/apt/latest focal main' > /etc/apt/sources.list.d/rocm.list \
```
tpadioleau, line 21: Is it better ?

.github/workflows/tests.yml
```diff
@@ -1,290 +1,152 @@
...
     - name: Identify repository
       id: identify_repo
       run: |
         echo "in_base_repo=${{ (github.event == 'push' && github.repository == 'Maison-de-la-Simulation/ddc') || github.event.pull_request.head.repo.full_name == 'Maison-de-la-Simulation/ddc' }}" >> "$GITHUB_OUTPUT"
         echo "$GITHUB_OUTPUT"
 
-  docker-build-test_env:
+  docker-build-env:
+    strategy:
+      matrix: { image: ['test_env', 'jammy_env'] }
```
tpadioleau, line 29: Shouldn't we rename `test_env` by `focal_env` ?

tpadioleau (approved):
That seems simpler. Can you clarify that now both images habe the same role ?

include/ddc/chunk_span.hpp
jbigot: Some things didn't compile, not sure this was the fix in the end. In any case I believe this is better than a copy & paste

docker/test_env/Dockerfile
jbigot: it works

.github/workflows/tests.yml
jbigot: yes, but should we keep a focal based image?

.github/workflows/tests.yml
```diff
@@ -1,290 +1,152 @@
...
-        docker run -v ${PWD}:/src:ro ${{ needs.docker-build-test_env.outputs.test_env_image }} bash /src/run.sh
-
-  test-Release:
+  test-build-and-run:
+    strategy:
+      fail-fast: false
+      matrix:
+        cxx_version:  ['17', '20', '23']
+        cmake_build_type: ['Debug', 'Release']
+        kokkos_backend: ['CPU-gnu', 'CPU-clang', 'GPU-HIP', 'GPU-Cuda']
```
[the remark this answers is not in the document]
jbigot: OK, that's what I had first, I read it Cuda someplace. Will change

jbigot:
> That seems simpler. Can you clarify that now both images habe the same role ?

what do you mean by that ? They used to and still have similar roles

.github/workflows/tests.yml
tpadioleau: I guess it provides different compiler versions ? But this may be an indirect advantage. Maybe in the future we should have only one image but with all the desired compilers ?

jbigot: I'd rather remove test_env, switch to jamy as minimum version and add a lunar based image for "latest" compilers

tpadioleau:
I thought one of them was to ensure it works for local development and the other one for GPUs ?
I am not sure anybody is still developing using the distrib packages now.

.github/workflows/tests.yml
tpadioleau: I do not see any problem with this.

.github/workflows/tests.yml
```diff
@@ -1,290 +1,152 @@
...
-          export CC=/opt/rocm/hip/bin/hipcc
-          export CXX=/opt/rocm/hip/bin/hipcc
-          git config --global --add safe.directory /src/vendor/kokkos
-          cmake -DCMAKE_CXX_FLAGS="-Wall -Wextra -pedantic -Werror=vla" -DBUILD_BENCHMARKS=ON -DKokkos_ENABLE_HIP=ON -DKokkos_ARCH_VEGA90A=ON -DCMAKE_PREFIX_PATH=/opt/rocm -DKokkos_ENABLE_DEPRECATED_CODE_3=OFF -DKokkos_ENABLE_DEPRECATED_CODE_4=OFF -DKokkos_ENABLE_DEPRECATION_WARNINGS=OFF /src
-          make -j 2 VERBOSE=1
+          git config --global --add safe.directory '*'
+          case "${{matrix.kokkos_backend}}" in
+          'GPU-CUDA')
+            export CC=gcc-10
+            export CXX=g++-10
```
tpadioleau, line 112: You prefer not to specify the `nvcc_wrapper` in the CUDA case btw ?

jbigot: I didn't change it, that's what we did before

.github/workflows/tests.yml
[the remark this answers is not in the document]
jbigot: done

.github/workflows/tests.yml
jbigot: I'd say we can change it if better, but I would do that in a separate PR

tpadioleau: Indeed I am surprised. Well let's keep it, just to let you know that Kokkos will you use `nvcc_wrapper` in the end to compile the project if `Kokkos_ENABLE_CUDA` is `ON`

jbigot:
> > > I thought one of them was to ensure it works for local development and the other one for GPUs ? I am not sure anybody is still developing using the distrib packages now.
> > 
> > 
> > ```
> > * we had focal because we said that was the oldest ubuntu version we supported and hence we tested on it
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > We need to define the minimum version of the libraries that we support as part of our contract with the users. Saying «whatever version is available on _this_ version of Ubuntu» is a good way to do it.
> > ```
> > * cuda was not available (or not recent enough) on focal, so we tested cuda on jammy
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I propose to switch to : «we test on the latest version of ubuntu and on the oldest we claim to support» And I propose to switch to jammy as «the oldest version of Ubuntu we claim to support»
> 
> Yes that works for me.

I would do that in a separate PR. What do you say?

tpadioleau:
Yes

jbigot:
In that case, as soon as you OK it, I think we can merge

tpadioleau (approved):

PR #140, opened 2023-06-21
No implicit fallthrough

jbigot:


tpadioleau:
It could be a good idea to add these checks in Gysela

tpadioleau (approved):
Seems good to me

PR #143, opened 2023-06-22
Add junit report

jbigot:


jbigot:
too bad the report only shows in the run page such as https://github.com/Maison-de-la-Simulation/ddc/actions/runs/5347548794 I hoped they would show here... anyway, this is ready from my point of view

tpadioleau:
Cool that seems good to me! (up to my little knowledge about this)

tpadioleau (approved):

PR #147, opened 2023-06-29
Experimental templatization gys changes

blegouix:
ddc changes required for gysela templatization


include/ddc/coordinate.hpp
```diff
@@ -15,7 +15,14 @@ using CoordinateElement = double;
...
  */
+
 template <class... CDims>
 using Coordinate = detail::TaggedVector<CoordinateElement, CDims...>;
 
+template <class T>
+struct CoordinateOf;
+
+template <class T>
+using coordinate_of = typename CoordinateOf<T>::type;
```
jbigot, line 26: don't we use the `_t` convention for these?

include/ddc/discrete_domain.hpp
```diff
@@ -365,6 +365,24 @@ using convert_type_seq_to_discrete_domain = typename ConvertTypeSeqToDiscreteDom
...
+};
+
+template <typename DDom1, typename DDom2, typename... Tail>
+struct CartesianProd<DDom1, DDom2, Tail...>
+{
+    using type = typename CartesianProd<typename CartesianProd<DDom1, DDom2>::type, Tail...>::type;
+};
+
+template <typename... DDom>
+using cartesian_prod = typename CartesianProd<DDom...>::type;
```
jbigot, line 384: same about `_t`

jbigot: in what case do we need this?

include/ddc/coordinate.hpp
blegouix: We can, I don't like it much because using struct is an internal requirement due to specialization limitations so I consider "_t" should not be visible to the user (and is a bit confusing), but we can.

include/ddc/discrete_domain.hpp
blegouix: In Gysela
using FdistribuDDom = ddc::cartesian_prod<SpDDom, SpatialDDom, VelocityDDom>;

jbigot: sounds nice. you can add the handling of the degenerate case `ddc::cartesian_prod<SingleDom>` while you're at it

include/ddc/coordinate.hpp
tpadioleau: I agree with @jbigot, we tried to follow a convention. I don't think it is a good idea to introduce a new style in a MR about a new feature.

tpadioleau (approved):

PR #152, opened 2023-06-29
fft normalization test

blegouix:


tests/fft.hpp
```diff
@@ -5,6 +5,8 @@
 #include <ddc/ddc.hpp>
 #include <ddc/kernels/fft.hpp>
 
+#include "ddc/discrete_element.hpp"
+
```
tpadioleau, line 9: Not needed
```suggestion
```

tests/fft.hpp
```diff
@@ -214,3 +216,87 @@ static void test_fft()
...
+#if 0
+    std::cout << "\n input:\n";
+    ddc::for_each(
+            ddc::policies::serial_host,
+            ddc::get_domain<DDim<X>>(f_host),
+            [=](DElem<DDim<X>> const e) {
+                (std::cout << coordinate(ddc::select<DDim<X>>(e)))
+                        << "->" << f_host(e) << ", ";
+            });
+#endif
```
tpadioleau, line 260: To be removed

tests/fft.hpp
```diff
@@ -214,3 +216,87 @@ static void test_fft()
...
+    std::cout << "\n output:\n";
+    ddc::for_each(
+            ddc::policies::serial_host,
+            ddc::get_domain<DFDim<ddc::Fourier<X>>>(Ff_host),
+            [=](DElem<DFDim<ddc::Fourier<X>>> const e) {
+                (std::cout << coordinate(ddc::select<DFDim<ddc::Fourier<X>>>(e)))
+                        << "->" << Kokkos::abs(Ff_host(e)) << " "
+                        << ", ";
+            });
+#endif
```
tpadioleau, line 276: To be removed

tests/fft.hpp
```diff
@@ -214,3 +216,87 @@ static void test_fft()
...
+
+    double Ff0_expected;
+    switch (norm) {
+    case ddc::FFT_Normalization::OFF:
+        Ff0_expected = 2;
```
tpadioleau, line 281: Is it 2 because you have 2 points in your mesh ?

tests/fft.hpp
```diff
@@ -214,3 +216,87 @@ static void test_fft()
...
+        break;
+    case ddc::FFT_Normalization::FORWARD:
+        Ff0_expected = 1;
+        break;
+    case ddc::FFT_Normalization::BACKWARD:
+        Ff0_expected = 2;
+        break;
```
tpadioleau, line 288: Shall we also test `ifft` because the behavior shall be reversed ?

tpadioleau (changes requested):
Seems globally good!

tests/fft.hpp
```diff
@@ -214,3 +216,87 @@ static void test_fft()
...
+    case ddc::FFT_Normalization::FULL:
+        Ff0_expected = 1 / Kokkos::sqrt(2 * Kokkos::numbers::pi);
+        break;
+    }
+
+    double criterion = Kokkos::abs(Ff(Ff.domain().front()) - Ff0_expected);
+
+    std::cout << "\n Distance between analytical prediction and numerical result : " << criterion;
+    double epsilon = 1e-6;
+    ASSERT_LE(criterion, epsilon);
```
jbigot, line 301: I'd rather we use `EXPECT_*` instead of `ASSERT_*`

jbigot: here you can use `EXPECT_NEAR`

tests/fft.hpp
```diff
@@ -214,3 +216,87 @@ static void test_fft()
...
+        Ff0_expected = Kokkos::sqrt(2.);
+        break;
+    case ddc::FFT_Normalization::FULL:
+        Ff0_expected = 1 / Kokkos::sqrt(2 * Kokkos::numbers::pi);
+        break;
+    }
+
+    double criterion = Kokkos::abs(Ff(Ff.domain().front()) - Ff0_expected);
+
+    std::cout << "\n Distance between analytical prediction and numerical result : " << criterion;
```
jbigot, line 299: move that to the `EXPECT_*`, `cout` or `cerr` are not supposed to be used in gtest

tests/fft.hpp
jbigot, line 301:
```suggestion

    EXPECT_NEAR(Ff(Ff.domain().front()), Ff0_expected, 1e-6) << " Distance between analytical prediction and numerical result > 10^6";
```

tests/fft.hpp
```diff
@@ -214,3 +216,87 @@ static void test_fft()
...
+            = DDom<DDim<X>>(ddc::init_discrete_space(DDim<X>::
+                                                             init(ddc::Coordinate<X>(-1. / 4),
+                                                                  ddc::Coordinate<X>(1. / 4),
+                                                                  DVect<DDim<X>>(2))));
+    ddc::Chunk _f = ddc::Chunk(x_mesh, Allocator<MemorySpace, Tin>());
+    ddc::ChunkSpan f = _f.span_view();
+    ddc::for_each(
+            policy<ExecSpace>(),
+            ddc::get_domain<DDim<X>>(f),
+            DDC_LAMBDA(DElem<DDim<X>> const e) { f(e) = static_cast<Tin>(1); });
```
jbigot, line 233: what's `DElem` ?

tests/fft.hpp
```diff
@@ -214,3 +216,87 @@ static void test_fft()
...
+
+    ddc::init_fourier_space<X>(x_mesh);
+    DDom<DFDim<ddc::Fourier<X>>> const k_mesh
+            = ddc::FourierMesh(x_mesh, is_complex<Tin>::value && is_complex<Tout>::value);
+
+    ddc::Chunk _f_bis = ddc::Chunk(ddc::get_domain<DDim<X>>(f), Allocator<MemorySpace, Tin>());
+    ddc::ChunkSpan f_bis = _f_bis.span_view();
+    ddc::deepcopy(f_bis, f);
+
+    ddc::Chunk _Ff = ddc::Chunk(k_mesh, Allocator<MemorySpace, Tout>());
```
jbigot, line 243: `_Ff` is a reserved keyword in C++, we do not use underscores at the start of variable names for this reason .

In fact we should get back our coding standard that we lost in the split with gyselalib++ https://github.com/gyselax/gyselalibxx/blob/main/doc/CODING_STANDARD.md

tests/fft.hpp
```diff
@@ -214,3 +216,87 @@ static void test_fft()
     ASSERT_LE(criterion, epsilon);
     ASSERT_LE(criterion2, epsilon);
 }
+
+template <typename ExecSpace, typename MemorySpace, typename Tin, typename Tout, typename X>
+static void test_fft_norm(ddc::FFT_Normalization const norm)
+{
+    DDom<DDim<X>> const x_mesh
```
jbigot, line 223: What's `DDim`?

tests/fft.hpp
```diff
@@ -214,3 +216,87 @@ static void test_fft()
...
+            = DDom<DDim<X>>(ddc::init_discrete_space(DDim<X>::
```
jbigot, line 224: What's `DDom`?

tests/fft.hpp
```diff
@@ -214,3 +216,87 @@ static void test_fft()
...
+                                                                  DVect<DDim<X>>(2))));
+    ddc::Chunk _f = ddc::Chunk(x_mesh, Allocator<MemorySpace, Tin>());
+    ddc::ChunkSpan f = _f.span_view();
+    ddc::for_each(
+            policy<ExecSpace>(),
+            ddc::get_domain<DDim<X>>(f),
+            DDC_LAMBDA(DElem<DDim<X>> const e) { f(e) = static_cast<Tin>(1); });
+
+    ddc::init_fourier_space<X>(x_mesh);
+    DDom<DFDim<ddc::Fourier<X>>> const k_mesh
```
jbigot, line 236: What's `DFDim`?

tests/fft.hpp
tpadioleau: @blegouix you can find more information in this post https://stackoverflow.com/questions/228783/what-are-the-rules-about-using-an-underscore-in-a-c-identifier

tpadioleau: we should check if clang-tidy is able to detect this ?

tests/fft.hpp
blegouix: Yes

tests/fft.hpp
blegouix: Is there a convention to name the chunk _FFf in `ddc::ChunkSpan FFf = _FFf.span_view();` ?

tpadioleau: In the MR #158 I used `FFf_alloc`, I don't think we have a clear convention about that naming scheme.

tests/fft.hpp
tpadioleau: Ok then could you relate the two numbers by defining a variable ?

blegouix: That's the sum of elements of the input. Here f_1 + f_2 = 1+1 = 2.

`Ff0_expected = Ff(Ff.domain().front()) + Ff(Ff.domain.back())` ?

blegouix: I will commit that. If you change the `2` you also have to change the boundaries though:

`init(ddc::Coordinate<X>(-3. / 8), ddc::Coordinate<X>(3. / 8), DVect<DDim<X>>(4))));`

Reason is : for both configurations the FFT treats the domain [-1/2;1/2]. The boundaries of the mesh (-1/4 and 1/4, or -3/8 and 3/8) are the centers of first and last cells.

tests/fft.hpp
blegouix: I propose to merge this first and I will align MR #158 afterward

tpadioleau (dismissed):
Could you replace the calls to `get_domain<...>` by the member function `.domain()` of `Chunk` and `ChunkSpan` classes ?

tests/fft.hpp
jbigot: uhm... ok, I see. I'm not sure it's a good idea to introduce file-specific aliases, it makes the understanding all the more complicated for the reader.

tpadioleau (approved):

tests/fft.hpp
blegouix: I put as resolved to merge (because not in the scope of current merge request), if it is really important we can speak about it again later. Imo, most users will use aliases like we do in Gysela for big projects (far more readable) so using it in tests too gives them conventions to follow.

PR #154, opened 2023-07-05
Add a clang-tidy test

tpadioleau:


jbigot (approved):
nice!

tpadioleau:
Well yes but I am not able to build the docker images successfully. Shall we split them into smaller images ?

tpadioleau (commented):
ok

tpadioleau:
@jbigot @blegouix what do you think ? This currently enables `bugprone-reserved-identifier` and `hicpp-avoid-c-arrays` clang-tidy checks. Do you think there is any benefit of putting the checks into a `.clang-tidy` file at the root of the project ?

jbigot:
> @jbigot @blegouix what do you think ? This currently enables `bugprone-reserved-identifier` and `hicpp-avoid-c-arrays` clang-tidy checks. Do you think there is any benefit of putting the checks into a `.clang-tidy` file at the root of the project ?

To me, that would be nice so as to use it locally also

jbigot (approved):
nice!

PR #156, opened 2023-07-05
Split images into smaller images

tpadioleau:


tpadioleau:
@jbigot can you make a quick review of `tests.yml` before triggering the building of all the images ?

tpadioleau:
I suggest we switch to `oldest/latests` in the issue regarding the upgrade of images ?

.github/workflows/tests.yml
```diff
@@ -37,53 +45,58 @@ jobs:
...
     strategy:
       fail-fast: false
       matrix:
+        image: ['focal', 'jammy']
+        kokkos_backend: ['CPU', 'CUDA', 'HIP']
         cxx_version:  ['17', '20', '23']
         cmake_build_type: ['Debug', 'Release']
-        kokkos_backend: ['CPU-gnu', 'CPU-clang', 'GPU-HIP', 'GPU-CUDA']
-        image: ['test_env', 'jammy_env']
+        host_compiler: ['GCC', 'Clang']
```
jbigot, line 82: shouldn't this only be a choice when kokkos_backend is CPU?

jbigot:
looks good, but I'd trigger a run to know more

jbigot:
> I suggest we switch to `oldest/latest` in the issue regarding the upgrade of images ?

That would be great! :+1:

.github/workflows/tests.yml
tpadioleau: Why not but I have difficulties to get the right image depending on the backend and to retrieve the compiler when it is CPU. That is why I introducing a new dimension.

Maybe I could use `include` instead of `exclude` with this new dimension ?

tpadioleau:
> I suggest we switch to `oldest/latest` in the issue regarding the upgrade of images ?

Done in this PR

jbigot (approved):

jbigot (approved):

PR #159, opened 2023-07-10
Fft cleaning

tpadioleau:


tests/fft.hpp
```diff
@@ -20,198 +20,106 @@ using DDom = ddc::DiscreteDomain<DDim...>;
...
 
     double criterion2 = Kokkos::sqrt(ddc::transform_reduce(
-            ddc::get_domain<DDim<X>...>(FFf_host),
+            FFf_host.domain(),
             0.,
             ddc::reducer::sum<double>(),
             [=](DElem<DDim<X>...> const e) {
-                return Kokkos::pow(Kokkos::abs(FFf_host(e)) - Kokkos::abs(f_host(e)), 2)
-                       / Kokkos::pow(Nx, sizeof...(X));
+                double const diff = Kokkos::abs(FFf_host(e)) - Kokkos::abs(f_host(e));
```
tpadioleau, line 116: @blegouix Are we sure we want the difference of the absolute values and not the absolute value of the difference ?

tests/fft.hpp
```diff
@@ -20,198 +20,106 @@ using DDom = ddc::DiscreteDomain<DDim...>;
...
-                                                                            DFDim<ddc::Fourier<X>>>(
-                                                                         e)),
-                                                                 2)
-                                                     + ...)
-                                                   / 2),
-                                   2)
-                       / (LastSelector<std::size_t, X, X...>(Nx / 2, Nx) * ...);
+                double const xn2
+                        = (Kokkos::pow(ddc::coordinate(ddc::select<DFDim<ddc::Fourier<X>>>(e)), 2)
+                           + ...);
```
tpadioleau, line 104: In order to avoid the costly `pow` function I can suggest this trick:
```suggestion
                double tmp = 0;
                double const xn2 = ((tmp = ddc::coordinate(ddc::select<DFDim<ddc::Fourier<X>>>(e)), tmp * tmp) + ...);
```
I don't know if this is better or not

jbigot: or create a `ddc::detail::pow2(double)` function

tpadioleau: That is a good idea, a lambda should also do the trick!

tests/fft.hpp
tpadioleau: Let us see that later

PR #161, opened 2023-08-02
Successful build of gyselalibxx is not a requirement anymore

tpadioleau:


jbigot (approved):
looks good

tpadioleau:
> looks good

I cannot find where to trigger gyselalibxx build ?

jbigot:
I believe this will appear in the "actions" tab once in main

tpadioleau:
> I'd remove both, if we remove the action on gitlab side, the trigger will return an error

Ok good point

tpadioleau:
I think we are good

jbigot (approved):

PR #170, opened 2023-08-29
Use unique_ptr to automatically call the destroy function

tpadioleau:


include/ddc/kernels/fft.hpp
```diff
@@ -431,6 +433,13 @@ void core(
 
         cufftHandle plan = -1;
         cufftResult cufft_rt = cufftCreate(&plan);
+
+        if (cufft_rt != CUFFT_SUCCESS)
+            throw std::runtime_error("cufftPlan failed");
+
+        std::unique_ptr<cufftHandle, std::function<void(cufftHandle*)>> const
+                raii_plan(&plan, [](cufftHandle* handle) { cufftDestroy(*handle); });
```
tpadioleau, line 441: @jbigot Unfortunately the `plan` variable needs to stay in the scope because I pass its address to the `unique_ptr`. Not a really satisfying solution. A better solution would be to wrap it and call `*Create` and `*Destroy` in the `ctor` and `dtor` respectively.

tpadioleau: shall do the job for now

blegouix (approved):
ok!

PR #174, opened 2023-09-04
Ensure serial algorithms can be used inside parallel algorithms

tpadioleau:


tpadioleau:
The `deepcopy` algorithm has semantic issues. In this new implementation you need to pass `parallel_device` (or even `parallel_host` in fact) to trigger the copy between host and device.

jbigot (approved):
Looks great!

jbigot:
Is there more required to move this out of WIP status?

tpadioleau:
I think it is close to be ready:
- it needs a rebase
- maybe explore again if `constexpr` can only be replaced by `KOKKOS_FUNCTION`
- I guess some documentation about what is possible and what is not like (callable from, executed on) ?

tpadioleau:
could help #193

include/ddc/for_each.hpp
```diff
@@ -22,13 +22,13 @@ namespace ddc {
 namespace detail {
 
 template <class RetType, class Element, std::size_t N, class Functor, class... Is>
-void for_each_serial(
+KOKKOS_FUNCTION void for_each_serial(
```
jbigot, line 25: don't you end up with the previous problem if the provided functor is CPU-only even if it is called from CPU only?

jbigot (approved):
Looks great!

include/ddc/for_each.hpp
tpadioleau: Good question, actually I did not get any warning for now.

I forced the compiler to compile a host-only `operator()` and indeed I got a warning.

tpadioleau: I see multiple options:
1. we keep this implementation and hope it does not trigger too many warnings, a bit disappointing though
2. we keep this implementation and ask people to always use `KOKKOS_LAMBDA`, does not cover our usage in gysela
3. we ask people to use it only inside a parallel algorithm, does not cover our usage in gysela
4. we implement `host_for_each` that should be used when it is called inside a host-only function (note that we should have the same problem when called inside a device-only function, should be quite rare though)
5. we implement `host_for_each` and a `parallel_for_each` that should accept a `ThreadHandle` that you get only from a `parallel_*` algorithm.

I guess Kokkos algorithms don't have this issue because they don't have sequential algorithms.

tpadioleau: Some piece of information https://llvm.org/docs/CompileCudaWithLLVM.html#overloading-based-on-host-and-device-attributes about what they call the "wrong-side rule"

jbigot: What about having a different implementation for host and for device? something like:
```c++
#ifdef ON_HOST
using current_execution_space = Kokkos::DefaultExecutionSpace;
#else
using current_execution_space = Kokkos::DefaultHostExecutionSpace;
#endif

template <class ExecutionSpace = current_execution_space, class RetType, class Element, std::size_t N, class Functor, class... Is>
struct for_each;

template <class RetType, class Element, std::size_t N, class Functor, class... Is>
struct for_each <Kokkos::DefaultExecutionSpace, RetType,Element,N, Functor, Is...> {
KOKKOS_FUNCTION void operator()(...);
};

template <class RetType, class Element, std::size_t N, class Functor, class... Is>
struct for_each <Kokkos::DefaultHostExecutionSpace, RetType,Element,N, Functor, Is...> {
void operator()(...);
};
```

tpadioleau: How the user would call it ? With a global object like ranges ?

tpadioleau: I think I have implemented your proposal, see https://godbolt.org/z/r7cK5cfd8. It emits the same warning.

To make it work we would need to know whether the host or the device is calling the `for_each` function but I am not aware of any way to access this information with CUDA.

jbigot: Indeed, my understanding of `__host__` behavior for nvcc seems wrong, Cf. https://godbolt.org/z/xddsdKvGx
I'm afraid there's no good solution

blegouix: Don't know if it suits your problem but macros KOKKOS_IF_ON_DEVICE() and KOKKOS_IF_ON_HOST() allow to execute different codes depending of if a KOKKOS_FUNCTION is called from host or device.

tpadioleau: Unfortunately it is not enough. The problem is that on the device compilation pass, `nvcc` sees that we call a host function from a host-device fonction therefore emits a warning. I do not think we can do much about it. I will make a host-only `host_for_each` function.

include/ddc/transform_reduce.hpp
```diff
@@ -58,9 +58,54 @@ T transform_reduce_serial(
...
+        }
+        return result;
+    }
+    DDC_IF_NVCC_THEN_POP
+}
+
 } // namespace detail
 
 /** A reduction over a nD domain in serial
+ * This version must only be called from a host-device function.
```
tpadioleau, line 108:
```suggestion
 * This version must only be called from a host-device function and provide host-device reducer and transform functions.
```

include/ddc/for_each.hpp
```diff
@@ -38,14 +38,47 @@ void for_each_serial(
...
+        for (Element ii = begin[I]; ii < end[I]; ++ii) {
+            host_for_each_serial<RetType>(begin, end, f, is..., ii);
+        }
+    }
+}
+
 } // namespace detail
 
 /** iterates over a nD domain in serial
+ * This version must only be called from a host-device function.
```
tpadioleau, line 61:
```suggestion
 * This version must only be called from a host-device function and provide host-device reducer and transform functions.
```

include/ddc/transform_reduce.hpp
```diff
@@ -58,9 +58,54 @@ T transform_reduce_serial(
     DDC_IF_NVCC_THEN_POP
 }
 
+/** A serial reduction over a nD domain
+ * @param[in] domain the range over which to apply the algorithm
```
tpadioleau, line 62:
```suggestion
/** A serial reduction over a nD domain
 * This version must only be called from the host.
 * @param[in] domain the range over which to apply the algorithm
```

include/ddc/transform_reduce.hpp
tpadioleau, line 108:
```suggestion
 * This version must only be called with `reducer`  and `transform` functions compiled for all execution spaces. See also `host_transform_reduce` when these functions are compiled only for the host.
```

include/ddc/for_each.hpp
```diff
@@ -58,7 +58,7 @@ void host_for_each_serial(
 } // namespace detail
 
 /** iterates over a nD domain in serial
- * This version must only be called from a host-device function.
+ * This version must only be called with `f` being a function compiled for all execution spaces. See also `host_for_each` when this function is only compiled for the host.
```
jbigot, line 61:
```suggestion
 * This version must only be called with `f` being a function compiled for all execution spaces. See also \ref host_for_each when this function is only compiled for the host.
```

tpadioleau:
closing for now, archiving the branch

PR #178, opened 2023-09-12
ddc policy obtained from Kokkos ExecSpace

blegouix:


include/ddc/for_each.hpp
```diff
@@ -197,6 +197,17 @@ inline constexpr serial_host_policy serial_host;
...
+template <typename ExecSpace>
+constexpr auto policy = [] {
+    if constexpr (std::is_same_v<ExecSpace, Kokkos::Serial>) {
+        return ddc::policies::serial_host;
+    } else if constexpr (std::is_same_v<ExecSpace, Kokkos::OpenMP>) {
+        return ddc::policies::parallel_host;
+    } else {
+        return ddc::policies::parallel_device;
+    }
+};
```
tpadioleau, line 209: What do you think of a version using objets like:
```suggestion
template <typename ExecSpace>
constexpr auto policy = [] ([[maybe_unused]] ExecSpace exec_space) {
    if constexpr (std::is_same_v<ExecSpace, Kokkos::Serial>) {
        return ddc::policies::serial_host;
    } else if constexpr (std::is_same_v<ExecSpace, Kokkos::OpenMP>) {
        return ddc::policies::parallel_host;
    } else {
        return ddc::policies::parallel_device;
    }
};
```
That can be called
```cpp
ddc::policies::policy(Kokkos::DefaultExecutionSpace());
```

blegouix: I do not understand why but it does not like it:
`/home/BL262078/ddc/tests/fft.hpp(131): error: function "lambda [](Kokkos::Serial)->ddc::serial_host_policy" cannot be called with the given argument list
            object type is: const lambda [](Kokkos::Serial)->ddc::serial_host_policy
          detected during instantiation of "void test_fft_norm<ExecSpace,MemorySpace,Tin,Tout,X>(ddc::FFT_Normalization) [with ExecSpace=Kokkos::Serial, MemorySpace=Kokkos::Serial::memory_space, Tin=flo
at, Tout=Kokkos::complex<float>, X=RDimX]"
/home/BL262078/ddc/tests/fft.cpp(16): here
`
And lot of similar errors.

Compiled with nvcc

tpadioleau (approved):
Ok for me

PR #179, opened 2023-09-15
Kokkos policy change how ExecSpace is passed

blegouix:


tests/fft.hpp
```diff
@@ -42,7 +42,7 @@ static void test_fft()
     ddc::Chunk _f(x_mesh, ddc::KokkosAllocator<Tin, MemorySpace>());
     ddc::ChunkSpan f = _f.span_view();
     ddc::for_each(
-            ddc::policies::policy<ExecSpace>(),
+            ddc::policies::policy(ExecSpace()),
```
tpadioleau, line 45: Could you create a named variable for the execution space and pass it both to ddc::policies::policy and fft ?

tpadioleau (changes requested):

tpadioleau (approved):
Looks good to me thanks

PR #180, opened 2023-09-15
TypeSeqReplace

blegouix:
Example usage:

`type_seq_replace_t<TypeSeq<X,Y,Z>,TypeSeq<Y,Z>,TypeSeq<BSplines<Y>,BSplines<Z>>>` will give `TypeSeq<X,BSplines<Y>,BSplines<Z>>`

### DIFF
--- a/include/ddc/detail/type_seq.hpp
+++ b/include/ddc/detail/type_seq.hpp
@@ -101,6 +101,70 @@ struct TypeSeqMerge<TypeSeq<TagsA...>, TypeSeq<HeadTagsB, TailTagsB...>, TypeSeq
 {
 };
 
+/// A is replaced by element of C at same position than the first element of B equal to A.
+/// Remark : It may not be usefull in its own, it is an helper for TypeSeqReplace
+template <class TagA, class TagSeqB, class TagSeqC>
+struct TypeSeqReplaceSingle;
+
+template <class TagA>
+struct TypeSeqReplaceSingle<TagA, TypeSeq<>, TypeSeq<>>
+{
+    using type = TagA;
+};
+
+template <class TagA, class... TagsB>
+struct TypeSeqReplaceSingle<TagA, TypeSeq<TagsB...>, TypeSeq<>>
+{
+    using type = TagA;
+};
+
+template <class TagA, class... TagsC>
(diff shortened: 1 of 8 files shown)


tests/type_seq.cpp
```diff
@@ -6,11 +6,15 @@
 
 #include <gtest/gtest.h>
 
+#include "ddc/detail/type_seq.hpp"
+
```
tpadioleau, line 10:
```suggestion
```

include/ddc/detail/type_seq.hpp
```diff
@@ -101,6 +101,70 @@ struct TypeSeqMerge<TypeSeq<TagsA...>, TypeSeq<HeadTagsB, TailTagsB...>, TypeSeq
...
+{
+    using type = TagA;
+};
+
+template <class TagA, class... TagsC>
+struct TypeSeqReplaceSingle<TagA, TypeSeq<>, TypeSeq<TagsC...>>
+{
+    using type = TagA;
+};
+
```
tpadioleau, line 126:
```suggestion
```

I think it can work without these two specializations. Am I correct that we only want to cover the case where `B` and `C` have the same size ?

tpadioleau (changes requested):
Yes, good idea.

blegouix:
I also added the function replace_dim_of which replaces in DDom_a the dimension Dim1 by the dimension Dim2 of DDom_b

It is then very conveniant to produce the domain of batched splines whatever dimension is transformed.

tests/discrete_domain.cpp
```diff
@@ -129,6 +133,16 @@ TEST(ProductMDomainTest, Diff)
     EXPECT_EQ(subdomain, dom_x);
 }
 
+TEST(ProductMDomainTest, Replace)
+{
+    DDomXY const dom_x_y = DDomXY();
+    DDomZ const dom_z = DDomZ();
+    DDomXZ const dom_x_z = DDomXZ();
```
tpadioleau, line 140: Can you give them different sizes to be sure this information gets also correctly replaced in `subdomain`.

tpadioleau (approved):
Seems good to me! let's merge

PR #181, opened 2023-09-23
DiscreteElement/DiscreteDomain constructors + Chunk/Chunspan operator() generalized 

blegouix:
First merge #180 

Supports any combination of dimensionality.

(Schematic) example usage:

```
// Two lines below are not equivalent because order of template dimensions is not the same:
auto spline_domain = ddc::replace_dim_of<interpolation_mesh_type,bsplines_type>(vals_domain,bsplines_domain); // replaces interpolation dim with bsplines_dim
auto spline_domain_tr = ddc::DiscreteDomain(bsplines_domain, ddc::remove_dims_of(vals_domain, interpolation_domain)); // remove interpolation dim and pushes bsplines_dim as the first dimension
```

It can then be used to perform layout transpose

```
ddc::Chunkspan spline //... ;
ddc::Chunkspan spline_tr //... ;
for (ddc::DiscreteElement<bsplines_type> i) {
  for  (ddc::DiscreteElement<batch_type1, batch_type2> j) {
    spline_tr(i,j) = spline(i,j); // <- underlying data array do not have same layout!
  }
}
// Or deepcopy, but it is an example usage of the generalized DiscreteElement()
```

### DIFF
--- a/include/ddc/discrete_element.hpp
+++ b/include/ddc/discrete_element.hpp
@@ -84,7 +84,7 @@ constexpr DiscreteElement<QueryTag> take(
         DETail const&... tail)
 {
     DDC_IF_NVCC_THEN_PUSH_AND_SUPPRESS(implicit_return_from_non_void_function)
-    /* TODO
+    /* TODO : Restore
     static_assert(
             !type_seq_contains_v<detail::TypeSeq<HeadTag>, detail::TypeSeq<TailTags...>>,
             "ERROR: tag redundant");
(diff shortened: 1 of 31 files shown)


EmilyBourne (approved):

include/ddc/discrete_element.hpp
```diff
@@ -81,24 +82,44 @@ KOKKOS_FUNCTION constexpr DiscreteElement<QueryTags...> select(
...
+              std::true_type,
+              typename Contains<QueryTag, DETail...>::type>
+{
+};
+
+template <class QueryTag, class... DE>
+using contains_t = typename Contains<QueryTag, DE...>::type;
+
+template <class QueryTag, class... DE>
+inline constexpr bool contains_v = contains_t<QueryTag, DE...>::value;
```
tpadioleau, line 106: I think we can develop a conversion tool between `DiscreteElement`/`DiscreteVector` and `TypeSeq` to reuse more what is already available in `TypeSeq`

include/ddc/discrete_element.hpp
```diff
@@ -154,15 +175,14 @@ class DiscreteElement
 
     template <class... OTags>
     explicit KOKKOS_FUNCTION constexpr DiscreteElement(
-            DiscreteElement<OTags> const&... other) noexcept
-        : m_values {take<Tags>(other...).uid()...}
+            DiscreteElement<OTags...> const& other) noexcept
+        : m_values {other.template uid<Tags>()...}
     {
     }
```
tpadioleau, line 181: I am not sure when this constructor will be preferred over the next constructor

tpadioleau (commented):
I will also suggest an other implementation. The case of `DiscreteVector` also needs to be addressed

tpadioleau:
Addressed in #207

PR #183, opened 2023-09-28
Kokkos For 0D

blegouix:
Not working

### DIFF
--- a/include/ddc/for_each.hpp
+++ b/include/ddc/for_each.hpp
@@ -30,17 +30,43 @@ class ForEachKokkosLambdaAdapter
 public:
     ForEachKokkosLambdaAdapter(F const& f) : m_f(f) {}
 
+    KOKKOS_IMPL_FORCEINLINE void operator()() const
+    {
+        m_f(DiscreteElement<>());
+    }
+
+    KOKKOS_FORCEINLINE_FUNCTION void operator()(use_annotated_operator) const
+    {
+        m_f(DiscreteElement<>());
+    }
+
+    template <std::size_t N = sizeof...(DDims), std::enable_if_t<(N > 0), bool> = true>
     KOKKOS_IMPL_FORCEINLINE void operator()(index_type<DDims>... ids) const
     {
         m_f(DiscreteElement<DDims...>(ids...));
     }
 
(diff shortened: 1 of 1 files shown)


blegouix:
Working

tests/for_each.cpp
```diff
@@ -113,6 +146,8 @@ void TestForEachParallelDeviceOneDimension()
 
 } // namespace
 
+
+
```
tpadioleau, line 150:
```suggestion
```

tests/for_each.cpp
```diff
@@ -92,6 +101,30 @@ TEST(ForEachParallelHost, TwoDimensions)
...
+    EXPECT_EQ(sum, dom.size());
+}
+} // namespace
+TEST(ForEachParallelDevice, ZeroDimension)
+{
+    TestForEachParallelDeviceZeroDimension();
+}
+namespace {
 void TestForEachParallelDeviceOneDimension()
 {
```
tpadioleau, line 129:
```suggestion
    EXPECT_EQ(sum, dom.size());
}

} // namespace

TEST(ForEachParallelDevice, ZeroDimension)
{
    TestForEachParallelDeviceZeroDimension();
}

namespace {

void TestForEachParallelDeviceOneDimension()
{
```

tpadioleau (changes requested):
Thanks seems good for the `for_each` algorithm. Could you also update `transform_reduce` algorithm ? If you cannot do it now, could you open an issue relative to the other DDC algorithms ?

include/ddc/transform_reduce.hpp
```diff
@@ -137,14 +137,34 @@ class TransformReducerKokkosLambdaAdapter
     TransformReducerKokkosLambdaAdapter(Reducer const& r, Functor const& f) : reducer(r), functor(f)
     {
     }
+    template <std::size_t N = sizeof...(DDims), std::enable_if_t<(N == 0), bool> = true>
```
tpadioleau, line 140:
```suggestion

    template <std::size_t N = sizeof...(DDims), std::enable_if_t<(N == 0), bool> = true>
```

tpadioleau (dismissed):
Seems good to me!

tpadioleau (approved):

PR #184, opened 2023-10-02
Add tagged vector conversion operator

EmilyBourne:


include/ddc/detail/tagged_vector.hpp
```diff
@@ -401,6 +401,12 @@ class TaggedVector : public ConversionOperators<TaggedVector<ElementType, Tags..
         ((m_values[type_seq_rank_v<Tags, tags_seq>] *= rhs.template get<Tags>()), ...);
         return *this;
     }
+
+    template <class OElementType>
+    explicit operator TaggedVector<OElementType, Tags...>()
```
tpadioleau, line 406: Looking at the constructors, there should already be one conversion constructor. It is not selected because it takes input parameters by `const&` thus disabling the conversion from a `Chunk<T> const&` to a `ChunkSpan<T>` (only `ChunkSpan<const T>` is allowed from a `Chunk<T> const&`).

I still need to think if it is compatible with the rest of the API. Is it urgent ?

blegouix:
Should we add:

`
template<class ElementType, class SupportType, class Allocator, class Tags...>
using VectorField = ddc::detail::TaggedVector<ddc::Chunk<ElementType, SupportType, Allocator>, Tags...>
`
And same for `VectorFieldSpan `and `VectorFieldView `? Or let user choose how to build his own vector fields and rename ddc::detail::TaggedVector in ddc::Array (more evocative imo because ` std::array<ElementType, sizeof...(Tags)> m_values;`) ?

Maybe we could add post-process computation of flux density field in HeatEquation example to show to the user this solution for vector fields.

include/ddc/detail/tagged_vector.hpp
EmilyBourne: No I can use a workaround in the meantime

jbigot: since we can do it, offering a conversion constructor should be preferred to a conversion operator.

What about offering `template <class OElementType> TaggedVector::TaggedVector(TaggedVector<OElementType, Tags...>)`?

EmilyBourne: The conversion constructor already exists:
https://github.com/Maison-de-la-Simulation/ddc/blob/d401a6d36477775dbca108a6490c5a119cb99ee5/include/ddc/detail/tagged_vector.hpp#L260-L265

jbigot: but it takes a `cont &` and this is not a solution in your case. We could add a non-const ref one

jbigot: Would it not solve the issue?

EmilyBourne: Fixed. I think this fixes the issue. But I don't have the problem anymore as I changed strategy and ended up writing the `VectorField` class

jbigot: Thanks, looks good to me, @tpadioleau ?

jbigot:
Also I wonder whether we should add an `enable_it` `std::is_convertible` between `OElementType` and `ElementType` (this is already a question for the const-ref version)

jbigot (commented):
I'd say let's go, but I think @tpadioleau has a better view of the things at stake here.

tests/tagged_vector.cpp
```diff
@@ -73,3 +73,13 @@ TEST(TaggedVector, Operators)
     ASSERT_EQ(4 + c, (ddc::detail::TaggedVector<int, double>(5)));
     ASSERT_EQ(4 * a, (ddc::detail::TaggedVector<int, double, float>(4, 8)));
 }
+
+TEST(TaggedVector, Conversion)
+{
+    struct Tag1;
+    struct Tag2;
+    ddc::detail::TaggedVector<int, Tag1, Tag2> a(1, 2);
+    ddc::detail::TaggedVector<double, Tag1, Tag2> b(a);
```
jbigot, line 82:
```suggestion
    ddc::detail::TaggedVector<double, Tag2, Tag1> b(a);
```

Should that work also?

EmilyBourne: Yes it should. I have added an additional test

jbigot: great, thanks :+1:

PR #186, opened 2023-10-04
Debug operator chunkspan

blegouix:
No merge planned. Just debug purpose


tpadioleau:
This seems to work. I think it raises the question of annotationg functions even though they are marked `constexpr`.

tpadioleau (approved):

PR #191, opened 2023-10-06
Add element_type member of TaggedVector

EmilyBourne:


include/ddc/detail/tagged_vector.hpp
```diff
@@ -242,6 +242,8 @@ class TaggedVector : public ConversionOperators<TaggedVector<ElementType, Tags..
     std::array<ElementType, sizeof...(Tags)> m_values;
 
 public:
+    using element_type = ElementType;
```
tpadioleau, line 245: To mimic `std::array` I would expect `value_type`, what do you think ?
```suggestion
    using value_type = ElementType;
```

EmilyBourne: This is fine for me. I used `element_type` to match `Chunk`

EmilyBourne: It seems that there is a `value_type` too. What is the difference?

tpadioleau (approved):

include/ddc/detail/tagged_vector.hpp
tpadioleau: In `ChunkSpan` ? It is inspired from `mdspan`, they seem to define it as `using value_type = remove_cv_t<element_type>`

PR #192, opened 2023-10-07
kokkos::finalize() when sigint

blegouix:
Calls Kokkos::finalize() in ScopeGuard when SIGINT.

Purpose: Kokkos-Tools like https://github.com/kokkos/kokkos-tools/blob/develop/profiling/simple-kernel-timer/kp_kernel_timer.cpp may write profiling files when Kokkos is finalized. It is usefull to profile Gysela without reaching the end of the simulation.

(Maybe it could be proposed to Kokkos)

### DIFF
--- a/include/ddc/scope_guard.hpp
+++ b/include/ddc/scope_guard.hpp
@@ -5,6 +5,7 @@
 #include <optional>
 
 #include <Kokkos_Core.hpp>
+#include <signal.h>
 
 #include "discrete_space.hpp"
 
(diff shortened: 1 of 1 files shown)


jbigot:
shouldn't the destructor be call is the signal is not handled?

tpadioleau:
Runtime parameters of Gysela like the final time or the number of time steps are not enough ?

blegouix:
It is just for conveniance, like getting the Kokkos-tool writing its report without having to carefully choose the final time to get a precise-duration simulation when debugging. But I can understand it can be annoying for user to not have SIGINT immediatly kill its process. Even an "option" in ScopeGuards template like `FinalizeWhenSigint ` with default value to off won't be satisfying ?

jbigot:
While I understand the point of this, having a library handle a signal is just not possible. What happens if another library tries it too? This is very much like handling errors with `abort`, only the top level user can make a choice. Hence my question. What about handling the signal with `exit()` in user code? This should unwrap the call trace and do what you want.

tpadioleau:
In order to profile the code, don't you need to perform one or two complete time iterations ? The parameter `nbiter` seems adapted to that workflow.

You also need to be careful if you stop the execution during the kernel you are profiling. The numbers printed would need to be interpreted with care.

I am am not in favor of this feature in DDC

PR #201, opened 2023-10-12
Splines release

blegouix:
First merge #181 .

Release of Splines on GPU with Ginkgo. No LAPACK backend, no non-periodic boundary conditions, no polar coordinates and no 2d or nd spline_builder/evaluator.

Spline_builder.hpp and Spline_Evaluator.hpp are non-batched 1D tools which are residuals from Gysela/sll. I need to keep them to avoid interfere with futur work on Bcs, Polar or 2d/nd tools but it is planned to remove them at the very end of the project and rename Spline_builder/evaluator_batched.hpp to Spline_builder/evaluator.hpp


examples/characteristics_advection.cpp
```diff
@@ -0,0 +1,328 @@
...
+    // max(1/dy^2)
+    double const invdy2_max = ddc::transform_reduce(
+            y_domain,
+            0.,
+            ddc::reducer::max<double>(),
+            [](ddc::DiscreteElement<DDimY> iy) {
+                return 1.
+                       / (ddc::distance_at_left(iy)
+                          * ddc::distance_at_right(iy));
+            });
```
tpadioleau, line 159:
```suggestion
```

Does not seem to be needed

examples/characteristics_advection.cpp
```diff
@@ -0,0 +1,328 @@
...
+    ddc::Coordinate<T> const max_dt {0.1};
```
tpadioleau, line 160: To be moved to the parameters section

examples/characteristics_advection.cpp
```diff
@@ -0,0 +1,328 @@
...
+    ddc::for_each(
+            ddc::policies::parallel_device,
+            x_mesh,
+            DDC_LAMBDA(ddc::DiscreteElement<DDimX, DDimY> const ixy) {
+                double const x
+                        = ddc::coordinate(ddc::select<DDimX>(ixy));
+                double const y
+                        = ddc::coordinate(ddc::select<DDimY>(ixy));
+                // initial_density(ixy) = 9.999 * Kokkos::exp(-(x * x + y * y) / 0.1 / 2);
+                initial_density(ixy) = 9.999 * ((x * x + y * y) < 0.25);
```
tpadioleau, line 203: Looks quite discontinuous to be precisely represented by splines ?

examples/characteristics_advection.cpp
```diff
@@ -0,0 +1,328 @@
...
+                ddc::policies::parallel_device,
+                feet_coords.domain(),
+                DDC_LAMBDA(ddc::DiscreteElement<DDimX, DDimY> const e) {
+                    feet_coords(e) = ddc::Coordinate<X, Y>(
+                            ddc::coordinate(ddc::select<DDimX>(e))
+                                    - ddc::Coordinate<X>(
+                                            vx * ddc::step<DDimT>()),
+                            ddc::coordinate(ddc::select<DDimY>(e)));
+                    // Remark : feet_coords could be a chunk of ddc::Coordinate<X> and this line would be replaced by:
+                    // feet_coords(e) = ddc::coordinate(e) - ddc::Coordinate<X>(vx*ddc::step<DDimT>());
```
tpadioleau, line 295: Seems better to avoid extra storage ? I would even consider only this case

include/ddc/ddc.hpp
```diff
@@ -35,3 +37,41 @@
 #if defined(DDC_BUILD_PDI_WRAPPER)
 #include "ddc/pdi.hpp"
 #endif
+
+#if ginkgo_AVAIL
+#include <ginkgo/ginkgo.hpp>
+#endif
```
tpadioleau, line 43: The following function already assumes `Ginkgo` is available, no need to `ifdef` only that part
```suggestion
#include <ginkgo/ginkgo.hpp>
```

include/ddc/kernels/splines/bspline.hpp
tpadioleau, line 1: to be removed

include/ddc/kernels/splines/bernstein.hpp
tpadioleau, line 1: to be removed

tests/splines/bernstein.cpp
tpadioleau, line 1: to be removed

tests/splines/chunk.cpp
tpadioleau, line 1: to be removed

tests/splines/gauss_legendre_integration.cpp
tpadioleau, line 1: to be removed

include/ddc/kernels/splines/gauss_legendre_integration.hpp
tpadioleau, line 1: to be removed

tests/splines/non_periodic_spline_builder.cpp
tpadioleau, line 1: to be removed

include/ddc/kernels/splines/mapping/barycentric_coordinates.hpp
tpadioleau, line 1: to be removed

include/ddc/kernels/splines/matrix_sparse.hpp
```diff
@@ -0,0 +1,255 @@
...
+                b_last_buffer("b_last_buffer", n, cols_per_last_par_chunk);
+
+        // Sequential loop
+        for (int i = 0; i < n_seq_chunks; i++) {
+            int n_par_chunks_in_seq_chunk = i < n_seq_chunks - 1 ? par_chunks_per_seq_chunk
+                                                                 : par_chunks_per_last_seq_chunk;
+            // Parallel loop
+            Kokkos::parallel_for(
+                    Kokkos::RangePolicy<
+                            Kokkos::DefaultHostExecutionSpace>(0, n_par_chunks_in_seq_chunk),
```
tpadioleau, line 180: I have a few questions regarding this loop:
- what about a pure OpenMP run ? In the current implementation I think each OpenMP thread will open a new parallel region inside Ginkgo ?
- Have you checked that it is valid to call these Kokkos functions inside a `parallel_for`, be it CPU ?
- In terms of performance, can you show using NVIDIA tools that we have a benefit of using OpenMP + CUDA ?

include/ddc/kernels/splines/matrix_sparse.hpp
```diff
@@ -0,0 +1,255 @@
...
+                            Kokkos::deep_copy(
+                                    Kokkos::subview(b_view, Kokkos::ALL, par_chunk_window),
+                                    b_par_chunk);
+                        }
+                    });
+        }
+        return 1;
+    }
+};
+} // namespace ddc::detail
```
tpadioleau, line 255:
```suggestion
};

} // namespace ddc::detail
```

CMakeLists.txt
```diff
@@ -261,6 +261,26 @@ if("${HIP_FOR_NVIDIA}")
...
+  if(NOT TARGET Ginkgo::ginkgo)
+	  find_package(Ginkgo QUIET)
+	  if(NOT Ginkgo_FOUND)
+  	    add_subdirectory("${GINKGO_PATH}")
+	  endif()
+  endif()
+elseif("${DDC_Ginkgo_DEPENDENCY_POLICY}" STREQUAL "EMBEDDED")
+  add_subdirectory("${GINKGO_PATH}")
+elseif("${DDC_Ginkgo_DEPENDENCY_POLICY}" STREQUAL "INSTALLED")
+  find_package(Ginkgo REQUIRED)
```
tpadioleau, line 280:
```suggestion
  find_package(Ginkgo REQUIRED VERSION 1.6.0)
```

examples/characteristics_advection.cpp
tpadioleau, line 203: Looks quite discontinuous to be precisely represented by splines. Let's use the gaussian function.
```suggestion
                initial_density(ixy) = 9.999 * Kokkos::exp(-(x * x + y * y) / 0.1 / 2);
```

include/ddc/kernels/splines/matrix_maker.hpp
```diff
@@ -0,0 +1,18 @@
...
+class MatrixMaker
+{
+public:
+    template <typename ExecSpace>
+    static std::unique_ptr<Matrix> make_new_sparse(int const n)
+    {
+        return std::make_unique<Matrix_Sparse<ExecSpace>>(n);
+    }
+};
+} // namespace ddc::detail
```
tpadioleau, line 18: I would simplify like this

```suggestion
namespace ddc::detail {

template <typename ExecSpace>
std::unique_ptr<Matrix> make_new_sparse(int const n)
{
    return std::make_unique<Matrix_Sparse<ExecSpace>>(n);
}

} // namespace ddc::detail
```

include/ddc/kernels/splines/spline_evaluator_batched.hpp
```diff
@@ -0,0 +1,263 @@
+#pragma once
+
+#include <array>
+
+#include <ddc/ddc.hpp>
+
+#include "ddc/for_each.hpp"
+
+#include "Kokkos_Macros.hpp"
```
tpadioleau, line 9:
```suggestion
#include <ddc/ddc.hpp>

#include <Kokkos_Core.hpp>
```

include/ddc/kernels/splines/spline_evaluator_batched.hpp
```diff
@@ -0,0 +1,263 @@
...
+        return ddc::discrete_space<bsplines_type>().full_domain();
+    }
+
+    KOKKOS_INLINE_FUNCTION batch_domain_type const batch_domain() const noexcept
+    {
+        return ddc::remove_dims_of(spline_domain(), bsplines_domain());
+    }
+    template <class Layout>
+    double operator()(
+            ddc::Coordinate<IDimX...> const& coord_eval,
```
tpadioleau, line 110: I think you meant `tag_type` instead of `IDimX`

include/ddc/kernels/splines/spline_evaluator_batched.hpp
```diff
@@ -0,0 +1,263 @@
...
+            ddc::ChunkSpan<
+                    ddc::Coordinate<CoordsDims...> const,
+                    vals_domain_type,
+                    Layout2,
+                    memory_space> const coords_eval,
+            ddc::ChunkSpan<double const, spline_domain_type, Layout3, memory_space> const
+                    spline_coef) const
+    {
+        interpolation_domain_type interpolation_domain
+                = ddc::select<interpolation_mesh_type>(spline_eval.domain());
```
tpadioleau, line 129:
```suggestion
        interpolation_domain_type const interpolation_domain(spline_eval.domain());
```

include/ddc/kernels/splines/spline_evaluator_batched.hpp
```diff
@@ -0,0 +1,263 @@
...
+                KOKKOS_CLASS_LAMBDA(typename batch_domain_type::discrete_element_type const j) {
+                    const auto spline_eval_1D = spline_eval[j];
+                    const auto coords_eval_1D = coords_eval[j];
+                    const auto spline_coef_1D = spline_coef[j];
+                    for (int i = 0; i < interpolation_domain.size(); i++) {
+                        spline_eval_1D(typename interpolation_domain_type::discrete_element_type(i))
+                                = eval(coords_eval_1D(typename interpolation_domain_type::
+                                                              discrete_element_type(i)),
+                                       spline_coef_1D);
+                    }
```
tpadioleau, line 142:
```suggestion
                    for ( auto const i : interpolation_domain ) {
                        spline_eval_1D(i) = eval(coords_eval_1D(i), spline_coef_1D);
                    }
```

include/ddc/kernels/splines/spline_evaluator_batched.hpp
```diff
@@ -0,0 +1,263 @@
...
+            ddc::ChunkSpan<
+                    ddc::Coordinate<CoordsDims...> const,
+                    vals_domain_type,
+                    Layout2,
+                    memory_space> const coords_eval,
+            ddc::ChunkSpan<double const, spline_domain_type, Layout3, memory_space> const
+                    spline_coef) const
+    {
+        interpolation_domain_type interpolation_domain
+                = ddc::select<interpolation_mesh_type>(spline_eval.domain());
```
tpadioleau, line 167:
```suggestion
        interpolation_domain_type const interpolation_domain(spline_eval.domain());
```

include/ddc/kernels/splines/spline_evaluator_batched.hpp
```diff
@@ -0,0 +1,263 @@
...
+                    const auto spline_eval_1D = spline_eval[j];
+                    const auto coords_eval_1D = coords_eval[j];
+                    const auto spline_coef_1D = spline_coef[j];
+                    for (int i = 0; i < interpolation_domain.size(); i++) {
+                        spline_eval_1D(typename interpolation_domain_type::discrete_element_type(i))
+                                = eval_no_bc<eval_deriv_type>(
+                                        coords_eval_1D(typename interpolation_domain_type::
+                                                               discrete_element_type(i)),
+                                        spline_coef_1D);
+                    }
```
tpadioleau, line 181:
```suggestion
                    for ( auto const i : interpolation_domain ) {
                        spline_eval_1D(i) = eval_no_bc<eval_deriv_type>(coords_eval_1D(i), spline_coef_1D);
                    }
```

include/ddc/kernels/splines/spline_evaluator_batched.hpp
```diff
@@ -0,0 +1,263 @@
...
+    }
+
+    template <class Layout1, class Layout2, class Layout3, class... CoordsDims>
+    void deriv(
+            ddc::ChunkSpan<double, vals_domain_type, Layout1, memory_space> const spline_eval,
+            ddc::ChunkSpan<
+                    ddc::Coordinate<CoordsDims...> const,
+                    vals_domain_type,
+                    Layout2,
+                    memory_space> const coords_eval,
```
tpadioleau, line 162: Is templating over `CoordDims` necessary ?

include/ddc/ddc.hpp
blegouix: I will try to support the case with gingko_AVAIL=false, to use DDC without splines

tpadioleau: Then I would advise a name related to splines. I would also suggest to move source code out of this header. I think it is better that it remains only a list of private headers.

examples/characteristics_advection.cpp
blegouix: I like to see distorsion appearing on the discontinuity and integral staying unchanged despite of it, but as you prefer

include/ddc/kernels/splines/matrix_maker.hpp
blegouix: I can't do this without making the different instances of sll diverge (in Gysela and in splines branch). Let's keep this change for later.

include/ddc/kernels/splines/spline_evaluator_batched.hpp
blegouix: I would not say necessary, but I do not see reasons to restrict it to `tag_type ` and I can imagine situations where it would avoid a ddc::select in the call.

include/ddc/kernels/splines/matrix_sparse.hpp
blegouix: I realized Ginkgo works in Serial with a multiple-rhs call in OpenMP. Not sure why however. So calling Ginkgo in parallel region seems necessary in OpenMP configuration to benefit from parallelism. GPU is different.

I just checked the tests results are good and each thread acts on a different dataset so no race condition happens. Also we can assume those basic Kokkos functions to be thread-safe, no ?

I dont think we have benefit of using OpenMP+CUDA, reason why par_chunks_per_seq_chunk=1 in CUDA configuration.

include/ddc/kernels/splines/bspline.hpp
blegouix: I can't do this without making the different instances of sll diverge (in Gysela and in splines branch). Let's keep this change for later.

CMakeLists.txt
```diff
@@ -261,6 +261,14 @@ if("${HIP_FOR_NVIDIA}")
 	endif()
 endif()
 
+# Ginkgo
+# workaround for Ginkgo+CUDA install
+find_package(CUDAToolkit REQUIRED)
+target_link_libraries(DDC INTERFACE CUDA::cublas)
```
tpadioleau, line 267: I think it needs to be inside a condition on whether CUDA has been enabled on Kokkos

include/ddc/kernels/splines/bspline.hpp
blegouix: It is used in the BSplines classes

`using continuous_dimension_type = BSpline<Tag>;`

So I cannot make such change without making different instances of sll diverge, lets keep it for later

CMakeLists.txt
blegouix: No, the point is we need it to compile on Persee when CUDA is OFF. It is a workaround related to the fact that Ginkgo in Spack is compiled with CUDA

tpadioleau: Yes you are right, I can deploy an omp only version. That way if we want to test without CUDA, we activate this new environment

CMakeLists.txt
```diff
@@ -261,6 +261,14 @@ if("${HIP_FOR_NVIDIA}")
...
+# end of workaround
```
tpadioleau, line 268:
```suggestion
```

include/ddc/ddc.hpp
blegouix: commit 78b7c484c3f10d8997ee02101d0e28b1a7ac8bf6 introduces cmake options to disable FFT or splines tests/examples/benchmarks. It also loaded CUDA toolkit, Ginkgo etc... only if required.

blegouix: Not sure where to put the code which is in ddc.hpp then... Ginkgo may not be specific to splines even if its the case atm.

blegouix: Also, this code would be removed if https://github.com/ginkgo-project/ginkgo/pull/1358 is merged (but no update for a long time)

include/ddc/kernels/splines/matrix_sparse.hpp
```diff
@@ -0,0 +1,292 @@
...
+        } else {
+#ifdef KOKKOS_ENABLE_SERIAL
+            if (std::is_same_v<ExecSpace, Kokkos::Serial>) {
+                m_par_chunks_per_seq_chunk = 1;
+            }
+#endif
+#ifdef KOKKOS_ENABLE_OPENMP
+            if (std::is_same_v<ExecSpace, Kokkos::OpenMP>) {
+                // TODO: Investigate OpenMP parallelism in Ginkgo
+                m_par_chunks_per_seq_chunk = ExecSpace::concurrency();
```
tpadioleau, line 92: It does not compile, `concurrency` is not a static method

include/ddc/kernels/splines/math_tools.hpp
```diff
@@ -0,0 +1,111 @@
...
+    return val;
+}
+
+template <typename T>
+inline T modulo(T x, T y)
+{
+    return x - y * std::floor(double(x) / y);
+}
+
+inline double ipow(double a, std::size_t i)
```
tpadioleau, line 56:
```suggestion
KOKKOS_INLINE_FUNCTION double ipow(double a, std::size_t i)
```

include/ddc/kernels/splines/math_tools.hpp
```diff
@@ -0,0 +1,111 @@
...
+{
+    double r(1.0);
+    for (std::size_t j(0); j < i; ++j) {
+        r *= a;
+    }
+    return r;
+}
+
+inline double ipow(double a, int i)
```
tpadioleau, line 65:
```suggestion
KOKKOS_INLINE_FUNCTION double ipow(double a, int i)
```

include/ddc/kernels/splines/matrix_sparse.hpp
```diff
@@ -0,0 +1,292 @@
...
+                                    m_n,
+                                    n_equations_in_par_chunk,
+                                    gko_exec);
+                            // Create the solver TODO: pass in constructor ?
+                            std::shared_ptr<gko::log::Stream<>> stream_logger = gko::log::Stream<>::
+                                    create(gko::log::Logger::all_events_mask
+                                                   ^ gko::log::Logger::linop_factory_events_mask
+                                                   ^ gko::log::Logger::
+                                                           polymorphic_object_events_mask,
+                                           std::cout);
```
tpadioleau, line 258: Seems to be an unused variable
```suggestion
```

include/ddc/kernels/splines/matrix_sparse.hpp
```diff
@@ -0,0 +1,292 @@
...
+                            // gko_exec->add_logger(stream_logger);
+                            std::shared_ptr<gko::stop::ResidualNorm<>::Factory> residual_criterion
+                                    = gko::stop::ResidualNorm<>::build()
+                                              .with_reduction_factor(1e-20)
+                                              .on(gko_exec);
+                            std::shared_ptr<const gko::log::Convergence<>> convergence_logger
+                                    = gko::log::Convergence<>::create(gko_exec);
```
tpadioleau, line 265: Seems to be an unused variable
```suggestion
```

examples/characteristics_advection.cpp
```diff
@@ -0,0 +1,302 @@
...
+                    DDimT::
+                            init(ddc::Coordinate<T>(start_time),
+                                 ddc::Coordinate<T>(end_time),
+                                 nb_time_steps + 1));
+    //! [time-domains]
+
+    //! [data allocation]
+    // Maps density into the full domain twice:
+    // - once for the last fully computed time-step
+    ddc::Chunk _last_density(
```
tpadioleau, line 160: last_density_alloc

examples/characteristics_advection.cpp
```diff
@@ -0,0 +1,302 @@
...
+            ddc::DiscreteDomain<DDimX, DDimY>(x_domain, y_domain),
+            ddc::DeviceAllocator<double>());
+
+    // - once for time-step being computed
+    ddc::Chunk _next_density(
```
tpadioleau, line 165: next_density_alloc

examples/characteristics_advection.cpp
```diff
@@ -0,0 +1,302 @@
...
+                double const x
+                        = ddc::coordinate(ddc::select<DDimX>(ixy));
+                double const y
+                        = ddc::coordinate(ddc::select<DDimY>(ixy));
+                initial_density(ixy) = 9.999 * Kokkos::exp(-(x * x + y * y) / 0.1 / 2);
+                // initial_density(ixy) = 9.999 * ((x * x + y * y) < 0.25);
+            });
+    //! [initial-conditions]
+
+    ddc::Chunk _host_density(
```
tpadioleau, line 188: host_density_alloc

examples/characteristics_advection.cpp
```diff
@@ -0,0 +1,302 @@
...
+                    ddc::g_null_boundary<BSplinesX>);
+    //! [instantiate solver]
+
+    //! [instantiate intermediate chunks]
+    // Instantiate chunk of spline coefs to receive output of spline_builder
+    ddc::Chunk coef_alloc(
+            spline_builder.spline_domain(),
+            ddc::KokkosAllocator<
+                    double,
+                    Kokkos::DefaultExecutionSpace::memory_space>());
```
tpadioleau, line 234: DeviceAllocator

examples/characteristics_advection.cpp
```diff
@@ -0,0 +1,302 @@
...
+    ddc::ChunkSpan coef = coef_alloc.span_view();
+
+    // Instantiate chunk to receive feet coords
+    ddc::Chunk feet_coords_alloc(
+            spline_builder.vals_domain(),
+            ddc::KokkosAllocator<
+                    ddc::Coordinate<X>,
+                    Kokkos::DefaultExecutionSpace::memory_space>());
```
tpadioleau, line 242: DeviceAllocator

examples/characteristics_advection.cpp
```diff
@@ -0,0 +1,302 @@
...
+    //! [instantiate intermediate chunks]
+
+
+    //! [time iteration]
+    for (auto const iter :
+         time_domain.remove_first(ddc::DiscreteVector<DDimT>(1))) {
+        //! [time iteration]
+
+        //! [boundary conditions]
+        //! [boundary conditions]
```
tpadioleau, line 253:
```suggestion
```

include/ddc/kernels/splines/matrix_sparse.hpp
```diff
@@ -0,0 +1,292 @@
...
+                m_par_chunks_per_seq_chunk = 1;
+            }
+#endif
+        }
+
+        if (preconditionner_max_block_size.has_value()) {
+            m_preconditionner_max_block_size = preconditionner_max_block_size.value();
+        } else {
+            m_preconditionner_max_block_size = 1u;
+        }
```
tpadioleau, line 106: The variable seems to unused

include/ddc/kernels/splines/matrix_maker.hpp
tpadioleau: I think they already diverge don't they ?

tpadioleau (commented):
Can you remove the `Kokkos_Core_fwd.hpp` header, I am not 100% sure it is a public header. Also I found some `Kokkos_Random.hpp`, it is public but does not seem to be necessary.

Also in the tests, can you only use our public header regarding splines ?

**Edit (BL): DONE**

tests/CMakeLists.txt
```diff
@@ -1,6 +1,9 @@
 # SPDX-License-Identifier: MIT
 cmake_minimum_required(VERSION 3.15)
 
+option(BUILD_FFT_KERNEL      "Build DDC kernel for FFT" ON)
+option(BUILD_SPLINES_KERNEL  "Build DDC kernel for splines" ON)
+
```
tpadioleau, line 6: Seems to be redundant with the top-level directory `CMakeLists.txt`
```suggestion
```

include/ddc/kernels/splines/matrix_maker.hpp
blegouix: Not really, for example non-periodic tests runs, PolarSplineBuilder and PolarSplineEvaluator still compile in my working branch "splines" etc. I really want to avoid breaking changes until everything have been ported to GPU. Moreover I will have to integrate the changes from Pauline, which are in Gysela/sll and it will be far easier to do so before beginning with the GPU port of polar splines.

include/ddc/kernels/splines/spline_evaluator_batched.hpp
blegouix: Ok there is a big subject underlying,

In current Gysela, we have successive calls to 1D interpolators, which call 1D SplineBuilder and 1D SplineEvaluator. So we never compute multidimensionnal splines in interpolators (we do 2D splines in poisson). However, there is another way to do it: a nD interpolator which calls nD SplineBuilder then nD SplineEvaluator. nD SplineBuilder is a succession of 1D SplineBuilder, nD SplineEvaluator is a succession of 1D SplineEvaluator. So, it just changes the order of the operations (BuilderX, BuilderY, EvaluatorX, EvaluatorY in place of BuilderX, EvaluatorX, BuilderY, EvaluatorY)

But there is an advantage to do so: by using an nD SplineBuilder you compute the nD Splines coef, which is a completly equivalent way to store the distribution function (basis change). And in this basis, we can do interesting things like integration on the whole space (simple transform_reduce), finite elements etc...

It may permit to mutualize operations in favorable cases, to "share" a basis change between two operators (example : you can do BuilderVx, BuilderVy, ChargeDensityComputation (integration on velocities), EvaluatorVx, EvaluatorVy to mutualize VelocityAdvection and ChargeDensityComputation). We can do even better things if we have schemes which permit to call operators in parallel (Kevin tried to perform advection and poisson in parallel in Gysela fortran).

However in the case of this question of templating over CoordsDims... it just a matter of getting a ddc::select in the call of the evaluator or not... Not a big deal, I can change it and template with tag_type if you want.

tests/CMakeLists.txt
blegouix: Ah, I don't think so, my understanding is that options are in the scope of current CMakeLists and this is the way to transmit it to tests/CMakeLists. I checked the behavior which seems correct

include/ddc/kernels/splines/spline_evaluator_batched.hpp
blegouix: I realize I may be a out of subject here, i'm ok with you whatever the dimensionnality of the SplineEvaluator is we can impose `ddc::Coordinate<dimensions of interest...>`, so here `ddc::Coordinate<tag_type>`

blegouix: In tests/batched_splines_evaluator.cpp, I have the following code:

`// Instantiate chunk of coordinates of dom_interpolation
    ddc::Chunk coords_eval_alloc(dom_vals, ddc::KokkosAllocator<Coord<X...>, MemorySpace>());
    ddc::ChunkSpan coords_eval = coords_eval_alloc.span_view();
    ddc::for_each(
            ddc::policies::policy(exec_space),
            coords_eval.domain(),
            DDC_LAMBDA(Index<IDim<X, I>...> const e) { coords_eval(e) = ddc::coordinate(e); });

[...]

    // Call spline_evaluator on the same mesh we started with
    spline_evaluator_batched(spline_eval, coords_eval.span_cview(), coef.span_cview());
    `

Sure, I could write `coords_eval(e) = ddc::coordinate(ddc::select<IDim<I,I>(e));` but in pratical cases like the one I described in my long message, we would like to have an array of full coordinates (so I was not so out of topic after all).

And, at this point (this is related to discussion on Vector Fields), we have no tool to project a Chunk<Coordinates<X...>> so we cannot do the call if we remove the templating on CoordsDims...

This is more than just a `ddc::select` in the call after all.

include/ddc/kernels/splines/bspline.hpp
tpadioleau: ok for me

include/ddc/kernels/splines/spline_evaluator_batched.hpp
tpadioleau: Ok let's accept the feature though I am still not yet really convinced we "need" it, not clear yet to me if we will have AoS or SoA data structures.

tpadioleau (commented):
Can you run clang-format ?

tpadioleau:
Also please move the source code in ddc/ddc.hpp to an other header.

tpadioleau:
@blegouix can you sync with the main branch ?

tpadioleau (approved):

PR #204, opened 2023-10-20
remove tests/fft.hpp to align fft and futur splines tests

blegouix:
Small standardization


tpadioleau (commented):
Do not forget to also `git rm` the `fft.hpp` file

tpadioleau (approved):
Looks good to me

PR #207, opened 2023-10-25
Second implementation of the combination of DiscreteElements

tpadioleau:


jbigot (dismissed):

tpadioleau:
@jbigot I also want to update the `TaggedVector` version of the `take` function. I wonder if we should allow any combination of `ElementType` in objects passed to `take` or should we constrain it on having the same ?
Should we allow `take<CDimX>(TaggedVector<float, CDimX>, TaggedVector<double, CDimY>)` ? I do not see a good reason to disallow it.

jbigot:
> @jbigot I also want to update the `TaggedVector` version of the `take` function. I wonder if we should allow any combination of `ElementType` in objects passed to `take` or should we constrain it on having the same ? Should we allow `take<CDimX>(TaggedVector<float, CDimX>, TaggedVector<double, CDimY>)` ? I do not see a good reason to disallow it.

I agree

PR #208, opened 2023-10-25
Add allocate function that takes a std::string

tpadioleau:


examples/heat_equation.cpp
```diff
@@ -199,13 +199,15 @@ int main(int argc, char** argv)
     // Maps temperature into the full domain (including ghosts) twice:
     // - once for the last fully computed time-step
     ddc::Chunk ghosted_last_temp(
+            "ghosted_last_temp",
```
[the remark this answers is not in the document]
tpadioleau: good idea

tpadioleau:
The current template argument `Allocator` in `Chunk` requires more features, linked to Kokkos memory spaces, than a standard allocator. It shall be possible to adapt `Chunk` for both types of allocators but I suggest to do that later.

tpadioleau:
Using the Kokkos Tool `libkp_memory_events.so` I get on the heat equation example:
```bash
# Memory Events
# Time     Ptr                  Size        MemSpace      Op         Name
0.000040        0x13192c0           9792             Host Allocate   ghosted_last_temp
0.000051        0x131bb40           9792             Host Allocate   ghosted_next_temp
0.000511        0x131e580           9792             Host Allocate   ghost_temp
0.465690        0x131e580          -9792             Host DeAllocate ghost_temp
0.465703        0x13192c0          -9792             Host DeAllocate ghosted_last_temp
0.465710        0x131bb40          -9792             Host DeAllocate ghosted_next_temp
```

include/ddc/kokkos_allocator.hpp
```diff
@@ -50,6 +51,11 @@ class KokkosAllocator
         return static_cast<T*>(Kokkos::kokkos_malloc<MemorySpace>(sizeof(T) * n));
     }
 
+    [[nodiscard]] T* allocate(std::string const& label, std::size_t n) const
+    {
+        return static_cast<T*>(Kokkos::kokkos_malloc<MemorySpace>(label, sizeof(T) * n));
```
jbigot, line 56: shouldn't we use the lower-level `MemorySpace::allocate`?

include/ddc/chunk.hpp
```diff
@@ -131,11 +143,11 @@ class Chunk<ElementType, DiscreteDomain<DDims...>, Allocator>
             return *this;
         }
         if (this->m_internal_mdspan.data_handle()) {
-            std::allocator_traits<
-                    Allocator>::deallocate(m_allocator, this->data_handle(), this->size());
+            m_allocator.deallocate(this->data_handle(), this->size());
         }
         static_cast<base_type&>(*this) = std::move(static_cast<base_type&>(other));
         m_allocator = std::move(other.m_allocator);
+        m_label = std::move(other.m_label);
```
jbigot, line 150: can we not retrieve the label already stored in the allocation?

include/ddc/kokkos_allocator.hpp
tpadioleau: Maybe in the end. What would be the decision criterion ? The fact that we don't take advantage of the memory tracking ?

include/ddc/chunk.hpp
tpadioleau: I cannot find such a public function, all I know is here: https://kokkos.github.io/kokkos-core-wiki/API/core/c_style_memory_management/malloc.html?highlight=kokkos_malloc

include/ddc/kokkos_allocator.hpp
jbigot: exactly

include/ddc/chunk.hpp
jbigot: It might not be stored at all if we switch to allocate, maybe that's the best option.

include/ddc/kokkos_allocator.hpp
jbigot: ok :D let's go with kokkos_malloc then!

jbigot (approved):

jbigot:
Looks great!

PR #213, opened 2023-10-27
Implement correct CTAD

tpadioleau:


tpadioleau:
I notice that it is difficult to make CTAD work correctly when a variadic template class that has variadic template constructors. I suggest we do not support CTAD for those classes.

Issue #77, opened 2022-08-04
Use a signed integer type for DiscreteElementType


tpadioleau:
With the recent possibility for mdspan to choose the `index_type`, I think we should a signed integer type common for `DiscreteElementType`, `std::extents` and iterators difference_type that is to say `std::ptrdiff_t`.

tpadioleau:
Re-enable `-Wsign-compare` in the CI

EmilyBourne:
A use case for this is when trying to find a sub-set of a domain. The intuitive way of restricting an index to the domain is:
```cpp
            first_lagrange_knot = closest_knot + step;
            if (first_lagrange_knot < first_possible) {
                first_lagrange_knot = first_possible;
            } elif (first_lagrange_knot > last_possible) {
                first_lagrange_knot = last_possible;
            }
```
but as `closest_knot + step` may be negative leading to underflow this must be written as:
```cpp
            if ((first_possible - closest_knot) > step) {
                // if first_lagrange_knot would be < 0
                first_lagrange_knot = first_possible;
            } else {
                first_lagrange_knot = closest_knot + step;
                if (first_lagrange_knot > last_possible) {
                    first_lagrange_knot = last_possible;
                }
            }
```

PR #218, opened 2023-11-08
fix warnings

blegouix:


tpadioleau:
As discussed on slack, we will go for #77

PR #219, opened 2023-11-08
Move HIP LD_LIBRARY_PATH to Docker build

jbigot:


tpadioleau (approved):
I hope we can get rid of this soon. I don't understand why we need this extra LD_LIBRARY_PATH.

jbigot:
I'm affraid we're hidding a bug on our or ginko's side. In any case, any ubuntu user will face the same issue...

tpadioleau:
That is also my opinion. I do not see how it can be from our side though ? We do not explicitly depend on it

PR #235, opened 2023-11-20
Fix a dangling pointer expose to PDI

jbigot:
Also use PDI in heat_equation to test it

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -1,7 +1,10 @@
 # SPDX-License-Identifier: MIT
 
 add_executable(heat_equation heat_equation.cpp)
-target_link_libraries(heat_equation PUBLIC DDC::DDC DDC::PDI_Wrapper)
+target_link_libraries(heat_equation PUBLIC DDC::DDC)
+if("${DDC_BUILD_PDI_WRAPPER}")
+  target_link_libraries(heat_equation PUBLIC DDC::PDI_Wrapper)
+endif()
 if("${BUILD_FFT_KERNEL}")
   add_executable(heat_equation_spectral heat_equation_spectral.cpp)
   target_link_libraries(heat_equation_spectral PUBLIC DDC::DDC)


tpadioleau:
Is this ready ?

jbigot:
It's better than what we had before, but does not cleanly solve the issue. I'd merge as there is no downside, but keep the issue open

tpadioleau:
I can merge but what would be the remaining issue ? The issue was about dangling pointers and this seems to fix it ?
We may have one remaining issue about the second `with` function for Arithmetic types. If a rvalue is passed, i think we end up in the same situation.

include/ddc/pdi.hpp
```diff
@@ -41,11 +47,17 @@ class PdiEvent
         static_assert(
                 !(access & PDI_IN) || (chunk_default_access_v<BorrowedChunk> & PDI_IN),
                 "Invalid access for constant data");
+        auto allocator = std::pmr::polymorphic_allocator<size_t>(&m_metadata);
         auto extents = detail::array(data.domain().extents());
-        size_t rank = extents.size();
+        size_t& rank = *allocator.allocate(1);
+        rank = data.domain().extents().size();
```
tpadioleau, line 53:
```suggestion
        rank = extents.size();
```

include/ddc/pdi.hpp
```diff
@@ -41,11 +47,17 @@ class PdiEvent
...
         PDI_share((name + "_rank").c_str(), &rank, PDI_OUT);
         m_names.push_back(name + "_rank");
-        PDI_share((name + "_extents").c_str(), extents.data(), PDI_OUT);
+        PDI_share(
+                (name + "_extents").c_str(),
+                std::vector<size_t, decltype(allocator)>(extents.begin(), extents.end(), allocator)
```
tpadioleau, line 58:
```suggestion
                std::pmr::vector<size_t>(extents.begin(), extents.end(), allocator)
```

examples/CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 # SPDX-License-Identifier: MIT
 
 add_executable(heat_equation heat_equation.cpp)
-target_link_libraries(heat_equation PUBLIC DDC::DDC)
+target_link_libraries(heat_equation PUBLIC DDC::DDC DDC::PDI_Wrapper)
```
tpadioleau, line 4: Shall we disable this example if `PDI_Wrapper` is not enabled ?

examples/heat_equation.cpp
```diff
@@ -65,13 +65,19 @@ void display(double time, ChunkType temp)
...
+            .and_with("mean_temp", mean_temp)
+            .and_with("temp_slice", temp_slice);
 }
 //! [display]
 
 
 //! [main-start]
 int main(int argc, char** argv)
 {
+    PDI_init(PC_parse_string(""));
```
tpadioleau, line 80: Does it leak when the string is empty ?

jbigot: Yes, but is a leak that happens only once really a leak? :)

examples/CMakeLists.txt
jbigot: True... Maybe `#ifdef` the PDI part

examples/heat_equation.cpp
tpadioleau: I agree maybe not a true leak but still, it triggers the sanitizers. For this reason I would be in favor of calling the destructor of the object. It also shows the correct usage of the paraconf library.

examples/CMakeLists.txt
tpadioleau: Can you also add a if in the CMakeLists.txt

tpadioleau (approved):
At some point we should also add a guard about the `Arithmetic` case. It is not critical because a `PdiEvent` object is usually being destroyed after all the temporaries that are passed to the `with`/`and_with` functions.

PR #237, opened 2023-11-21
Fix some warnings, keeping sign-compare warnings

tpadioleau:


tpadioleau:
A bit of cleaning in the warnings:
- regarding unused variables: I either deleted some, put the attribute `[[maybe_unused]]` or used some,
- regarding the sign conversion: I suggest to disable the warning in the CI as long as the issue #77 has not been completed,
- there are still warnings for HIP, I can have a look if the upgrade of Kokkos helps, if it does we wait for #238, if not we remove the warning.

PR #239, opened 2023-11-23
Splines tuning cpu

blegouix:
Re-tune splines to optimize performance for Release mode


tpadioleau (approved):
I have no good reason to keep or change these values but I think there is no universal good value for all architecture unless we have some theoretical justification.

blegouix:
It's true, it is just for conveniance when running on our machine at this point. 32u preconditionner is the highest possible value though.

PR #240, opened 2023-11-27
Additional comments and changed variable names in heat equation example

alex-m-h:
Some comments were added based on answers from Thomas and two variable names were changed for clarity in the heat equation example.

### DIFF
--- a/examples/heat_equation.cpp
+++ b/examples/heat_equation.cpp
@@ -45,7 +45,7 @@ using DDimT = ddc::UniformPointSampling<T>;
 template <class ChunkType>
 void display(double time, ChunkType temp)
 {
-    // For a chunk, the ()-operator is used to access elements with a DiscreteElement as
+    // For `Chunk`/`ChunkSpan`, the ()-operator is used to access stored values with a single `DiscreteElement` as
     // input. So it is used here as a function that maps indices of the temperature domain
     // to the temperature value at that point
     double const mean_temp = ddc::transform_reduce(
--- a/examples/heat_equation.cpp
+++ b/examples/heat_equation.cpp
@@ -264,7 +264,7 @@ int main(int argc, char** argv)
     display(ddc::coordinate(time_domain.front()),
             ghosted_temp[x_domain][y_domain]);
     // time of the iteration where the last output happened
-    ddc::DiscreteElement<DDimT> index_time_last_output
+    ddc::DiscreteElement<DDimT> last_output_iter
             = time_domain.front();
     //! [initial output]
 
@@ -333,8 +333,8 @@ int main(int argc, char** argv)
         //! [numerical scheme]
 
         //! [output]
-        if (iter - index_time_last_output >= t_output_period) {
-            index_time_last_output = iter;
+        if (iter - last_output_iter >= t_output_period) {
+            last_output_iter = iter;
             ddc::deepcopy(ghosted_temp, ghosted_last_temp);
             display(ddc::coordinate(iter),
                     ghosted_temp[x_domain][y_domain]);
@@ -348,7 +348,7 @@ int main(int argc, char** argv)
     }
 
     //! [final output]
-    if (index_time_last_output < time_domain.back()) {
+    if (last_output_iter < time_domain.back()) {
         ddc::deepcopy(ghosted_temp, ghosted_last_temp);
         display(ddc::coordinate(time_domain.back()),
                 ghosted_temp[x_domain][y_domain]);
--- a/examples/heat_equation.cpp
+++ b/examples/heat_equation.cpp
@@ -12,7 +12,7 @@
 
 //! [X-dimension]
 /// Our first continuous dimension
-struct X; 
+struct X;
 //! [X-dimension]
 
 //! [X-discretization]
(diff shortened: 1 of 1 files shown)


examples/heat_equation.cpp
```diff
@@ -191,11 +196,14 @@ int main(int argc, char** argv)
...
 
     // number of time intervals required to reach the end time
+    // The + .2 is used to make sure it is rounded to the correct number of steps
     ddc::DiscreteVector<DDimT> const nb_time_steps {
             std::ceil((end_time - start_time) / max_dt) + .2};
     // Initialization of the global domain in time:
     // - the number of discrete time-points is equal to the number of
     //   steps + 1
+    // init takes the values to make them available globally. This is needed
+    // for init_discrete_space
```
tpadioleau, line 206:
```suggestion
    // `init` takes required information to initialize the attributes of the dimension.
```

examples/heat_equation.cpp
```diff
@@ -215,6 +223,7 @@ int main(int argc, char** argv)
             ddc::DeviceAllocator<double>());
 
     // - once for time-step being computed
+    // The DeviceAllocator is responsible for allocating memory on CPU or GPU
```
tpadioleau, line 226:
```suggestion
    // The `DeviceAllocator` is responsible for allocating memory on the default memory space.
```

examples/heat_equation.cpp
```diff
@@ -224,6 +233,8 @@ int main(int argc, char** argv)
     //! [data allocation]
 
     //! [initial-conditions]
+    // The const qualifier makes it clear that ghosted_initial_temp always references
+    // the same chunk
```
tpadioleau, line 237:
```suggestion
    // The const qualifier makes it clear that ghosted_initial_temp always references
    // the same chunk, `ghosted_last_temp` in this case
```

examples/heat_equation.cpp
```diff
@@ -254,11 +265,12 @@ int main(int argc, char** argv)
...
             ghosted_temp[x_domain][y_domain]);
     // time of the iteration where the last output happened
-    ddc::DiscreteElement<DDimT> last_output = time_domain.front();
+    ddc::DiscreteElement<DDimT> index_time_last_output
+            = time_domain.front();
     //! [initial output]
 
     //! [time iteration]
-    for (auto const iter :
+    for (auto const idxDDimT :
```
tpadioleau, line 273: What do you think of `time_delem` instead of `idxDDimT` ?

examples/heat_equation.cpp
```diff
@@ -284,7 +296,8 @@ int main(int argc, char** argv)
         ddc::ChunkSpan const next_temp {
                 ghosted_next_temp[x_domain][y_domain]};
         // a read-only view of the temperature at the previous time-step
-        ddc::ChunkSpan const last_temp {ghosted_last_temp.span_view()};
+        // span_cview makes the ChunkSpan read-only
```
tpadioleau, line 299:
```suggestion
        // span_cview returns a read-only ChunkSpan
```

examples/heat_equation.cpp
```diff
@@ -321,10 +334,10 @@ int main(int argc, char** argv)
         //! [numerical scheme]
 
         //! [output]
-        if (iter - last_output >= t_output_period) {
-            last_output = iter;
+        if (idxDDimT - index_time_last_output >= t_output_period) {
```
tpadioleau, line 337:
```suggestion
        if (time_delem - last_output_time_delem >= t_output_period) {
```

tpadioleau (commented):
@alex-m-h Does it still answer your questions with these changes ?

examples/heat_equation.cpp
alex-m-h: The `delem` suffix could be confusing. I tried to follow the naming conventions described in Emily`s tutorial. Or maybe just `time_idx`? That should be clear for everyone.

tpadioleau: `time_idx` would be fine to me but would not match the current convention. May I ask why do you think `delem` is confusing ?

jbigot: If I'm correct, you're referring to Gyselalib++ naming convention tutorial. A different naming convention is used in DDC

alex-m-h: I was at first confused for a short time what `delem` is an abbreviation for, but it is `DiscreteElement`, right? I just got confused because I think of `double` when I see a `d` due to Gysela. But yes sorry I forgot that DDC does not follow Gysela convention, of course. Where can I learn the DDC naming convention i.e. where can I find it? In the other examples `ix` is used for a `ddc::DiscreteElement` in x-direction e.g. Is this the DDC convention?

tpadioleau: I don't think you need to bother learning DDC conventions, they are more for DDC developpers.

About `ix` you are right, we are currently not consistent with our new convention.

In the end, if we change `iter` there will still be inconsistencies with other discrete elements, what do you think we keep it this way and open a new issue to rename them all at once ?

alex-m-h: That sounds great. Then I will leave `iter` as it was.

examples/heat_equation.cpp
```diff
@@ -321,10 +333,10 @@ int main(int argc, char** argv)
...
```
tpadioleau, line 336:
```suggestion
        if (iter - last_output_iter >= t_output_period) {
```

tpadioleau (approved):

PR #250, opened 2023-12-10
Non-periodic splines

blegouix:
Implementation & tests.


include/ddc/kernels/splines/bsplines_non_uniform.hpp
```diff
@@ -449,7 +449,10 @@ KOKKOS_INLINE_FUNCTION ddc::DiscreteElement<NonUniformBSplines<Tag, D>> NonUnifo
                 d += a(k, s2) * ndu(pk, r);
             }
             derivs(r, k) = d;
-            std::swap(s1, s2);
+            // swap s1 <-> s2;
+            auto tmp = s1;
+            s1 = s2;
+            s2 = tmp;
```
tpadioleau, line 455: It looks like a manual inline of `std::swap`, why do you need that ?

include/ddc/kernels/splines/spline_builder_batched.hpp
```diff
@@ -68,13 +76,7 @@ class SplineBuilderBatched
...
                 par_chunks_per_seq_chunk,
                 preconditionner_max_block_size)
-        , m_vals_domain(vals_domain)
-    {
-        static_assert(
-                BcXmin == BoundCond::PERIODIC && BcXmax == BoundCond::PERIODIC,
-                "Boundary conditions other than PERIODIC are not supported yet in "
-                "SpSplineBuilderBatched");
-    };
+        , m_vals_domain(vals_domain) {};
```
tpadioleau, line 79:
```suggestion
        , m_vals_domain(vals_domain) {}
```

tpadioleau (commented):
A short pre-review, can you also update Kokkos

include/ddc/kernels/splines/bsplines_non_uniform.hpp
blegouix: There is a warning because std::swap is host function called from KOKKOS_FUNCTION

tpadioleau: I see, indeed `std::swap` is not constexpr until C++20

include/ddc/kernels/splines/matrix_sparse.hpp
```diff
@@ -174,7 +174,7 @@ class Matrix_Sparse : public Matrix
             gko_exec = create_gko_exec<ExecSpace>();
         }
         std::shared_ptr<gko::stop::ResidualNorm<double>::Factory> residual_criterion
-                = gko::stop::ResidualNorm<double>::build().with_reduction_factor(1e-20).on(
+                = gko::stop::ResidualNorm<double>::build().with_reduction_factor(1e-17).on(
```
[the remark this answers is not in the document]
tpadioleau: I think a relative error near 1E-16 should be fine indeed

tests/splines/batched_spline_builder.cpp
```diff
@@ -142,13 +181,23 @@ struct DimsInitializer<IDimI, ddc::detail::TypeSeq<IDimX...>>
         ddc::init_discrete_space<IDimI>(
                 GrevillePoints<
                         BSplines<typename IDimI::continuous_dimension_type>>::get_sampling());
+#if defined(BC_HERMITE)
+        ddc::init_discrete_space(
+                IDimDeriv<ddc::Deriv<typename IDimI::continuous_dimension_type>>::
+                        init(Coord<ddc::Deriv<typename IDimI::continuous_dimension_type>>(1),
+                             Coord<ddc::Deriv<typename IDimI::continuous_dimension_type>>(
+                                     std::max(2, (int)s_degree_x / 2)),
```
blegouix, line 189: This is necessary because of the following static_assert:
https://github.com/CExA-project/ddc/blob/0326f51fccc7be0a8eec1ffca1082c49d6ae5eca/include/ddc/uniform_point_sampling.hpp#L117

Maybe we could relax that to a<=b ? This is a practical use-case where a=b get sense.

blegouix:
I am not sure what to do with the failing clang-tidy ci test

tpadioleau:
> I am not sure what to do with the failing clang-tidy ci test

I think the problem comes from the fact that `non_periodic_spline_builder.cpp` is not compiled. As soon as you add in the CMakeLists.txt you should be fine.

tests/splines/batched_spline_builder.cpp
```diff
@@ -68,6 +101,12 @@ using IDim = std::conditional_t<
         typename GrevillePoints<BSplines<X>>::interpolation_mesh_type,
         ddc::NonUniformPointSampling<X>>;
 #endif
+
+#if defined(BC_HERMITE)
+template <typename DerivI>
+using IDimDeriv = ddc::UniformPointSampling<DerivI>;
```
tpadioleau, line 107: I would rather use a simple `class DDimDeriv;`. I do not think we need a continuous dimension & discretization for the derivatives. I think only a discrete dimension is enough.

tests/splines/batched_spline_builder.cpp
tpadioleau: My answer for that depends on my other comment.

tests/splines/batched_spline_builder.cpp
blegouix: Yes that's right! we don't need nth-order derivative with n non-integer

include/ddc/kernels/splines/spline_builder_batched.hpp
```diff
@@ -68,11 +76,7 @@ class SplineBuilderBatched
                 preconditionner_max_block_size)
         , m_vals_domain(vals_domain)
     {
-        static_assert(
-                BcXmin == BoundCond::PERIODIC && BcXmax == BoundCond::PERIODIC,
-                "Boundary conditions other than PERIODIC are not supported yet in "
-                "SpSplineBuilderBatched");
```
EmilyBourne, line 74: It might be a good idea to keep a sanity check:
```
        static_assert(
                ((BcXmin == BoundCond::PERIODIC) == (BcXmax == BoundCond::PERIODIC)),
                "Incompatible boundary conditions in SpSplineBuilderBatched");
```

include/ddc/kernels/splines/spline_builder_batched.hpp
```diff
@@ -124,16 +128,58 @@ class SplineBuilderBatched
...
             ddc::ChunkSpan<double, spline_domain_type, Layout, memory_space> spline,
-            ddc::ChunkSpan<double, vals_domain_type, Layout, memory_space> vals) const;
+            ddc::ChunkSpan<double, vals_domain_type, Layout, memory_space> vals,
+            std::optional<ddc::ChunkSpan<double, derivs_domain_type, Layout, memory_space>> const
+                    derivs_xmin
+            = std::nullopt,
+            std::optional<ddc::ChunkSpan<double, derivs_domain_type, Layout, memory_space>> const
+                    derivs_xmax
+            = std::nullopt) const;
 };
```
EmilyBourne, line 138: For DDC I wonder if it would make more sense to provide different operators depending on the boundary conditions? This would mean that the boundary conditions need to be a template parameter but would avoid any overhead associated with `std::optional`

EmilyBourne: That said you would have 4 interfaces possible here so maybe this just gives unnecessary complexity? An optional array is not necessarily much more costly than an array

include/ddc/kernels/splines/spline_builder.hpp
```diff
@@ -81,12 +81,13 @@ class SplineBuilder
     // interpolator specific
     std::unique_ptr<ddc::detail::Matrix> matrix;
 
+    double m_dx; // average cell size for normalization of derivatives
```
EmilyBourne, line 84: Non-const variable shouldn't be public. This variable can be const.
```suggestion
    double const m_dx; // average cell size for normalization of derivatives
```
But I wonder if it would make sense for `SplineBuilderBatched` to inherit from this class rather than storing an instance?

include/ddc/kernels/splines/spline_builder_batched.hpp
```diff
@@ -124,16 +128,58 @@ class SplineBuilderBatched
...
+           != (!derivs_xmin.has_value() || derivs_xmin->template extent<deriv_type>() == 0));
+    assert((BcXmax == ddc::BoundCond::HERMITE)
+           != (!derivs_xmax.has_value() || derivs_xmax->template extent<deriv_type>() == 0));
+
+    // Hermite boundary conditions at xmin, if any
+    // NOTE: For consistency with the linear system, the i-th derivative
+    //       provided by the user must be multiplied by dx^i
+    if constexpr (BcXmin == BoundCond::HERMITE) {
+        std::cout << derivs_xmin->template extent<deriv_type>() << "\n";
+        std::cout << nbc_xmin << "\n";
```
EmilyBourne, line 169: `cout` in non-draft PR

include/ddc/kernels/splines/spline_builder.hpp
blegouix: SplineBuilder will disappear from ddc at some point. I just wait the moment when everything related to splines will be ddc-based to do so (avoiding severe break in the middle of the GPU port)

include/ddc/kernels/splines/spline_builder_batched.hpp
blegouix: I think this is the kind of very negligeable optimization which degrades readability and maintanability so I don't like it

include/ddc/kernels/splines/spline_builder_batched.hpp
```diff
@@ -124,16 +128,58 @@ class SplineBuilderBatched
...
+        assert(derivs_xmin->template extent<deriv_type>() == nbc_xmin);
+        auto const dx_proxy = spline_builder.m_dx;
+        ddc::for_each(
+                ddc::policies::policy(exec_space()),
+                batch_domain(),
+                DDC_LAMBDA(typename batch_domain_type::discrete_element_type j) {
+                    for (int i = nbc_xmin; i > 0; --i) {
+                        spline(ddc::DiscreteElement<bsplines_type>(nbc_xmin - i), j)
+                                = (*derivs_xmin)(ddc::DiscreteElement<deriv_type>(i - 1), j)
```
EmilyBourne, line 178: Should the `derivs_xmin` be saved to a local variable before the for_each to avoid incurring the cost of de-referencing the optional object inside the loop? Or do you expect the compiler to be smart enough to manage this itself?

include/ddc/kernels/splines/spline_builder_batched.hpp
blegouix, line 169:
```suggestion
```

tests/splines/batched_spline_builder.cpp
```diff
@@ -68,6 +101,12 @@ using IDim = std::conditional_t<
         typename GrevillePoints<BSplines<X>>::interpolation_mesh_type,
         ddc::NonUniformPointSampling<X>>;
```
EmilyBourne, line 102:
```
#elif defined(BSPLINES_TYPE_NON_UNIFORM)
template <typename X>
using BSplines = ddc::NonUniformBSplines<X, s_degree_x>;
template <typename X, typename I>
using IDim = ddc::NonUniformPointSampling<X>;
```
If the BSplines are non-uniform then the sampling will always be non-uniform

include/ddc/kernels/splines/spline_builder_batched.hpp
blegouix: mmm that I do not know

tests/splines/batched_spline_builder.cpp
blegouix: Yes, this is not what you expect ?

include/ddc/kernels/splines/spline_builder_batched.hpp
EmilyBourne: I don't think it degrades maintainability if it's written properly. You just need to have internal functions:
- fill_xmin_derivative_lines_in_rhs
- fill_value_lines_in_rhs(spline, vals)
- fill_xmax_derivative_lines_in_rhs

So then you would have for example (template logic is pseudo-code):
```
    template <class = std::enable_if_t<PERIODIC>>
    void operator()(
        ddc::ChunkSpan<double, spline_domain_type, Layout, memory_space> spline,
        ddc::ChunkSpan<double, vals_domain_type, Layout, memory_space> vals) const {
        fill_value_lines_in_rhs(spline, vals);
    }

    template <class = std::enable_if_t<BcXmin == HERMITE && BcXmax == HERMITE>>
    void operator()(
        ddc::ChunkSpan<double, spline_domain_type, Layout, memory_space> spline,
        ddc::ChunkSpan<double, vals_domain_type, Layout, memory_space> vals) const {
        fill_xmin_derivative_lines_in_rhs(spline, vals);
        fill_value_lines_in_rhs(spline, vals);
        fill_xmax_derivative_lines_in_rhs(spline, vals);
    }
```

You are probably right that this is slightly less readable than the version with `static_assert` but I think it should make the user code more readable as you no longer have to explicitly mention optional in lines like this:
https://github.com/CExA-project/ddc/blob/b0ee2a5cf72d08b5ea22d444f84840f5a930424b/tests/splines/batched_spline_builder.cpp#L309

tpadioleau: I kind of agree with Emily, it would make sense. Unfortunately i am afraid it could make the usage of it more complicated. I think it would require sfinae `if constexpr` if the user wants to switch boundary conditions depending on the context.

include/ddc/kernels/splines/spline_builder.hpp
tpadioleau: Actually i would prefer non const variables inside classes even if they don't get modified after initialization. The reason is that they don't get moved correctly if they are const. I agree that for a scalar it does not make a difference but then we end up with a lot of rules...

include/ddc/kernels/splines/spline_builder_batched.hpp
tpadioleau: Yes maybe avoid optional in a kernel. The kernel is small, i do not expect it will impact much the performance.

include/ddc/kernels/splines/spline_builder.hpp
EmilyBourne: Even for public variables?!

tpadioleau: No i would try to make it private, is it required to be public ?

include/ddc/kernels/splines/spline_builder_batched.hpp
jbigot: wouldn't a dedicated type templated and specialized for the boundary condition make sense?

```
template <Boundary>
class BoundaryParameter;

template <Boundary>
using BoundaryParameter_t = BoundaryParameter<Boundary>::type;

template <Boundary>
using default_boundary_param = BoundaryParameter<Boundary>::default_value;

void operator()(
        ddc::ChunkSpan<double, spline_domain_type, Layout, memory_space> spline,
        ddc::ChunkSpan<double, vals_domain_type, Layout, memory_space> vals,
        BoundaryParameter_t<BcXmin> params_xmin = default_boundary_param<BcXmin>,
        BoundaryParameter_t<BcXmax> params_xmin = default_boundary_param<BcXmax>
    ) const
{
        // [...]
}
```

include/ddc/kernels/splines/spline_builder.hpp
jbigot: There's no good reason to have public variables, if you need to expose it, do it through an accessor

include/ddc/kernels/splines/spline_builder_batched.hpp
blegouix: @EmilyBourne I think you also want to handle HERMITE Xmin and GREVILLE Xmax, and vice-versa.

I propose to keep it like that and eventually improve this in an upcoming MR if you think it has to.

tests/splines/batched_spline_builder.cpp
blegouix: I don't know which comment

include/ddc/kernels/splines/spline_builder_batched.hpp
EmilyBourne: In an ideal world I think it would be nice to keep this flexibility. I can imagine it being useful for example in a radial direction to use Greville for the centre but Hermite in the wall.
However in practice I have never seen different boundary conditions used at each end of the same spline so I would be okay with a solution which doesn't keep this flexibility

blegouix: I think it requires far more work at this point to remove the flexibility than to keep it ^^, it would means removing BcXmin and BcXmax to put just Bc everywhere. Everything is already working with different boundary conditions so let's continue like that.

include/ddc/kernels/splines/spline_builder_batched.hpp
```diff
@@ -20,6 +20,8 @@ class SplineBuilderBatched
 
     using bsplines_type = typename SplineBuilder::bsplines_type;
 
+    using deriv_type = ddc::UniformPointSampling<ddc::Deriv<tag_type>>;
```
tpadioleau, line 23: I would not use `ddc::UniformPointSampling` this way, its attributes do not really make sense for derivatives. Actually I am wondering if the derivative dimension need attributes ? One attribute that could make sense is the order of the derivative, but is it worth it ?

include/ddc/kernels/splines/spline_builder_batched.hpp
```diff
@@ -124,16 +131,57 @@ class SplineBuilderBatched
...
+        auto derivs_xmin_values = *derivs_xmin;
+        auto const dx_proxy = spline_builder.dx();
+        ddc::for_each(
+                ddc::policies::policy(exec_space()),
+                batch_domain(),
+                DDC_LAMBDA(typename batch_domain_type::discrete_element_type j) {
+                    for (int i = nbc_xmin; i > 0; --i) {
+                        spline(ddc::DiscreteElement<bsplines_type>(nbc_xmin - i), j)
+                                = derivs_xmin_values(ddc::DiscreteElement<deriv_type>(i - 1), j)
+                                  * Kokkos::pow(dx_proxy, i + odd - 1);
```
tpadioleau, line 181:
```suggestion
                                  * ipow(dx_proxy, i + odd - 1);
```
For consistency with `spline_builder.hpp` maybe we should continue using `ipow` ? Same for the other loop

include/ddc/kernels/splines/spline_builder.hpp
```diff
@@ -137,6 +138,12 @@ class SplineBuilder
         return m_interpolation_domain;
     }
 
+    int dx() const noexcept
+    {
+        return m_dx;
+    }
+
+
```
tpadioleau, line 146:
```suggestion
```

include/ddc/kernels/splines/deriv.hpp
```diff
@@ -0,0 +1,7 @@
+#pragma once
+namespace ddc {
+template <class Tag>
+struct Deriv
+{
+};
+} // namespace ddc
```
tpadioleau, line 7:
```suggestion
#pragma once

namespace ddc {

template <class Tag>
struct Deriv
{
};

} // namespace ddc
```

tests/splines/batched_spline_builder.cpp
tpadioleau: About the value of `IDimDeriv`, if it needs initialization or not

tpadioleau (changes requested):
Seems almost good to me

blegouix:
I realized left_bc and right_bc are unused in evaluator, I have to restore the part of code dedicated to it

include/ddc/kernels/splines/spline_builder_batched.hpp
blegouix: `spline_builder.hpp` will be removed in close futur, ipow is host function exclusively used from CPU side in spline_builder.hpp whereas here we are in device section.

tpadioleau:
> I realized left_bc and right_bc are unused in evaluator, I have to restore the part of code dedicated to it

Maybe do an other MR for that, this one shall be only about the spline builder ?

include/ddc/kernels/splines/spline_builder_batched.hpp
blegouix: Yes I use order of derivative as attribute. I think I cannot really avoid that if I want to pass derivatives as Chunkspans in `builder/operator()` right ? ie. I may have to pass derivative of 1st order and 2nd order in X direction for all Y and Z values, in the same Chunkspan, so I need a domain associated to it.

blegouix:
In fact, I think left_bc and right_bc are quite confusing names for what they are in evaluator because it informs on how to evaluate outside of the domain (extrapolation), not really on boundary conditions. Of course the two are kind of related one to the other (ie. constant extrapolation makes sense for 0-derivative boundary condition, etc.) but they are not the same thing (and we have not "periodic extrapolation" yet).

Then yes doing it in another MR looks good to me and maybe rename with `left_extrap_rule` and `right_extrap_rule` ? @EmilyBourne what do you think of it ?

EmilyBourne:
> In fact, I think left_bc and right_bc are quite confusing names for what they are in evaluator because it informs on how to evaluate outside of the domain (extrapolation), not really on boundary conditions. Of course the two are kind of related one to the other (ie. constant extrapolation makes sense for 0-derivative boundary condition, etc.) but they are not the same thing (and we have not "periodic extrapolation" yet).
> 
> Then yes doing it in another MR looks good to me and maybe rename with `left_extrap_rule` and `right_extrap_rule` ? @EmilyBourne what do you think of it ?

That sounds like a good idea. There is always confusion about this part. I think extrapolation rule is a good suggestion for a better name :+1:

include/ddc/kernels/splines/spline_builder_batched.hpp
blegouix: I speak about `derivs_domain_type` which is build from `deriv_type` and used as fondation DiscreteDomain of the chunkspans here:

`void operator()(
            ddc::ChunkSpan<double, spline_domain_type, Layout, memory_space> spline,
            ddc::ChunkSpan<double, vals_domain_type, Layout, memory_space> vals,
            std::optional<ddc::ChunkSpan<double, derivs_domain_type, Layout, memory_space>> const
                    derivs_xmin
            = std::nullopt,
            std::optional<ddc::ChunkSpan<double, derivs_domain_type, Layout, memory_space>> const
                    derivs_xmax
            = std::nullopt) const;`

include/ddc/kernels/splines/spline_builder_batched.hpp
tpadioleau: From what I see, you made `ipow` GPU compatible

tpadioleau: I don't expect any optimization for integer in the case of `pow`.

blegouix: Yes I forgot it

include/ddc/kernels/splines/spline_builder.hpp
```diff
@@ -82,11 +82,12 @@ class SplineBuilder
     std::unique_ptr<ddc::detail::Matrix> matrix;
 
 private:
+    double m_dx; // average cell size for normalization of derivatives
+
     const int m_offset;
 
     interpolation_domain_type m_interpolation_domain;
 
-    double m_dx; // average cell size for normalization of derivatives
```
tpadioleau, line 89: Can you revert the changes ? This modifies the order of initialization of the member variables

include/ddc/kernels/splines/spline_builder_batched.hpp
```diff
@@ -124,16 +131,57 @@ class SplineBuilderBatched
...
     void operator()(
             ddc::ChunkSpan<double, spline_domain_type, Layout, memory_space> spline,
-            ddc::ChunkSpan<double, vals_domain_type, Layout, memory_space> vals) const;
+            ddc::ChunkSpan<double, vals_domain_type, Layout, memory_space> vals,
+            std::optional<ddc::ChunkSpan<double, derivs_domain_type, Layout, memory_space>> const
+                    derivs_xmin
+            = std::nullopt,
+            std::optional<ddc::ChunkSpan<double, derivs_domain_type, Layout, memory_space>> const
+                    derivs_xmax
+            = std::nullopt) const;
```
tpadioleau, line 140: What do you think of setting the derivatives as read-only ?
```suggestion
            std::optional<ddc::ChunkSpan<const double, derivs_domain_type, Layout, memory_space>> const
                    derivs_xmin
            = std::nullopt,
            std::optional<ddc::ChunkSpan<const double, derivs_domain_type, Layout, memory_space>> const
                    derivs_xmax
            = std::nullopt) const;
```

include/ddc/kernels/splines/spline_builder_batched.hpp
```diff
@@ -20,6 +20,8 @@ class SplineBuilderBatched
 
     using bsplines_type = typename SplineBuilder::bsplines_type;
 
+    using deriv_type = ddc::Deriv<tag_type>;
```
tpadioleau, line 23: You need to include the header for `Deriv`

include/ddc/kernels/splines/spline_builder_batched.hpp
blegouix: I think `vals` can also be `ChunkSpan<double const>` right ? If so it should be addressed in another MR because it requires to replace .span_view with .span_cview in several places (I think it should also break gysela).

tpadioleau: I checked, the original `sll` version is using `const`. There is no automatic sync anymore, I would do the change as soon as possible. In the meantime we do the sync we can do the modifications in Gysela. Regarding the derivatives it should not cause any trouble yet to Gysela.

blegouix: Is 44828b05c3882c325f34106209303fb19a0d03eb ok ?

tpadioleau: Yes seems good

blegouix:
@tpadioleau ready when you are for merge

tests/splines/batched_spline_builder.cpp
```diff
@@ -148,32 +179,36 @@ struct DimsInitializer<IDimI, ddc::detail::TypeSeq<IDimX...>>
...
 template <typename ExecSpace, typename MemorySpace, typename I, typename... X>
-static void BatchedPeriodicSplineTest()
+static void BatchedSplineTest()
 {
     // Instantiate execution spaces and initialize spaces
     Kokkos::DefaultHostExecutionSpace host_exec_space = Kokkos::DefaultHostExecutionSpace();
     ExecSpace exec_space = ExecSpace();
 
-    std::size_t constexpr ncells = 10;
+    std::size_t constexpr ncells = 20;
```
tpadioleau, line 188: Why increasing the number of cells ?

include/ddc/kernels/splines/spline_builder_batched.hpp
```diff
@@ -124,16 +132,65 @@ class SplineBuilderBatched
...
+    // NOTE: For consistency with the linear system, the i-th derivative
+    //       provided by the user must be multiplied by dx^i
+    if constexpr (BcXmin == BoundCond::HERMITE) {
+        assert(derivs_xmin->template extent<deriv_type>() == nbc_xmin);
+        auto derivs_xmin_values = *derivs_xmin;
+        auto const dx_proxy = spline_builder.dx();
+        ddc::for_each(
+                ddc::policies::policy(exec_space()),
+                batch_domain(),
+                DDC_LAMBDA(typename batch_domain_type::discrete_element_type j) {
```
blegouix, line 186:
```suggestion
                KOKKOS_LAMBDA(typename batch_domain_type::discrete_element_type j) {
```

include/ddc/kernels/splines/spline_builder_batched.hpp
```diff
@@ -152,6 +209,24 @@ void SplineBuilderBatched<SplineBuilder, IDimX...>::operator()(
...
+    // NOTE: For consistency with the linear system, the i-th derivative
+    //       provided by the user must be multiplied by dx^i
+    if constexpr (BcXmax == BoundCond::HERMITE) {
+        assert(derivs_xmax->template extent<deriv_type>() == nbc_xmax);
+        auto derivs_xmax_values = *derivs_xmax;
+        auto const dx_proxy = spline_builder.dx();
+        ddc::for_each(
+                ddc::policies::policy(exec_space()),
+                batch_domain(),
+                DDC_LAMBDA(typename batch_domain_type::discrete_element_type j) {
```
blegouix, line 222:
```suggestion
                KOKKOS_LAMBDA(typename batch_domain_type::discrete_element_type j) {
```

tests/splines/batched_spline_builder.cpp
blegouix, line 188:
```suggestion
    std::size_t constexpr ncells = 10;
```

tpadioleau (approved):

PR #253, opened 2024-01-04
Use 5.7.3 hip version

tpadioleau:


docker/latest/Dockerfile
```diff
@@ -24,7 +24,7 @@ RUN chmod +x /bin/bash_run \
     wget \
  && if [ "xhip" = "x${BACKEND}" ] \
   ; then echo 'deb [arch=amd64] https://repo.radeon.com/amdgpu/latest/ubuntu jammy main' > /etc/apt/sources.list.d/amdgpu.list \
-      && echo 'deb [arch=amd64] https://repo.radeon.com/rocm/apt/latest jammy main' > /etc/apt/sources.list.d/rocm.list \
+      && echo 'deb [arch=amd64] https://repo.radeon.com/rocm/apt/5.7.3 jammy main' > /etc/apt/sources.list.d/rocm.list \
```
jbigot, line 27: if the goal of this is to test the latest versions available, shouldn't we keep latest here?

jbigot (approved):
sounds great for the oldest one, not sure for the latest

docker/latest/Dockerfile
tpadioleau: I could not make it work with latest, I think it shall be discussed with Ginkgo dev.

jbigot: Sure, but we should maybe create an issue to switch back to lastest in the jammy build (actually, we should maybe switch to mantic for "latest")

tpadioleau: I thought it was latest LTS ?

jbigot: From my point of view, the best is to test with the oldest version we claim to support and the latest that's easy to install. Then it's just a question of what we define as "easy to install"

PR #255, opened 2024-01-09
Remove DDC_LAMBDA

tpadioleau:


jbigot (approved):
Sounds great

jbigot:
Should we add a "migration guide" for users?

tpadioleau:
In a changelog ? The problem is that we don't do releases yet so how to say from "now on" use `KOKKOS_LAMBDA` ? I think it would be simpler to do the changes in gyselalibxx.

jbigot:
sure, perfect :+1:

PR #266, opened 2024-01-19
Add tag support

tpadioleau:


include/ddc/non_uniform_point_sampling.hpp
```diff
@@ -17,7 +17,7 @@
 namespace ddc {
 
 /// `NonUniformPointSampling` models a non-uniform discretization of the `CDim` segment \f$[a, b]\f$.
-template <class CDim>
```
blegouix, line 20: Maybe we can add a default tag ? For users who do not need several spaces on same dimension

tpadioleau: Should be possible if needed. I don't if this MR will be accepted anyway, there is also the `v2` solution

blegouix: This one is a better solution imo, less hacky. I still think that avoiding use of global variables would be the best thing to do but I suppose it would be lot of work and you may have good reasons to not go for it.

tpadioleau: What do you mean "less hacky" ? Actually I prefer the other solution, here you would still have to use aliases because the name is too long. There is also the question of whether the dimension is identified by the tag or by both the discretization and the tag. In what I did for now, usage of both `ddc::UniformPointSampling<T, struct x_dimension>` and `ddc::NonUniformPointSampling<T, struct x_dimension>` work which is weird in my opinion. Maybe CRTP could solve this but CRTP is quite ugly...

About the usage of global variables, yes I agree but I don't know how to make it work without it.

blegouix: `struct DDimX2 : ddc::UniformPointSampling<X>
{
};`

Relies on inheritance and looks like a trick. `ddc::UniformPointSampling<CDim, Tag>` sounds more like "First template argument is the dimension, second is a label for the region we are going to mesh".

But this is kind of a detail, if `v2` works I cannot anticipate additionnal problems due to this usage. It is just that the rule "use inheritance to declare new samplings" may be surprising for new users.

blegouix: Maybe `struct DDimX2 : ddc::UniformPointSampling<X> { };` is more convenient after all, if you already have a DDimX accessible and want a DDimX2 locally in the code, you should be able to do `struct DDimX_local : DDimX {};` ?

blegouix: Maybe we can support both ?

tpadioleau:
Closing because it is not likely to be merged, see #121

PR #271, opened 2024-01-23
Splines recover transpose

blegouix:


tpadioleau (commented):
Can you try
```
m_solver->transpose()->apply(to_gko_dense(gko_exec, b_subview), to_gko_dense(gko_exec, x_subview));
```

If it does cost more it would a much simpler code

tpadioleau (approved):

PR #272, opened 2024-01-24
Splines remove _batched suffix

blegouix:


tpadioleau (commented):
Can you also rename `batched_` files in the `tests` directory along with the possible variables inside them ?

blegouix:
the `batched_` tests precisely test the batching, so I would let the prefix in the name. There are other test which ramp on spline order and so on but are non-batched. Maybe I could unify them in a single file but this is not a priority I think

tpadioleau (approved):

PR #282, opened 2024-02-12
Set required version to 0.6.0

tpadioleau:


jbigot (dismissed):
Maybe you can be slightly less strict about version and accept a range for patch versions?

tpadioleau:
Like `0.6...<0.7` ?

jbigot:
> Like `0.6...<0.7` ?

Yes, I believe this is also equivalent to just `0.6 EXACT`, no?

tpadioleau:
It does not seem to work, tried with `find_package(Ginkgo 1.7 EXACT REQUIRED)` though version `1.7.0` is available.

tpadioleau:
Cmake supports ranges only from version 3.19, Ubuntu focal seems to use 3.16.

tpadioleau:
Merging, feel free to reopen the issue if you have a better option

PR #283, opened 2024-02-12
Deleting for_each_n function

thierryantoun:


tpadioleau (changes requested):
Please also remove `for_each_n` at the end of the file. Double check with a `grep` command.

tpadioleau (changes requested):
Please also remove the doc string of the `for_each_n` function.

tpadioleau (approved):
Perfect thanks

PR #299, opened 2024-02-21
Fix excessive loop unrolling

EmilyBourne:
In the uniform bsplines there is a function with the following prototype:
```cpp
template <class Tag, std::size_t D>
template <class MemorySpace>
template <std::size_t Size>
KOKKOS_INLINE_FUNCTION ddc::DiscreteElement<UniformBSplines<Tag, D>> UniformBSplines<Tag, D>::
        Impl<MemorySpace>::eval_basis(
                std::array<double, Size>& values,
                ddc::Coordinate<Tag> const& x,
                std::size_t const deg) const
```
There is an assertion about the size:
```cpp
assert(values.size() == deg + 1);
```
followed by a loop which modifies `values`:
```cpp
for (std::size_t j = 1; j < deg + 1; ++j) {
```
When optimisations are turned on the compiler may try to unroll this loop. As `deg + 1` is not a `constexpr` the loop is unrolled beyond what is necessary leading to warnings such as:
```
/builds/gysela-developpers/gyselalibxx/vendor/ddc/include/ddc/kernels/splines/bsplines_uniform.hpp:255:19: warning: array subscript '_Type {aka const double [5]}[0]' is partly outside array bounds of 'std::array<double, 4> [1]' [-Warray-bounds]
  255 |         values[j] = saved;
```

The simple fix for this is to use the value which is known at compile time for the loop bound

### DIFF
--- a/include/ddc/kernels/splines/bsplines_uniform.hpp
+++ b/include/ddc/kernels/splines/bsplines_uniform.hpp
@@ -243,7 +243,7 @@ KOKKOS_INLINE_FUNCTION ddc::DiscreteElement<UniformBSplines<Tag, D>> UniformBSpl
     // 3. Compute values of aforementioned B-splines
     double xx, temp, saved;
     values[0] = 1.0;
-    for (std::size_t j = 1; j < deg + 1; ++j) {
+    for (std::size_t j = 1; j < Size; ++j) {
         xx = -offset;
         saved = 0.0;
         for (std::size_t r = 0; r < j; ++r) {


EmilyBourne:
> Thanks for the PR, i would be even in favor of removing the last parameter of the function. What do you think ?

I think that makes sense as long as it is done in a way that doesn't obfuscate the meaning of this parameter

EmilyBourne:
Actually this isn't possible without renaming the function. It will conflict with this:
https://github.com/CExA-project/ddc/blob/611204f1b973863229cb1b9fce77a958102078c8/include/ddc/kernels/splines/bsplines_uniform.hpp#L131-L135

I don't have any suggestions for better names and I'm not sure that we want this version to be public

tpadioleau:
What do you think if we only keep the following function
```
template <std::size_t N>
KOKKOS_INLINE_FUNCTION discrete_element_type 
eval_basis(std::array<double, N>& values, ddc::Coordinate<Tag> const& x) const
```
along with a doc like "Fills the `values` array with the b-splines evaluated at position `x`" ?

EmilyBourne:
> What do you think if we only keep the following function
> 
> ```
> template <std::size_t N>
> KOKKOS_INLINE_FUNCTION discrete_element_type 
> eval_basis(std::array<double, N>& values, ddc::Coordinate<Tag> const& x) const
> ```
> 
> along with a doc like "Fills the `values` array with the b-splines evaluated at position `x`" ?
> 
> In the first place, do you remember why we only have the version with size `D+1` in the public API ?

The version with size `D+1` is the version that is pertinent for this b-spline. If we want to make the general version public then it does not really make sense for it to be associated with a spline of a specified degree.

Another way of implementing this would be to keep only the public version and to rewrite the call in the integrals:
```cpp
        std::array<double, degree() + 2> edge_vals_ptr;
        eval_basis(edge_vals_ptr, rmin(), degree() + 1);
```
as:
```cpp
        std::array<double, degree() + 2> edge_vals_ptr;
        UniformBSplines<Tag, D+1>::Impl<MemorySpace> higher_degree_spline(m_domain);
        higher_degree_spline.eval_basis(edge_vals_ptr, rmin());
```

tpadioleau:
I see, well I do not see a better solution for now.

tpadioleau (approved):

PR #313, opened 2024-02-26
ExecSpace&& to ExecSpace const&

thierryantoun:


tpadioleau (approved):
Good catch for the fft part

PR #323, opened 2024-03-04
Add a test with 'undefined' sanitizer

tpadioleau:


.github/workflows/tests.yml
```diff
@@ -189,6 +189,66 @@ jobs:
...
+      if: needs.id_repo.outputs.in_base_repo == 'false'
+      run: |
+        docker load < latest_cpu.tar
+        rm latest_cpu.tar
+    - name: Test
+      id: test
+      run: |
+        cat<<-'UBSAN_EOF' > ubsan.supp
+          vptr:ginkgo/core/solver/solver_base.hpp
+        UBSAN_EOF
```
tpadioleau, line 217: Should be reported to Ginkgo devs.

Issue #287, opened 2024-02-13
Runtime error encountered when using `discrete_space`


tpadioleau:
See https://gitlab.maisondelasimulation.fr/gysela-developpers/gyselalibxx/-/merge_requests/379

tpadioleau:
cc @yasahi-hpc

yasahi-hpc:
Do you have a minimum reproducer?

tpadioleau:
Unfortunately no, I was able to reproduce it in a minimal example.

yasahi-hpc:
Not really. In the end, your reproducer worked without errors?

tpadioleau:
No sorry i had made a typo in my previous comment.

tpadioleau:
```
#include <ddc/ddc.hpp>

#include <gtest/gtest.h>

#include <Kokkos_Core.hpp>

namespace {

class SpeciesInformation
{
public:
    using discrete_dimension_type = SpeciesInformation;

    using discrete_element_type = ddc::DiscreteElement<SpeciesInformation>;

    using discrete_domain_type = ddc::DiscreteDomain<SpeciesInformation>;

    using discrete_vector_type = ddc::DiscreteVector<SpeciesInformation>;

public:
    template <class MemorySpace>
    class Impl
    {
        template <class OMemorySpace>
        friend class Impl;

    private:
        // charge of the particles (kinetic + adiabatic)
        ddc::Chunk<int, discrete_domain_type, ddc::KokkosAllocator<int, MemorySpace>> m_charge;

    public:
        using discrete_dimension_type = SpeciesInformation;

        template <class OMemorySpace>
        explicit Impl(Impl<OMemorySpace> const& impl) : m_charge(impl.m_charge.domain())
        {
            ddc::deepcopy(m_charge, impl.m_charge);
        }

        explicit Impl(ddc::Chunk<int, discrete_domain_type, ddc::KokkosAllocator<int, MemorySpace>>
                              charge)
            : m_charge(std::move(charge))
        {
        }

        auto charges() const
        {
            return m_charge.span_view();
        }
    };
};


} // namespace

void initialization()
{
    ddc::DiscreteDomain<SpeciesInformation> const
            dom(ddc::DiscreteElement<SpeciesInformation>(0),
                ddc::DiscreteVector<SpeciesInformation>(2));
    ddc::Chunk<int, ddc::DiscreteDomain<SpeciesInformation>> charges_host(dom);
[...]

PR #325, opened 2024-03-05
Warn about the correct usage of `discrete_space`

tpadioleau:


include/ddc/discrete_space.hpp
```diff
@@ -194,6 +194,12 @@ std::enable_if_t<2 <= sizeof...(Args), std::tuple<Args...>> init_discrete_space(
     return detail::extract_after(std::move(a), std::index_sequence_for<Args...>());
 }
 
+/**
+ * @tparam DDim a discrete dimension
+ * @return the discrete space instance associated with `DDim`.
+ * It is recommended to call this function from a `KOKKOS_FUNCTION`.
+ * Prefer `ddc::host_discrete_space` for a host-only function.
```
yasahi-hpc, line 201: This function must be called from a `KOKKOS_FUNCTION`.
Call `ddc::host_discrete_space` for a host-only function instead.

yasahi-hpc (changes requested):
If calling this function inside a host function may cause a problem, it would be better to explicitly claim that.

include/ddc/discrete_space.hpp
tpadioleau: Yes you are right. The case we encountered is not easy to explain in one sentence. Typically if one wants to retrieved variables from a discrete dimension before a kernel, that is meant to be captured by the lambda, the user "must" call `host_discrete_space`.

yasahi-hpc (approved):

yasahi-hpc:
Resolves #287

PR #274, opened 2024-01-31
Splines recover lapack backend

blegouix:

PR #276, opened 2024-02-06
Splines kokkos-kernels

blegouix:
Attempt to replace the lapack solver calls with kokkos-kernels ones (or similar)

TODO : consider replacing matrix multiplications with Kokkos-kernels routines

PR #329, opened 2024-03-06
Lapack matrices

blegouix:
Add new matrices necessary for the LAPACK backend available in #274

Matrices classes (including `matrix_sparse`) support strided multiple-rhs


include/ddc/kernels/splines/view.hpp
```diff
@@ -6,141 +6,70 @@
 
 #include <experimental/mdspan>
 
-namespace ddc::detail {
+namespace ddc {
```
blegouix, line 9: I am not sure what I should keep and remove in this file. I let you tell me if you want me to restore chunks of codes.

include/ddc/kernels/splines/matrix.hpp
```diff
@@ -103,7 +90,7 @@ class Matrix
         return bx;
     }
 
-    int get_size() const
+    int KOKKOS_INLINE_FUNCTION get_size() const
```
tpadioleau, line 93:
```suggestion
    int get_size() const
```

include/ddc/kernels/splines/matrix.hpp
```diff
@@ -89,12 +74,14 @@ class Matrix
...
-            Kokkos::View<double**, Args...> const bx) const
+    virtual ddc::DSpan2D_stride solve_transpose_inplace(ddc::DSpan2D_stride const bx) const
     {
         assert(int(bx.extent(0)) == m_n);
-        int const info = solve_inplace_method(bx.data(), 'N', bx.extent(1));
+        int const info = solve_inplace_method(
+                bx.data_handle(),
+                'T',
+                bx.extent(1),
+                std::max(bx.stride(0), bx.stride(1)));
```
tpadioleau, line 84: The value of the parameter `stride` looks suspicious to me. Shouldn't it be only `bx.stride(0)` ?

include/ddc/kernels/splines/matrix.hpp
```diff
@@ -43,44 +46,26 @@ class Matrix
...
     virtual ddc::DSpan1D solve_transpose_inplace(ddc::DSpan1D const b) const
     {
-        assert(int(b.extent(0)) == m_n);
-        int const info = solve_inplace_method(b.data_handle(), 'T', 1);
-
-        if (info < 0) {
-            std::cerr << -info << "-th argument had an illegal value" << std::endl;
-            // TODO: Add LOG_FATAL_ERROR
-        }
+        ddc::DSpan2D_left b_left(b.data_handle(), b.extent(0), 1);
```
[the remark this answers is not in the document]
tpadioleau: Does not compile or does not run correctly ? I think I agree that in this specific case it should give the same result.

tpadioleau (changes requested):
Review of `matrix.hpp`

include/ddc/kernels/splines/matrix.hpp
blegouix: The thing is `ddc::DSpan2D_stride` does not suppose a direction for the layout. In the current implementation I use a right layout for ginkgo backend (matrix_sparse) and a left layout for lapack backend (all the rest). It is not correctly asserted currently.

What do you think of changing the arguments of `solve_inplace_method` to stop using a raw pointer `double* b` and `size `and `stride` in favor of passing directly the `ddc::DSpan2D_stride` object ? I think it would simplify several things and make assertions easier to write

include/ddc/kernels/splines/matrix.hpp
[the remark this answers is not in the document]
blegouix: Runtime

image

include/ddc/kernels/splines/matrix.hpp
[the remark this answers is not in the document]
blegouix: 1. That's true, it compiles with `private`. However in other classes we rely on `protected `to let child classes access to parents members, for example MatrixPeriodicBanded accesses `m_k` from MatrixCornerBlock, and I think creating getters for everything would be excessive. So using protected here all is just to align with the rest of the Matrix classes.
2. Not sure why. I will remove it.
3. Discussion here, I found no workaround except this `reset()` : https://ddc-lib.slack.com/archives/C037DMGQWSH/p1706779201991609
4. Yes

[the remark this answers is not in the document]
tpadioleau: 1. I think it is out of the scope of the PR. The PR should introduce the matrices with the less possible changes from existing sll matrices. If a change is needed, one should do it in an other PR and I think we should go for the other direction to avoid access to parent member variables. Please revert the change.
3. Looking rapidly at the implementation of `reset` functions, you are not capturing the class anymore (and I think it is better now), so calling `deep_copy` inside the constructor is ok.

include/ddc/kernels/splines/matrix.hpp
tpadioleau: Does it mean Ginkgo and lapack do not agree on which dimension should be contiguous ?

blegouix: `Arrays are passed as pointers, not as a pointer to pointers. All the LAPACKE routines that take one or more 2D arrays as a pointer receive a single extra parameter of type int. This argument must be equal to either LAPACK_ROW_MAJOR or LAPACK_COL_MAJOR which are defined in lapacke.h, specifying whether the arrays are stored in row-major or column-major order. If a routine has multiple array inputs, they must all use the same ordering.

Note that using row-major ordering may require more memory and time than column-major ordering, because the routine must transpose the row-major order to the column-major order required by the underlying LAPACK routine.`

Ok so the choice looks available in LAPACKE, not LAPACK, so we really don't want to support RightLayout with Lapack

include/ddc/kernels/splines/matrix.hpp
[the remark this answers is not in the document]
blegouix: I think by stop using a raw pointer in `solve_inplace_method ` arguments but a mdspan the ambiguity should disappear

include/ddc/kernels/splines/matrix_banded.hpp
```diff
@@ -0,0 +1,160 @@
...
+
+        /*
+     * Given the linear system A*x=b, we assume that A is a square (n by n)
+     * with ku super-diagonals and kl sub-diagonals.
+     * All non-zero elements of A are stored in the rectangular matrix q, using
+     * the format required by DGBTRF (LAPACK): diagonals of A are rows of q.
+     * q has 2*kl rows for the subdiagonals, 1 row for the diagonal, and ku rows
+     * for the superdiagonals. (The kl additional rows are needed for pivoting.)
+     * The term A(i,j) of the full matrix is stored in q(i-j+2*kl+1,j).
+     */
```
tpadioleau, line 67:
```suggestion
        /*
         * Given the linear system A*x=b, we assume that A is a square (n by n)
         * with ku super-diagonals and kl sub-diagonals.
         * All non-zero elements of A are stored in the rectangular matrix q, using
         * the format required by DGBTRF (LAPACK): diagonals of A are rows of q.
         * q has 2*kl rows for the subdiagonals, 1 row for the diagonal, and ku rows
         * for the superdiagonals. (The kl additional rows are needed for pivoting.)
         * The term A(i,j) of the full matrix is stored in q(i-j+2*kl+1,j).
         */
```

include/ddc/kernels/splines/matrix_banded.hpp
tpadioleau, line 1: From what I understand of this matrix, it only uses the GPU as a storage place. No computation is done on the GPU right ? If it is the case, no need to store data on the GPU. Keep the template to indicate where should the input data of `solve_inplace_method` will be located but use directly host data for the lapack solver.

To summarize, the diff with the original implementation should be small, I suggest:
- replace unique_ptr with a host View or DualView
- if you use a DualView, `get_element` and `set_element` should work only on the host view
- if you use a DualView, `factorize`  will also synchronize device view
- after calling `factorize`, `get_element` and `set_element` should not be called anymore
- in `solve_inplace_method` make a host copy of `b` and use it in the lapack function

tpadioleau (changes requested):
Review of `matrix_banded.hpp`

include/ddc/kernels/splines/matrix_banded.hpp
blegouix: But I wanted Kokkos-kernel branch to be as small as possible to facilitate Yuuichi works, so it means I need again another MR to move the matrix on the GPU...? Cannot we just go simple with the three MR I have already opened ? Lapack solvers calls on CPU are just placeholders.

I don't like DualView very much as at the end we just need all the data on the GPU because that's where we will perform the operations.

Abdelhadi also have his data on the GPU and fill it directly on the GPU (because his matrix changes and he needs performance) so using DualViews would definitively break the alignment between his class and mine.

include/ddc/kernels/splines/matrix_corner_block.hpp
```diff
@@ -0,0 +1,328 @@
...
+                "calculate_delta_to_factorize",
+                Kokkos::MDRangePolicy<
+                        Kokkos::DefaultHostExecutionSpace,
+                        Kokkos::Rank<2>>({0, 0}, {m_k, m_k}),
+                [&](const int i, const int j) {
+                    double val = 0.0;
+                    for (int l = 0; l < m_nb; ++l) {
+                        val += lambda_host(i, l) * Abm_1_gamma_host(l, j);
+                    }
+                    delta_proxy.set_element(i, j, delta_proxy.get_element(i, j) - val);
```
tpadioleau, line 175: Can you comment on the thread safety of `set_element` ?

include/ddc/kernels/splines/matrix_corner_block.hpp
```diff
@@ -0,0 +1,328 @@
...
+                });
+    }
+    virtual ddc::DSpan2D_stride solve_lambda_section(
+            ddc::DSpan2D_stride const v,
+            ddc::DView2D_stride const u) const
+    {
+        Kokkos::parallel_for(
+                "solve_lambda_section",
+                Kokkos::TeamPolicy<ExecSpace>(v.extent(1), Kokkos::AUTO),
+                KOKKOS_CLASS_LAMBDA(
```
tpadioleau, line 185: Please use a `KOKKOS_LAMBDA` instead

include/ddc/kernels/splines/matrix_corner_block.hpp
```diff
@@ -0,0 +1,328 @@
...
+        return v;
+    }
+    virtual ddc::DSpan2D_stride solve_lambda_section_transpose(
+            ddc::DSpan2D_stride const u,
+            ddc::DView2D_stride const v) const
+    {
+        Kokkos::parallel_for(
+                "solve_lambda_section_transpose",
+                Kokkos::TeamPolicy<ExecSpace>(u.extent(1), Kokkos::AUTO),
+                KOKKOS_CLASS_LAMBDA(
```
tpadioleau, line 208: Please use a `KOKKOS_LAMBDA` instead

include/ddc/kernels/splines/matrix_corner_block.hpp
```diff
@@ -0,0 +1,328 @@
...
+        return u;
+    }
+    virtual ddc::DSpan2D_stride solve_gamma_section(
+            ddc::DSpan2D_stride const u,
+            ddc::DView2D_stride const v) const
+    {
+        Kokkos::parallel_for(
+                "solve_gamma_section",
+                Kokkos::TeamPolicy<ExecSpace>(u.extent(1), Kokkos::AUTO),
+                KOKKOS_CLASS_LAMBDA(
```
tpadioleau, line 231: Please use a `KOKKOS_LAMBDA` instead

include/ddc/kernels/splines/matrix_corner_block.hpp
```diff
@@ -0,0 +1,328 @@
...
+        return u;
+    }
+    virtual ddc::DSpan2D_stride solve_gamma_section_transpose(
+            ddc::DSpan2D_stride const v,
+            ddc::DView2D_stride const u) const
+    {
+        Kokkos::parallel_for(
+                "solve_gamma_section_transpose",
+                Kokkos::TeamPolicy<ExecSpace>(v.extent(1), Kokkos::AUTO),
+                KOKKOS_CLASS_LAMBDA(
```
tpadioleau, line 254: Please use a `KOKKOS_LAMBDA` instead

tpadioleau (changes requested):

include/ddc/kernels/splines/matrix_banded.hpp
blegouix: I did the change but I am still wondering why we would want the matrix being on the host, it is in contradiction with the tendency "put everything on ExecSpace" I tried to follow with splines, and I think I will have to revert the change in one month or two when the solvers will be on GPU. I am ok to do small PRs when it is possible but undo what has been done to redo it later looks waste of time to me.

include/ddc/kernels/splines/matrix.hpp
```diff
@@ -43,44 +44,22 @@ class Matrix
 
     virtual ddc::DSpan1D solve_inplace(ddc::DSpan1D const b) const
     {
-        assert(int(b.extent(0)) == m_n);
-        int const info = solve_inplace_method(b.data_handle(), 'N', 1);
-
-        if (info < 0) {
-            std::cerr << -info << "-th argument had an illegal value" << std::endl;
-            // TODO: Add LOG_FATAL_ERROR
-        }
```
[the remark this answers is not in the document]
blegouix: This function is rebased on `solve_inplace(DSpan2D_stride)`, in which the check is still present.

include/ddc/kernels/splines/matrix.hpp
```diff
@@ -123,7 +100,7 @@ class Matrix
 protected:
     virtual int factorize_method() = 0;
 
-    virtual int solve_inplace_method(double* b, char transpose, int n_equations) const = 0;
+    virtual int solve_inplace_method(ddc::DSpan2D_stride const b, char const transpose) const = 0;
```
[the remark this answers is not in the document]
blegouix: Just for my understanding of `const`s, could you explain why you want this change ?

include/ddc/kernels/splines/matrix_corner_block.hpp
blegouix: Inlining of:
`delta_proxy.set_element(i, j, delta_proxy.get_element(i, j) - val);`

Is just:
'delta_proxy.m_a(i,j) = delta_proxy.m_a(i,j) - val;`

So I think there is no problem with it as we just act on a single element ?

include/ddc/kernels/splines/matrix_banded.hpp
tpadioleau: The goal is to have a robust implementation with a fast `solve_inplace_method`. Therefore, if solving the matrix is done on the CPU, it is better to keep the matrix on the CPU rather than going back and forth between the CPU and the GPU.

When we are able to solve on the GPU, the filling of the matrix on the GPU is still questionable. It might be faster to do it in parallel on the CPU, factorize again on the CPU and then push it to the GPU. Moreover the documentation of Kokkos does not mention if calling `Kokkos::deep_copy` in parallel is allowed.

The size of the matrix is quite small in the end, having a copy of it on the CPU should not be a problem.

include/ddc/kernels/splines/matrix.hpp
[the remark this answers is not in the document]
tpadioleau: A declaration does not need to expose that the parameters passed by copy will not be modified. This `const` is not part of the function signature, see also https://clang.llvm.org/extra/clang-tidy/checks/readability/avoid-const-params-in-decls.html

include/ddc/kernels/splines/matrix_corner_block.hpp
tpadioleau: I agree now that we are not calling `deep_copy` anymore

include/ddc/kernels/splines/view.hpp
tpadioleau: If you made sure it is not used anymore in either the library or gysela, it is to me. In the future please keep this kind of change for an other MR

tests/splines/matrix.cpp
```diff
@@ -48,14 +46,14 @@ void check_inverse(ddc::DSpan2D matrix, ddc::DSpan2D inv)
         for (std::size_t j(0); j < N; ++j) {
             double id_val = 0.0;
             for (std::size_t k(0); k < N; ++k) {
-                id_val += matrix(i, k) * inv(j, k);
+                id_val += matrix(i, k) * inv(k, j);
```
tpadioleau, line 49: Was it a silent bug ?

tests/splines/matrix.cpp
```diff
@@ -76,6 +74,199 @@ class MatrixSizesFixture : public testing::TestWithParam<std::tuple<std::size_t,
...
+    matrix->factorize();
+    matrix->solve_inplace(inv);
+    check_inverse(val, inv);
+}
+
+TEST_P(MatrixSizesFixture, OffsetBanded)
+{
+    auto const [N, k] = GetParam();
+    std::unique_ptr<ddc::detail::Matrix> matrix = ddc::detail::MatrixMaker::make_new_banded<
+            Kokkos::DefaultHostExecutionSpace>(N, 0, 2 * k, true);
```
tpadioleau, line 108: Can you make it `DefaultExecutionSpace` ? Like the `Sparse` test.

include/ddc/kernels/splines/matrix_sparse.hpp
```diff
@@ -122,17 +122,20 @@ class Matrix_Sparse : public Matrix
...
+        m_matrix_dense->fill(0);
+        m_matrix_sparse->clear();
     }
 
-    virtual double get_element([[maybe_unused]] int i, [[maybe_unused]] int j) const override
+    double get_element([[maybe_unused]] int i, [[maybe_unused]] int j) const override
     {
         throw std::runtime_error("MatrixSparse::get_element() is not implemented because no API is "
                                  "provided by Ginkgo");
+        return 0;
```
tpadioleau, line 135:
```suggestion
```

harmless but not required

include/ddc/kernels/splines/matrix_sparse.hpp
```diff
@@ -122,17 +122,20 @@ class Matrix_Sparse : public Matrix
         std::shared_ptr const gko_exec = create_gko_exec<ExecSpace>();
         m_matrix_dense = gko::matrix::Dense<
                 double>::create(gko_exec->get_master(), gko::dim<2>(mat_size, mat_size));
-        m_matrix_dense->fill(0);
         m_matrix_sparse = matrix_sparse_type::create(gko_exec, gko::dim<2>(mat_size, mat_size));
+
+        m_matrix_dense->fill(0);
+        m_matrix_sparse->clear();
```
tpadioleau, line 128:
```suggestion
        m_matrix_sparse->clear();
```

Can you explain what it does and why it is required ? Looking at the doc https://ginkgo-project.github.io/ginkgo-generated-documentation/doc/develop/classgko_1_1PolymorphicObject.html#a8bb6f9995e8166099ca3ecab51d98df5 I am not sure we need to call it just after the object has been created

CMakeLists.txt
```diff
@@ -241,6 +246,10 @@ if("${DDC_BUILD_KERNELS_SPLINES}")
   find_package(Ginkgo 1.7.0 EXACT REQUIRED)
   target_link_libraries(DDC INTERFACE Ginkgo::ginkgo)
   target_compile_definitions(DDC INTERFACE ginkgo_AVAIL)
+
+  # Lapack
+  find_package(LAPACK REQUIRED COMPONENTS CXX)
```
tpadioleau, line 251:
```suggestion
  find_package(LAPACK REQUIRED)
```

We also need  to fix this in Gysela, there is no such component in `FindLAPACK.cmake`

include/ddc/kernels/splines/matrix_periodic_banded.hpp
```diff
@@ -0,0 +1,192 @@
...
+                                    Kokkos::atomic_sub(&v(i, j), lambda_device(i, l) * u(l, j));
+                                }
+                                // Lower diagonals in lambda
+                                for (int l = i + 1; l < k_proxy + 1; ++l) {
+                                    Kokkos::atomic_sub(
+                                            &v(i, j),
+                                            lambda_device(i, l) * u(nb_proxy - 1 - k_proxy + l, j));
+                                }
+                            });
+                });
```
tpadioleau, line 154: You seem to have changed your approach of parallelisation compared to `calculate_delta_to_factorize`. The approach in `calculate_delta_to_factorize` seems simpler, do you think the hierarchical parallelisation is worth it ?

tests/splines/matrix.cpp
blegouix: For example if you look there: https://github.com/CExA-project/ddc/blob/417bd9004fba6ccf1f2f35ee195666146178fe35/include/ddc/kernels/splines/matrix.hpp#L70

include/ddc/kernels/splines/matrix_sparse.hpp
blegouix: dont you think it will trigger a warning at compile time ?

include/ddc/kernels/splines/matrix_periodic_banded.hpp
blegouix: `calculate_delta_to_factorize` is perfomed once on CPU during factorization. Here we need a performant matrix manipulation performed at each solver call.

include/ddc/kernels/splines/view.hpp
blegouix: Gysela has it own view.hpp

include/ddc/kernels/splines/matrix_sparse.hpp
tpadioleau: Should not, to my knowledge compilers handle fine the fact that there is a missing return after a `throw`

include/ddc/kernels/splines/matrix_periodic_banded.hpp
tpadioleau: I agree it needs to be efficient.

How far have you been in the analysis of performance ?
- We need to know if we have coalescent accesses on GPU. Are `u` and `v` really non-contiguous ? if not, left or right ?
- Atomics have a cost, are atomics really necessary here ?

blegouix: 1) Yes u and v are really non-contiguous because rhs structure is something like this:

```
uuuuuuuuuu
uuuuuuuuuu
uuuuuuuuuu
vvvvvvvvvvvv
vvvvvvvvvvvv
```

In Left layout.

2) Looking at it again, I don't think atomics are necessary as long as we keep the loops on `l` sequential (but if we use TeamThreads/ThreadVector we can still use a scalar buffer to compute the sum on `l` before doing the substraction).
3) In my original implementation lambda was on the device. So no, I did not look at the performance impact of it since the change.

include/ddc/kernels/splines/matrix.hpp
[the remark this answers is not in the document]
blegouix: Ok I'm glad I asked x)

include/ddc/kernels/splines/matrix_sparse.hpp
```diff
@@ -122,17 +122,19 @@ class Matrix_Sparse : public Matrix
...
```
blegouix, line 128:
```suggestion
```

include/ddc/kernels/splines/matrix_periodic_banded.hpp
blegouix: Should I replace sequential for loops on `l` with ThreadVector actually ?

tpadioleau: No i think this work would require its own PR to make a study of what execution policy better suits this loop.

include/ddc/kernels/splines/matrix.hpp
```diff
@@ -47,44 +48,22 @@ class Matrix
...
-            std::cerr << -info << "-th argument had an illegal value" << std::endl;
-            // TODO: Add LOG_FATAL_ERROR
-        }
+        ddc::DSpan2D_left b_2d(b.data_handle(), b.extent(0), 1);
+        solve_transpose_inplace(b_2d);
         return b;
     }
 
-    virtual ddc::DSpan2D solve_multiple_inplace(ddc::DSpan2D const bx) const
+    virtual ddc::DSpan2D_stride solve_inplace(ddc::DSpan2D_stride const bx) const
```
EmilyBourne, line 63: What is the difference between `DSpan2D_left` and `DSpan2D_stride`?

include/ddc/kernels/splines/matrix_banded.hpp
```diff
@@ -0,0 +1,138 @@
...
+template <class ExecSpace>
+class Matrix_Banded : public Matrix
+{
+protected:
+    int const m_kl; // no. of subdiagonals
+    int const m_ku; // no. of superdiagonals
+    int const m_c; // no. of columns in q
+    Kokkos::View<int*, Kokkos::HostSpace> m_ipiv; // pivot indices
+    // TODO: double**
+    Kokkos::View<double*, Kokkos::HostSpace> m_q; // banded matrix representation
```
EmilyBourne, line 46: Is this TODO before merging?

include/ddc/kernels/splines/matrix_banded.hpp
```diff
@@ -0,0 +1,138 @@
...
+        return info;
+    }
+    int solve_inplace_method(ddc::DSpan2D_stride b, char const transpose) const override
+    {
+        assert(b.stride(0) == 1);
+        int const n_equations = b.extent(1);
+        int const stride = b.stride(1);
+
+        Kokkos::View<double**, Kokkos::LayoutStride, typename ExecSpace::memory_space>
+                b_view(b.data_handle(), Kokkos::LayoutStride(get_size(), 1, n_equations, stride));
```
EmilyBourne, line 109: Doesn't "ExecSpace" stand for "Execution Space"? I.e. the place (CPU/GPU) where the calculation will be performed? For Lapack functions you don't have a choice about this so I'm not clear on why you parametrise by this. Is it simply a roundabout way of passing the memory space?

blegouix: Because Lapack is a placeholder for Kokkos-Kernel. In the current branch we create a mirror on the CPU to call the solver but in #276 we expect to solve on GPU.

include/ddc/kernels/splines/matrix.hpp
blegouix: The same between Kokkos::LayoutLeft and Kokkos::LayoutStride. LayoutLeft is necessarily contiguous in left dimension and forbid `extent!=stride` in the right dimension.

include/ddc/kernels/splines/matrix_banded.hpp
```diff
@@ -0,0 +1,138 @@
...
+extern "C" int dgbtrf_(
+        int const* m,
+        int const* n,
+        int const* kl,
+        int const* ku,
+        double* a_b,
+        int const* lda_b,
+        int* ipiv,
+        int* info);
+extern "C" int dgbtrs_(
```
EmilyBourne, line 24: Not sure if you are aware but this only works with certain compilers. The mangled name will be different for different compilers. But maybe this is not so critical if the plan is to remove LAPACK almost immediately

blegouix: I did not know that, maybe we may wait for a problem to be reported before trying to solve it ? I mean, Lapack is indeed temporary but kokkos-kernel depends on Yuuichi's work too so I cannot tell for how long. But we are the only users of the splines kernel and I guess we will remain for a long time

EmilyBourne: In that case you may want to look into something like https://www.netlib.org/lapack/lapwrapc/

include/ddc/kernels/splines/matrix_center_block.hpp
```diff
@@ -0,0 +1,191 @@
...
+#include <string.h>
+
+#include "matrix_corner_block.hpp"
+#include "view.hpp"
+
+namespace ddc::detail {
+class Matrix;
+
+template <class ExecSpace>
+class Matrix_Center_Block : public Matrix_Corner_Block<ExecSpace>
```
EmilyBourne, line 19: In the rest of the code CamelCase is used for class names. Is there any reason that it isn't used here?

include/ddc/kernels/splines/matrix_banded.hpp
yasahi-hpc: My current target  is `tbsv`, and thus it will take a certain amount of time to work on `dgbtrf` and `dgbtrs`. If I understand correctly, this one is called only once in the initialization. Is it correct @blegouix ?

EmilyBourne: No. `dgbtrf` will only be called once in the initialisation but `dgbtrs` is often called at every time step

yasahi-hpc: OK. So, we can keep this for `dgbtrf` but not for `dgbtrs`. Anyway, it will take some time. Developing fully portable Linear algebra function is not that easy.

EmilyBourne: Ok. If there is a chance that the methods are kept long term then we definitely should not merge with the assumed name mangling `dgbtrf_`

yasahi-hpc: Makes sense

blegouix: Ah yes anyway the factorization methods are definitive

include/ddc/kernels/splines/matrix_center_block.hpp
blegouix: No, it's residual from SLL version, I will make the change

include/ddc/kernels/splines/matrix_banded.hpp
blegouix: Looking at it again I think this is better to keep it 1D and follow closely the convention described in Lapack dgbtrs
```suggestion
    Kokkos::View<double*, Kokkos::HostSpace> m_q; // banded matrix representation
```

include/ddc/kernels/splines/matrix_banded.hpp
blegouix: @tpadioleau what do we do for this ?

tpadioleau: I am fine with LAPACKE, it seems to be provided by the LAPACK package, https://netlib.org/lapack/lapacke.html

tpadioleau:
@blegouix Same question here about the branch

PR #348, opened 2024-03-21
Expose ExtrapolationRules from SplineEvaluator

### LINKED ISSUE #345: Expose ExtrapolationRules from SplineEvaluator
# Description

The `SplineBuilder` class has public static constexpr objects to access the boundary conditions that were passed as template arguments:
https://github.com/CExA-project/ddc/blob/638d138e69756a9954e692464f3b7cf19df15dcf/include/ddc/kernels/splines/spline_builder.hpp#L99-L100

However the `SplineEvaluator` does not expose the types of the ExtrapolationRules passed as template arguments. It would be useful to expose these.

# Context

I would like to create a refined `SplineEvaluator` from a template `SplineEvaluator` keeping all properties the same except the bsplines

EmilyBourne:
Homogenise names of extrapolation rules in `SplineEvaluator` and `SplineEvaluator2D`. Add public type identifiers for the extrapolation rule types and getters for the operators. Fixes #345

### DIFF
--- a/include/ddc/kernels/splines/spline_evaluator_2d.hpp
+++ b/include/ddc/kernels/splines/spline_evaluator_2d.hpp
@@ -48,6 +48,11 @@ public:
     using bsplines_type1 = BSplinesType1;
     using bsplines_type2 = BSplinesType2;
 
+    using left_extrapolation_rule_1 = LeftExtrapolationRule1;
+    using right_extrapolation_rule_1 = RightExtrapolationRule1;
+    using left_extrapolation_rule_2 = LeftExtrapolationRule2;
+    using right_extrapolation_rule_2 = RightExtrapolationRule2;
+
     using interpolation_domain_type1 = ddc::DiscreteDomain<interpolation_mesh_type1>;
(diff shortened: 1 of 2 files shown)


include/ddc/kernels/splines/spline_evaluator_2d.hpp
```diff
@@ -80,13 +85,13 @@ class SplineEvaluator2D
 private:
     spline_domain_type m_spline_domain;
 
-    LeftExtrapolationRule1 m_left1_bc;
+    LeftExtrapolationRule1 m_left_extrap_rule_1;
 
     RightExtrapolationRule1 m_right1_bc;
```
tpadioleau, line 90: Do you mind also renaming `m_right1_bc` ?

tpadioleau (changes requested):
Almost fine to me, thanks

include/ddc/kernels/splines/spline_evaluator_2d.hpp
EmilyBourne: Didn't realise I'd missed this one. Fixed 77ade06

PR #349, opened 2024-03-21
Decouple SplineEvaluator from domain

### LINKED ISSUE #344: Spline evaluation cannot be performed on a subdomain smaller than batch_domain
```
operator()(...)
{
        interpolation_domain_type const interpolation_domain(spline_eval.domain());
        ddc::parallel_for_each(
                exec_space(),
                batch_domain(),
```

With

```
KOKKOS_FUNCTION batch_domain_type batch_domain() const noexcept
    {
        return ddc::remove_dims_of(spline_domain(), bsplines_domain());
    }
```

Implies the array receiving the evaluated splines has to be defined on a domain whose "batched part" has same size as the one of `spline_domain`, and cannot be a subset of it.

EmilyBourne:
Decouple `SplineEvaluator` from spline domain to prevent potential segmentation faults when using a smaller batch domain or unexpectedly non-updated values when using a larger batch domain. Fixes #344

### DIFF
--- a/include/ddc/kernels/splines/spline_evaluator.hpp
+++ b/include/ddc/kernels/splines/spline_evaluator.hpp
@@ -125,6 +125,16 @@ public:
 
     SplineEvaluator& operator=(SplineEvaluator&& x) = default;
 
+    left_extrapolation_rule_type left_extrapolation_rule() const
+    {
+        return m_left_extrap_rule;
+    }
+
+    right_extrapolation_rule_type right_extrapolation_rule() const
+    {
+        return m_right_extrap_rule;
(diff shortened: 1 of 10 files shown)


tpadioleau (approved):

include/ddc/kernels/splines/spline_evaluator.hpp
```diff
@@ -156,6 +135,8 @@ class SplineEvaluator
         return m_right_extrap_rule;
     }
 
+
+
```
tpadioleau, line 139:
```suggestion
```

PR #357, opened 2024-03-22
Merge reuse into tests workflow

tpadioleau:


.github/workflows/tests.yml
```diff
@@ -18,6 +18,13 @@ jobs:
         extensions: 'hpp,cpp'
         clangFormatVersion: 12
 
+  reuse:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+    - name: REUSE Compliance Check
```
jbigot, line 25: I'd make it a step of the lint job

```suggestion
    - uses: actions/checkout@v4
      name: REUSE Compliance Check
```

jbigot (approved):

.github/workflows/tests.yml
jbigot:
```suggestion
    - name: REUSE Compliance Check
      if: always()
```
That way, it will

tpadioleau: Ok why not, why is it better ?

jbigot: We don't spin up a whole new VM for a 1s job. Should slightly decrease pressure on runners

PR #363, opened 2024-03-26
Add a check to ensure documentation is present

EmilyBourne:
The flag `DOXYGEN_EXTRACT_ALL` is changed to ensure warnings are raised:
image

and relevant warnings are activated.

A small CI script is added which filters the warnings. Any warnings coming from files which have been changed in the PR  cause the CI to fail. The test is quite strict and ensures that any file which is modified is fully documented.

Currently the changes check is filtered to only include files in the `include/ddc/kernels/` folder, which means that all files related to splines and FFT should be documented.


jbigot:
Nice! Maybe you could add this contrib to your list in `AUTHORS`

EmilyBourne:
> Nice! Maybe you could add this contrib to your list in `AUTHORS`

Done 8e0f52944c51615018fce18f805c5cabe16fc5f1

tpadioleau (approved):
Thanks for the work

PR #364, opened 2024-03-26
Fix out of bounds periodic evaluation

EmilyBourne:
When using `SplineEvaluator2D` to evaluate a point (r,theta) outside the domain, the periodic conditions should be applied before the non-periodic conditions. This is important because in a periodic direction no points are truly outside the domain. They should therefore be relocated before calling something like `ConstantExtrapolationRule`.

### DIFF
--- a/include/ddc/kernels/splines/spline_evaluator_2d.hpp
+++ b/include/ddc/kernels/splines/spline_evaluator_2d.hpp
@@ -651,47 +651,47 @@ private:
      */
     template <class Layout, class... CoordsDims>
     KOKKOS_INLINE_FUNCTION double eval(
-            ddc::Coordinate<CoordsDims...> const& coord_eval,
+            ddc::Coordinate<CoordsDims...> coord_eval,
             ddc::ChunkSpan<double const, bsplines_domain_type, Layout, memory_space> const
                     spline_coef) const
     {
-        ddc::Coordinate<typename interpolation_mesh_type1::continuous_dimension_type>
-                coord_eval_interpolation1(coord_eval);
-        ddc::Coordinate<typename interpolation_mesh_type2::continuous_dimension_type>
-                coord_eval_interpolation2(coord_eval);
+        using Dim1 = typename interpolation_mesh_type1::continuous_dimension_type;
+        using Dim2 = typename interpolation_mesh_type2::continuous_dimension_type;
         if constexpr (bsplines_type1::is_periodic()) {
-            if (coord_eval_interpolation1 < ddc::discrete_space<bsplines_type1>().rmin()
-                || coord_eval_interpolation1 > ddc::discrete_space<bsplines_type1>().rmax()) {
-                coord_eval_interpolation1
+            if (ddc::get<Dim1>(coord_eval) < ddc::discrete_space<bsplines_type1>().rmin()
+                || ddc::get<Dim1>(coord_eval) > ddc::discrete_space<bsplines_type1>().rmax()) {
+                ddc::get<Dim1>(coord_eval)
                         -= Kokkos::floor(
-                                   (coord_eval_interpolation1
+                                   (ddc::get<Dim1>(coord_eval)
                                     - ddc::discrete_space<bsplines_type1>().rmin())
                                    / ddc::discrete_space<bsplines_type1>().length())
                            * ddc::discrete_space<bsplines_type1>().length();
             }
-        } else {
-            if (coord_eval_interpolation1 < ddc::discrete_space<bsplines_type1>().rmin()) {
-                return m_left_extrap_rule_1(coord_eval, spline_coef);
-            }
-            if (coord_eval_interpolation1 > ddc::discrete_space<bsplines_type1>().rmax()) {
-                return m_right_extrap_rule_1(coord_eval, spline_coef);
-            }
         }
         if constexpr (bsplines_type2::is_periodic()) {
-            if (coord_eval_interpolation2 < ddc::discrete_space<bsplines_type2>().rmin()
-                || coord_eval_interpolation2 > ddc::discrete_space<bsplines_type2>().rmax()) {
-                coord_eval_interpolation2
+            if (ddc::get<Dim2>(coord_eval) < ddc::discrete_space<bsplines_type2>().rmin()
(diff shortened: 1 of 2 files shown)


include/ddc/kernels/splines/spline_evaluator_2d.hpp
tpadioleau, line 1: To be honest I do not see the difference, to me they should behave the same.

EmilyBourne: In the old implementation we proceed in order of dimensions. Now periodic boundary conditions are considered first. The coordinate is also copied and altered. For example consider a domain [0,1] x [0, 2*pi] periodic in the second dimension. If we evaluate at (1.2, -pi) then with the old implementation we call `ConstantExtrapolationRule` on (1.2, -pi). It then crashes because `-pi` is out of bounds. With the new implementation periodic boundary conditions are handled first so (1.2, -pi) becomes (1.2,pi). This coordinate can be successfully evaluated with `ConstantExtrapolationRule`

EmilyBourne: To be honest I suspect `ConstantExtrapolationRule` needs to know something about the rule in the second dimension but for periodic BCs this correction is sufficient for all implementations as (-pi) is not really out of bounds

tpadioleau: Oh yes I see, could be an idea of unit test

EmilyBourne: There is a unit test already. It was filtering out this case. I have made the modifications to put it back

blegouix: https://github.com/CExA-project/ddc/blob/069493399273becfbde4205761a6d5afdba3a7fc/include/ddc/kernels/splines/constant_extrapolation_rule.hpp#L124C5-L125C71

This was not enough ?

EmilyBourne: [5d59f35](https://github.com/CExA-project/ddc/pull/364/commits/5d59f35cabbe96faa506214acde0643877f8c6b3)
[7e91a7f](https://github.com/CExA-project/ddc/pull/364/commits/7e91a7f3da83975ea468a718457e6caa2137a17f)

tpadioleau: I agree we have a test but i think we could make it more unitary.

EmilyBourne: So duplicate this function to have one case where only one coordinate is out of bounds and another where both are out of bounds?

tpadioleau: It is fine for now, @blegouix are the modifications ok for you ?

tests/splines/extrapolation_rule.cpp
```diff
@@ -336,10 +336,8 @@ static void ExtrapolationRuleSplineTest()
                 // Set coords_eval outside of the domain
                 ddc::get<I1>(coords_eval(e))
                         = xN<I1>() + (ddc::select<I1>(ddc::coordinate(e)) - x0<I1>()) + 1;
-#if defined(BC_GREVILLE)
                 ddc::get<I2>(coords_eval(e))
-                        = xN<I2>() + (ddc::select<I2>(ddc::coordinate(e)) - x0<I2>()) + 1;
-#endif
+                        = xN<I2>() + (ddc::select<I2>(ddc::coordinate(e)) - x0<I2>());
```
blegouix, line 340:
```suggestion
                        = xN<I2>() + (ddc::select<I2>(ddc::coordinate(e)) - x0<I2>()) + 1;
```

The `+1` is to ensure we are strictly outside of the domain even for `e` on "left border" (`ddc::select<I2>(ddc::coordinate(e)) = x0<I2>()`)

blegouix (commented):
Very good finding

include/ddc/kernels/splines/spline_evaluator_2d.hpp
```diff
@@ -651,76 +651,70 @@ class SplineEvaluator2D
      */
     template <class Layout, class... CoordsDims>
     KOKKOS_INLINE_FUNCTION double eval(
-            ddc::Coordinate<CoordsDims...> const& coord_eval,
+            ddc::Coordinate<CoordsDims...> coord_eval,
```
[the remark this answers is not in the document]
EmilyBourne: In the updated version `coord_eval` is modified so I want a copy not a constant reference

tests/splines/extrapolation_rule.cpp
EmilyBourne: Ah. I removed it so that the solution in the periodic case would be found in `vals`. I see 3 possible solutions:
1. exclude the left border
2. exclude the right border and add `ddc::coordinate(e+1)-ddc::coordinate(e)`
3. Add 2*`xN<I2>()`
4. Use `evaluator` to calculate the solution on the fly instead of using `vals`

Which do you think would be best? I think 3. would be the simplest

blegouix: I agree 3. looks good

EmilyBourne: Fixed [afb31fa](https://github.com/CExA-project/ddc/pull/364/commits/afb31fa412e18d9bc569b1c760374817c911745b)

include/ddc/kernels/splines/spline_evaluator_2d.hpp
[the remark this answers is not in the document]
EmilyBourne: E.g. here:
https://github.com/CExA-project/ddc/blob/bfe0494819e49397361d2aeb74ff4b8b24bbb26a/include/ddc/kernels/splines/spline_evaluator_2d.hpp#L663-L668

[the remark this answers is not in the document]
EmilyBourne: I want to modify this not a 1D coordinate as I need to pass the ND coordinate to the extrapolation rule

blegouix:
Good to me

tpadioleau (approved):
Thanks for the bug fix!

PR #372, opened 2024-03-29
Fix doc and warn as error

tpadioleau:


include/ddc/discrete_space.hpp
```diff
@@ -185,16 +185,15 @@ Arg init_discrete_space(std::tuple<DDimImpl, Arg>&& a)
 
 /** Move construct a global singleton discrete space and pass through remaining arguments
  *
- * @param a - the discrete space to move at index 0
- *          - the (2+) arguments to pass through in other indices
+ * @param a - The discrete space to move at index 0
+ *          - The (2+) arguments to pass through in other indices
```
tpadioleau, line 189:
```suggestion
 * @param a - the discrete space to move at index 0
 *          - the (2+) arguments to pass through in other indices
```

tpadioleau:
I am not sure the clang assisted doxygen works in the CI, seems to work on my computer though.

PR #377, opened 2024-04-02
Workaround ginkgo omp problem (use Gmres)

blegouix:
Following https://github.com/ginkgo-project/ginkgo/issues/1563#issuecomment-2024928079

Using Gmres avoid the problem. Stopping criteria needed to be relaxed to make all tests pass (2D splines were slightly failing which means 1e-19 leads to some kind of non-convergency. It is also in accord with recommandation here: https://github.com/ginkgo-project/ginkgo/issues/1563#issuecomment-1986031356)

OMP preconditionner parameter (Jacobi-block-size) is set to 1 again because of Ginkgo recommandation, and because we do not observe huge performance improvement on splines with preconditionner (~18% at most in old benchmarks).

BiCGStab should be recovered once the problem has been fixed from Ginkgo side.

It solves all problems we had due to this issue (in Gysela or DDC)

### DIFF
--- a/include/ddc/kernels/splines/matrix_sparse.hpp
+++ b/include/ddc/kernels/splines/matrix_sparse.hpp
@@ -77,7 +77,7 @@ unsigned int default_preconditionner_max_block_size() noexcept
 #endif
 #ifdef KOKKOS_ENABLE_OPENMP
     if (std::is_same_v<ExecSpace, Kokkos::OpenMP>) {
-        return 32u;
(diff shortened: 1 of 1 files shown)


tpadioleau:
That is cool, do you know if it is also enough to solve issues we encountered on the Gysela side ?

blegouix:
Yes I checked it, it is

tpadioleau (approved):

PR #379, opened 2024-04-02
[FFT] init_fourier_space returns PeriodicSampling

### LINKED ISSUE #333: We should only keep a version of `init_fourier_space` returning a `FourierSampling` in our public API.
@blegouix can maybe answer better than me on this but if I remember correctly the idea was to ease initialization of multiple dimensions at once. I think the only place where it is used in a variadic context is the DDC tests. It could be moved to the tests so that we only keep a version returning a `FourierSampling` in our public API.

_Originally posted by @tpadioleau in https://github.com/CExA-project/ddc/pull/121#discussion_r1517362751_

blegouix:
Closes #333


tpadioleau (approved):
Looks good to me, thx

Issue #190, opened 2023-10-05
Improve remove_dims_of


blegouix:
support:
`remove_dims_of<DimsToRemove...>(dom)`
in addition to:
`remove_dims_of(dom, dom_to_remove)`

(and make the change in gysela advection)

tpadioleau:
Yes I think we get closer to something user-friendly.

I am also considering a member function like the current `DiscreteDomain::remove` that returns a new `DiscreteDomain` with the same dimensions. What do you think of `DiscreteDomain DiscreteDomain::remove_dims_from(DiscreteDomain<DimsToRemove...> const& x)` that would return a new `DiscreteDomain` without the dimensions found in `x` ?

An other possibility would be to introduce a template class `dims<Dims...>` that would be an empty object representing the dimensions (not intervals but complete dimensions). You would use it like `remove_dims_of(dom, dims(DimsToRemove()...))`.

PR #381, opened 2024-04-02
remove_dims_from

blegouix:
@tpadioleau I tried your suggestion in #190 but it does not compile and I am not sure why.


tpadioleau:
Actually i am not sure it is necessary anymore. Some constructors of DiscreteVector/DiscreteElement/DiscreteDomain have a slicing behavior.

tpadioleau:
I think a `const` qualifier on the member function. Thus it cannot be used from a const qualified object.

blegouix:
I think this is just a matter of API then, I have no preference, just tell me which one you prefer

PR #384, opened 2024-04-03
Update matrix_sparse.hpp

tpadioleau:
Update magic number


blegouix:
Can you explain why you look at this ? It helps for the issue with OMP in Gysela ?

tpadioleau:
Yes 1e-16 was a bit too restrictive

tpadioleau:
As a rule of thumb I think the stopping criterion, $||r||$ should be $10^{-16} \approx ||A|| ||r||$, so for a good condition number < 100, we should stop somewhere at $||r|| \in [10^{-15} , 10^{-14}]$

PR #403, opened 2024-04-12
Use doxygen docker image

tpadioleau:
Here is my proposal


docs/CMakeLists.txt
```diff
@@ -11,14 +11,12 @@ if("${DOXYGEN_VERSION}" VERSION_GREATER_EQUAL 1.9.4 AND CMAKE_CXX_COMPILER_ID ST
...
 set(DOXYGEN_CLANG_DATABASE_PATH "${CMAKE_BINARY_DIR}")
-else()
-set(DOXYGEN_WARN_AS_ERROR YES)
 endif()
 set(DOXYGEN_DISABLE_INDEX YES)
 set(DOXYGEN_ENUM_VALUES_PER_LINE 1)
 set(DOXYGEN_EXAMPLE_PATH "${DDC_SOURCE_DIR}/examples")
 set(DOXYGEN_EXPAND_ONLY_PREDEF YES)
-set(DOXYGEN_EXTRACT_ALL NO)
+set(DOXYGEN_EXTRACT_ALL YES)
```
EmilyBourne, line 19: This cannot be changed. See the doxygen documentation cited in the description of #363

docs/CMakeLists.txt
```diff
@@ -50,20 +48,17 @@ set(DOXYGEN_TOC_INCLUDE_HEADINGS 4)
...
-    "${DDC_SOURCE_DIR}/include/ddc/*.hpp")
-message(STATUS ${HPP_SOURCES})
-
 doxygen_add_docs(doc
         "${CMAKE_CURRENT_SOURCE_DIR}/About.md"
         "${CMAKE_CURRENT_SOURCE_DIR}/first_steps.md"
         "${CMAKE_CURRENT_SOURCE_DIR}/heat_equation.md"
-        ${HPP_SOURCES}
+        "${DDC_SOURCE_DIR}/include/ddc/"
         ALL)
```
EmilyBourne, line 64: Was this file changed back intentionally?

docs/CMakeLists.txt
```diff
@@ -50,20 +48,17 @@ set(DOXYGEN_TOC_INCLUDE_HEADINGS 4)
 set(DOXYGEN_TYPEDEF_HIDES_STRUCT YES)
 set(DOXYGEN_UML_LIMIT_NUM_FIELDS 20)
 set(DOXYGEN_UML_LOOK YES)
-set(DOXYGEN_WARN_NO_PARAMDOC YES)
-set(DOXYGEN_WARN_IF_UNDOCUMENTED YES)
+set(DOXYGEN_WARN_IF_UNDOCUMENTED NO)
 set(DOXYGEN_WARN_IF_DOC_ERROR YES)
-set(DOXYGEN_WARN_AS_ERROR NO)
+set(DOXYGEN_WARN_IF_INCOMPLETE_DOC YES)
```
EmilyBourne, line 53: What is the difference between `DOXYGEN_WARN_IF_UNDOCUMENTED` and `DOXYGEN_WARN_IF_INCOMPLETE_DOC` ? I googled but `DOXYGEN_WARN_IF_INCOMPLETE_DOC` only gives 2 results.

docs/CMakeLists.txt
```diff
@@ -50,20 +48,17 @@ set(DOXYGEN_TOC_INCLUDE_HEADINGS 4)
...
+set(DOXYGEN_WARN_NO_PARAMDOC YES)
+set(DOXYGEN_WARN_IF_UNDOC_ENUM_VAL YES)
+set(DOXYGEN_WARN_AS_ERROR FAIL_ON_WARNINGS)
 set(DOXYGEN_WARN_LOGFILE "doxygen.log")
 set(DOXYGEN_IMAGE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/images")
-
-FILE(GLOB_RECURSE HPP_SOURCES
-    "${DDC_SOURCE_DIR}/include/ddc/*.hpp")
-message(STATUS ${HPP_SOURCES})
```
EmilyBourne, line 62: This was added to allow me to do `set(DOXYGEN_EXTRACT_ALL NO)`. Without this doxygen does not raise warnings

docs/CMakeLists.txt
```diff
@@ -50,20 +48,17 @@ set(DOXYGEN_TOC_INCLUDE_HEADINGS 4)
...
 set(DOXYGEN_UML_LOOK YES)
-set(DOXYGEN_WARN_NO_PARAMDOC YES)
-set(DOXYGEN_WARN_IF_UNDOCUMENTED YES)
+set(DOXYGEN_WARN_IF_UNDOCUMENTED NO)
 set(DOXYGEN_WARN_IF_DOC_ERROR YES)
-set(DOXYGEN_WARN_AS_ERROR NO)
+set(DOXYGEN_WARN_IF_INCOMPLETE_DOC YES)
+set(DOXYGEN_WARN_NO_PARAMDOC YES)
+set(DOXYGEN_WARN_IF_UNDOC_ENUM_VAL YES)
+set(DOXYGEN_WARN_AS_ERROR FAIL_ON_WARNINGS)
```
EmilyBourne, line 56: Are you sure you want this to fail if there are warnings? Won't this prevent you updating the documentation until all the warnings are fixed?

.github/workflows/pages.yml
```diff
@@ -56,10 +56,70 @@ jobs:
...
+          cmake \
+              -DDDC_BUILD_DOCUMENTATION=ON \
+              -DCMAKE_CXX_STANDARD=20 \
+              -DDOXYGEN_EXTRACT_ALL=NO \
+              -DDOXYGEN_WARN_IF_UNDOCUMENTED=YES \
+              -DDOXYGEN_WARN_IF_DOC_ERROR=YES \
+              -DDOXYGEN_WARN_IF_INCOMPLETE_DOC=YES \
+              -DDOXYGEN_WARN_NO_PARAMDOC=YES \
+              -DDOXYGEN_WARN_IF_UNDOC_ENUM_VAL=YES \
+              -DDOXYGEN_WARN_AS_ERROR=NO \
```
EmilyBourne, line 94: Now I understand why you changed the parameters as you change them back here. But I am still not sure how well this works. Did you test to ensure that the errors are correctly raised for the .hpp files? 
There are some parameters mentioned in the CMake with the same configuration as here. Did you simply include all `DOXYGEN_WARN_XXX` flags for completeness?

tpadioleau: Yes I copy-pasted the content, in case I decide to change the default later on.

I have just commit a missing parameter documentation in a function. Let us see if it works

docs/CMakeLists.txt
tpadioleau: But we don't have such warnings for now. Maybe something is wrong in the CI ?

docs/CMakeLists.txt
tpadioleau: Here are the doxygen definitions, the prefix `DOXYGEN` in the option name comes from cmake,
```
WARN_IF_UNDOCUMENTED

    If the WARN_IF_UNDOCUMENTED tag is set to YES then doxygen will generate warnings for undocumented members. If [EXTRACT_ALL](https://www.doxygen.nl/manual/config.html#cfg_extract_all) is set to YES then this flag will automatically be disabled.

    The default value is: YES.

WARN_IF_INCOMPLETE_DOC

    If WARN_IF_INCOMPLETE_DOC is set to YES, doxygen will warn about incomplete function parameter documentation. If set to NO, doxygen will accept that some parameters have no documentation without warning.

    The default value is: YES.```

File: docs/CMakeLists.txt (hunk 2 above)
Comment (tpadioleau) [reply docs/CMakeLists.txt:64]: Yes indeed because I thought the recursive option of Doxygen would be enough.

File: docs/CMakeLists.txt (hunk 5 above)
Comment (EmilyBourne) [reply docs/CMakeLists.txt:56]: There are no warnings raised while generating the docs because `DOXYGEN_EXTRACT_ALL` is set to YES. So actually I suspect that these parameters are meaningless in this file

File: docs/CMakeLists.txt (hunk 2 above)
Comment (EmilyBourne) [reply docs/CMakeLists.txt:64]: If it is sufficient then I agree that this is best, but I'm not sure that it is enough when you have `set(DOXYGEN_EXTRACT_ALL NO)`. It needs to be tested

Comment (EmilyBourne) [reply docs/CMakeLists.txt:64]: I have tested locally and there doesn't seem to be a big difference between these two methods. If anything your method retrieves more information

Comment (EmilyBourne) [reply docs/CMakeLists.txt:64]: But the errors definitely aren't triggered in this version

Comment (EmilyBourne) [reply docs/CMakeLists.txt:64]: If I make this change in my branch I still have the errors. Could the problem be due to the use of `set(DOXYGEN_EXTRACT_ALL NO)` which doesn't check the cache? Could this be overwriting your command-line options?

File: .github/workflows/pages.yml (hunk 6 above)
Comment (EmilyBourne) [reply .github/workflows/pages.yml:94]: Errors are now raised: https://github.com/CExA-project/ddc/actions/runs/8663814850/job/23758836602?pr=403

**Review by EmilyBourne — APPROVED** [review]

File: docs/CMakeLists.txt [hunk 7]
```diff
@@ -11,14 +11,12 @@ if("${DOXYGEN_VERSION}" VERSION_GREATER_EQUAL 1.9.4 AND CMAKE_CXX_COMPILER_ID ST
...
 set(DOXYGEN_CLANG_DATABASE_PATH "${CMAKE_BINARY_DIR}")
-else()
-set(DOXYGEN_WARN_AS_ERROR YES)
 endif()
 set(DOXYGEN_DISABLE_INDEX YES)
 set(DOXYGEN_ENUM_VALUES_PER_LINE 1)
 set(DOXYGEN_EXAMPLE_PATH "${DDC_SOURCE_DIR}/examples")
 set(DOXYGEN_EXPAND_ONLY_PREDEF YES)
-set(DOXYGEN_EXTRACT_ALL NO)
+set(DOXYGEN_EXTRACT_ALL YES CACHE STRING "")
```
EmilyBourne, line 19: Why `CACHE STRING` instead of `CACHE BOOL`?

tpadioleau: Actually i don't know how it translates to the Doxyfile if we make it a BOOL in cmake. In the documentation they only mention `set(DOXYGEN_VAR YES)`, i don't think these local variables are typed are they ?

EmilyBourne: As far as I could tell CMake interprets YES as a BOOL set to ON:

In the [docs for if](https://cmake.org/cmake/help/latest/command/if.html#constant) they say:
> True if the constant is 1, ON, YES, TRUE, Y, or a non-zero number (including floating point numbers).

Back when it was designed it seems that YES was a valid boolean:
https://cmake.org/pipermail/cmake/2007-December/018548.html

tpadioleau: I agree with you for that part but what about Doxyfile ? I was thinking of the case where cmake would dump "1" in the Doxyfile which is also a valid boolean for cmake. I don't know if this would be a valid value for a Doxyfile.

EmilyBourne: Ah ok. Makes sense :+1:

include/ddc/parallel_for_each.hpp
```diff
@@ -150,7 +150,6 @@ void parallel_for_each(
 }
 
 /** iterates over a nD domain using the `Kokkos` default execution space
- * @param[in] domain the domain over which to iterate
```
EmilyBourne, line 153: It is expected that this doesn't raise an error as I only check files in include/ddc/kernels

EmilyBourne (approved):

include/ddc/parallel_for_each.hpp
tpadioleau: in the CI yes but `doxygen.log` should have a line about it. I have documented the `ddc` namespace so that it should work when `EXTRACT_ALL` is `NO`.

PR #405, opened 2024-04-15
[Splines] Fixes 402

### LINKED ISSUE #402: Boundary conditions in spline_builder_2d.hpp
It seems there is mix between the boundary conditions on the first dimension and the second dimension. 
In spline_builder_2d.hpp, at lines 
https://github.com/CExA-project/ddc/blob/d652df884e1a53c9409905ebb2712530bb6f0c67/include/ddc/kernels/splines/spline_builder_2d.hpp#L382
and https://github.com/CExA-project/ddc/blob/d652df884e1a53c9409905ebb2712530bb6f0c67/include/ddc/kernels/splines/spline_builder_2d.hpp#L406
I think it should be `BcXmin2` and `BcXmax2` instead of `BcXmin1` and `BcXmax1` . 
In the case of Hermite boundary conditions on the two dimensions, it doesnt seem to affect it. But when the 2D basis has Hermite boundary conditions for one dimension and periodic boundary conditions for the second dimension, we have a error at the compilation.

blegouix:
Closes #402

### DIFF
--- a/include/ddc/kernels/splines/spline_builder_2d.hpp
+++ b/include/ddc/kernels/splines/spline_builder_2d.hpp
@@ -379,7 +379,7 @@ operator()(
             ddc::KokkosAllocator<double, MemorySpace>());
     auto spline1_deriv_min = spline1_deriv_min_alloc.span_view();
     auto spline1_deriv_min_opt = std::optional(spline1_deriv_min.span_cview());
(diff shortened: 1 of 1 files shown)


tpadioleau (approved):
Yep looks good to me thx

PR #406, opened 2024-04-15
Ginkgo throw error if not converged

### LINKED ISSUE #385: Generate error if iterative solver has not converged

blegouix:
Closes #385

### DIFF
--- a/include/ddc/kernels/splines/matrix_sparse.hpp
+++ b/include/ddc/kernels/splines/matrix_sparse.hpp
@@ -215,11 +215,13 @@ public:
                 m_solver
                         ->apply(to_gko_dense(gko_exec, b_subview),
                                 to_gko_dense(gko_exec, x_subview));
+                m_solver->remove_logger(convergence_logger);
             } else if (transpose == 'T') {
                 m_solver_tr->add_logger(convergence_logger);
                 m_solver_tr
                         ->apply(to_gko_dense(gko_exec, b_subview),
                                 to_gko_dense(gko_exec, x_subview));
+                m_solver_tr->remove_logger(convergence_logger);
             } else {
                 throw std::domain_error("transpose option not recognized");
             }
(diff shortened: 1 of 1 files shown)


include/ddc/kernels/splines/matrix_sparse.hpp
```diff
@@ -210,17 +211,22 @@ class Matrix_Sparse : public Matrix
             Kokkos::deep_copy(x_subview, b_subview);
 
             if (transpose == 'N') {
+                m_solver->add_logger(convergence_logger);
                 m_solver
                         ->apply(to_gko_dense(gko_exec, b_subview),
                                 to_gko_dense(gko_exec, x_subview));
             } else if (transpose == 'T') {
+                m_solver_tr->add_logger(convergence_logger);
```
tpadioleau, line 219: I find a bit weird to add a logger each iteration, don't you think ?

tpadioleau (changes requested):

include/ddc/kernels/splines/matrix_sparse.hpp
blegouix: I agree but there is no way to get the convergence_logger from the m_solver or m_solver_tr if I attach it in factorize_method. Alternative would be to put convergence_logger as member of the class, you prefer it ?

blegouix: Looking at the implementation of add_logger it is filling a std::vector so the current implementation is bad indeed, I can call remove_logger or pass convergence_logger as member of the class.

tpadioleau: Mmmh could be a solution, I am wondering if it can be used multiple times though and I don't know about the cost of creating. Reading https://ginkgo-project.github.io/ginkgo-generated-documentation/doc/develop/classgko_1_1log_1_1EnableLogging.html we can at least remove it to avoid some kind of "leak"/accumulation.

PR #411, opened 2024-04-17
Remove spline_dim_type

blegouix:
spline_dim_type is unused in ddc and gysela.

If we have no practical use case I think we should remove it


EmilyBourne:
I agree that there is no reason to use `spline_dim_type` instead of `spline_domain_type`

PR #412, opened 2024-04-17
[Splines] Privatize compute_offset

blegouix:
`compute_offset` has no reason to be public

### DIFF
--- a/include/ddc/kernels/splines/spline_builder.hpp
+++ b/include/ddc/kernels/splines/spline_builder.hpp
@@ -161,6 +161,7 @@ private:
     // interpolator specific
     std::unique_ptr<ddc::detail::Matrix> matrix;
 
+    /// Calculate offset so that the matrix is diagonally dominant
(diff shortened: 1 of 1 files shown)


include/ddc/kernels/splines/spline_builder.hpp
```diff
@@ -161,9 +161,9 @@ class SplineBuilder
     // interpolator specific
     std::unique_ptr<ddc::detail::Matrix> matrix;
 
-public:
     int compute_offset(interpolation_domain_type const& interpolation_domain);
 
+public:
```
EmilyBourne, line 166: I agree with this change but I still think it is worth documenting this function even if only briefly so we remember what it is for. There is an explanation inside the function definition:
```suggestion
    /// Calculate offset so that the matrix is diagonally dominant
    int compute_offset(interpolation_domain_type const& interpolation_domain);

public:
```

tpadioleau: Can you clarify an offset of what ?

PR #415, opened 2024-04-18
BSplines doc

blegouix:


blegouix:
There are several TODO which identify things I am not sure how to document and that you may know better than me what they are.

include/ddc/kernels/splines/bsplines_non_uniform.hpp
```diff
@@ -28,39 +28,49 @@ struct NonUniformBSplinesBase
 
 } // namespace detail
 
-/// NonUniformPointSampling specialization of BSplines
+/**
+ * The type of a non-uniform BSplines 1D basis.
+ *
+ * @tparam Tag The tag identifying the continuous dimension which supports the building of the BSplines.
+ * @tparam D The degree of the BSplines.
+ */
```
EmilyBourne, line 36: I am not sure what the correct formatting should be for "BSplines" but I don't think it should be "BSplines if it doesn't refer to a class. Wikipedia and "The NURBS book" use "B-splines" it is an abbreviation of "basis spline" so I think the "-" makes sense.
```suggestion
/**
 * The type of a non-uniform 1D basis of B-splines.
 *
 * @tparam Tag The tag identifying the continuous dimension which supports the building of the B-splines.
 * @tparam D The degree of the B-splines.
 */
```
If you agree, then this change probably needs applying over both these files

include/ddc/kernels/splines/bsplines_non_uniform.hpp
```diff
@@ -28,39 +28,49 @@ struct NonUniformBSplinesBase
...
 template <class Tag, std::size_t D>
 class NonUniformBSplines : detail::NonUniformBSplinesBase
 {
     static_assert(D > 0, "Parameter `D` must be positive");
 
 public:
+    /// @brief The tag identifying the continuous dimension which supports the building of the BSplines.
     using tag_type = Tag;
```
EmilyBourne, line 44: What is meant by "the continuous dimension which supports the building of the BSplines." ? 

Are you referring to the mathematical term "support" ? I don't know if this can be used as a verb. Even if it can, because "supports" is such a common word I think it is important to mention functions in the same sentence so it is clear that you mean it in this sense
```suggestion
    /// @brief The tag identifying the continuous dimension on which the support of the B-spline functions are defined.
    using tag_type = Tag;
```

include/ddc/kernels/splines/bsplines_non_uniform.hpp
```diff
@@ -28,39 +28,49 @@ struct NonUniformBSplinesBase
...
 
-
+    /// @brief The discrete dimension identifying BSplines.
     using discrete_dimension_type = NonUniformBSplines;
 
-public:
+    /// @brief The degree of BSplines.
```
EmilyBourne, line 49:
```suggestion
    /// @brief The degree of the B-splines.
```

include/ddc/kernels/splines/bsplines_non_uniform.hpp
```diff
@@ -28,39 +28,49 @@ struct NonUniformBSplinesBase
...
     }
 
+    /// @brief Indicates if the BSplines are radial or not (should be deprecated soon because this concept is not in the scope of DDC).
     static constexpr bool is_radial() noexcept
     {
         return false;
     }
 
+    /// @brief Indicates if the BSplines are uniform or not (this is not the case here).
     static constexpr bool is_uniform() noexcept
```
EmilyBourne, line 68: Somewhere in this file, you should explain what is meant by "uniform" this is often misunderstood and different words are sometimes used for this context. I suspect adding a sentence here and elaborating in the class docstring

include/ddc/kernels/splines/bsplines_non_uniform.hpp
```diff
@@ -28,39 +28,49 @@ struct NonUniformBSplinesBase
...
     {
         return false;
     }
 
+    /// @brief Impl TODO
     template <class DDim, class MemorySpace>
     class Impl
```
EmilyBourne, line 75: TODO @tpadioleau

include/ddc/kernels/splines/bsplines_non_uniform.hpp
```diff
@@ -75,23 +85,20 @@ class NonUniformBSplines : detail::NonUniformBSplinesBase
         int m_nknots;
 
     public:
+        /// @brief The type of discrete dimension representing the BSplines.
         using discrete_dimension_type = NonUniformBSplines;
```
EmilyBourne, line 89:
```suggestion
        /// @brief The type of the discrete dimension representing the B-splines.
        using discrete_dimension_type = NonUniformBSplines;
```
This change also needs applying to the next 3 docsctrings. "the" or "a" is needed for non-plural nouns. "du" = "of the", "d'un" = "of a"

include/ddc/kernels/splines/bsplines_non_uniform.hpp
```diff
@@ -75,23 +85,20 @@ class NonUniformBSplines : detail::NonUniformBSplinesBase
...
 
+        /// @brief The type of discrete domain identifying the BSplines.
         using discrete_domain_type = DiscreteDomain<DDim>;
```
EmilyBourne, line 92: Does a domain identify an object?

include/ddc/kernels/splines/bsplines_non_uniform.hpp
```diff
@@ -75,23 +85,20 @@ class NonUniformBSplines : detail::NonUniformBSplinesBase
...
 
+        /// @brief The type of discrete element identifying a BSpline.
         using discrete_element_type = DiscreteElement<DDim>;
 
+        /// @brief The type of discrete vector representing an "indexes displacement" between two BSplines.
```
EmilyBourne, line 97: The plural of "index" is "indices"

include/ddc/kernels/splines/bsplines_non_uniform.hpp
blegouix: `discrete domain whose elements identify the BSplines` ?

include/ddc/kernels/splines/bsplines_non_uniform.hpp
```diff
@@ -108,92 +115,184 @@ class NonUniformBSplines : detail::NonUniformBSplinesBase
...
+        /// @brief Move-assigns
         Impl& operator=(Impl&& x) = default;
 
+        /** @brief Evaluates BSplines at a given coordinate
+         *
+         * The values are computed for every BSplines at the given coordinate x. A spline approximation at coordinate x is a linear combination of those BSplines evaluations weigthed with splines coefficients of the spline-transformed initial discrete function.
+         * @param[out] values The values of the BSplines evaluated at coordinate x. It has to be a (1+degree) 1D mdspan.
+         * @param[in] x The coordinate where BSplines are evaluated.
+         * @return TODO
+         */
```
EmilyBourne, line 147:
```suggestion
        /** @brief Evaluates non-zero B-splines at a given coordinate.
         *
         * The values are computed for every B-spline with support at the given coordinate x. There are only (degree+1)
         * B-splines which are non-zero at any given point. It is these B-splines which are evaluated.
         * This can be useful to calculate a spline approximation of a function. A spline approximation at coordinate x
         * is a linear combination of these B-spline evaluations weighted with spline coefficients of the spline-transformed
         * initial discrete function.
         *
         * @param[out] values The values of the B-splines evaluated at coordinate x. It has to be a 1D mdspan with (1+degree) elements.
         * @param[in] x The coordinate where B-splines are evaluated.
         * @return The index of the first B-spline which is evaluated.
         */
```

include/ddc/kernels/splines/bsplines_non_uniform.hpp
blegouix: `whose knots are uniformly distributed` is correct ?

include/ddc/kernels/splines/bsplines_non_uniform.hpp
```diff
@@ -108,92 +115,184 @@ class NonUniformBSplines : detail::NonUniformBSplinesBase
...
         KOKKOS_INLINE_FUNCTION discrete_element_type
         eval_basis(DSpan1D values, ddc::Coordinate<Tag> const& x) const;
 
+        /** @brief Evaluates BSplines derivatives at a given coordinate
```
EmilyBourne: Recurring grammar error. When an object is comprised of 2 words and is plural, only the second word takes an "s", so "B-spline derivatives" not "B-splines derivatives"

include/ddc/kernels/splines/bsplines_non_uniform.hpp
```diff
@@ -108,92 +115,184 @@ class NonUniformBSplines : detail::NonUniformBSplinesBase
...
         eval_deriv(DSpan1D derivs, ddc::Coordinate<Tag> const& x) const;
 
+        /** @brief Evaluates B-splines values and n derivatives at a given coordinate
+         *
+         * The values and derivatives are computed for every B-splines at the given coordinate x.
+         * @param[out] derivs The values and n derivatives of the B-splines evaluated at coordinate x. It has to be a (1+degree)*(1+n) 2D mdspan.
+         * @param[in] x The coordinates where B-splines derivatives are evaluated.
+         * @param[in] n The number of derivatives to evaluate (in addition to the B-splines values themselves).
+         * @return TODO
+         */
```
EmilyBourne, line 168: See above

include/ddc/kernels/splines/bsplines_non_uniform.hpp
```diff
@@ -108,92 +115,184 @@ class NonUniformBSplines : detail::NonUniformBSplinesBase
...
         KOKKOS_INLINE_FUNCTION discrete_element_type eval_basis_and_n_derivs(
                 ddc::DSpan2D derivs,
                 ddc::Coordinate<Tag> const& x,
                 std::size_t n) const;
 
+        /** @brief Compute the integrals of the B-splines.
+         *
+         * @param[out] int_vals The values of the integrals. It has to be a (1+nbasis) 1D mdspan.
```
EmilyBourne, line 176: Are you sure about this size? I thought you could either calculate `nbasis` functions or `size` (where the additional degree values are copied using the periodicity). I don't understand what would fill the extra 1 value here.

include/ddc/kernels/splines/bsplines_non_uniform.hpp
```diff
@@ -108,92 +115,184 @@ class NonUniformBSplines : detail::NonUniformBSplinesBase
...
+         * @return The values of the integrals.
+         */
```
EmilyBourne, line 178:
```suggestion
        /** @brief Compute the integrals of the B-splines.
         *
         * The integral of each of the B-splines over their support within the domain on which this basis was defined.
         *
         * @param[out] int_vals The values of the integrals. It has to be a (1+nbasis) 1D mdspan.
         * @return The values of the integrals.
         */
```
> over their support within the domain on which this basis was defined.

This is especially important to mention for the uniform case where B-splines have support outside the domain

include/ddc/kernels/splines/bsplines_non_uniform.hpp
```diff
@@ -108,92 +115,184 @@ class NonUniformBSplines : detail::NonUniformBSplinesBase
...
 
+        /** @brief Returns the coordinate of the knot corresponding to the given index for a B-spline.
+         *
+         * @param[in] knot_idx Integer identifying index of the knot.
+         * @return Coordinate of the knot.
+         */
         KOKKOS_INLINE_FUNCTION ddc::Coordinate<Tag> get_knot(int knot_idx) const noexcept
         {
             return ddc::coordinate(ddc::DiscreteElement<mesh_type>(knot_idx + degree()));
         }
```
EmilyBourne, line 192:
```suggestion
        /** @brief Returns the coordinate of the knot corresponding to the given index.
         *
         * @param[in] knot_idx Integer identifying index of the knot.
         * @return Coordinate of the knot.
         */
        KOKKOS_INLINE_FUNCTION ddc::Coordinate<Tag> get_knot(int knot_idx) const noexcept
        {
            return ddc::coordinate(ddc::DiscreteElement<mesh_type>(knot_idx + degree()));
        }
```

This knot is not related to any particular B-spline

include/ddc/kernels/splines/bsplines_non_uniform.hpp
blegouix: No it's a typo

include/ddc/kernels/splines/bsplines_non_uniform.hpp
```diff
@@ -108,92 +115,184 @@ class NonUniformBSplines : detail::NonUniformBSplinesBase
...
 
+        /** @brief Returns the coordinate of the first support knot associated to a discrete_element identifying a B-spline.
+         *
+         * @param[in] ix DiscreteElement identifying the B-spline.
+         * @return Coordinate of the knot.
+         */
```
EmilyBourne, line 198:
```suggestion
        /** @brief Returns the coordinate of the first knot in the support of the identified B-spline.
         *
         * Each B-spline has a support defined over (degree+2) knots. For a B-spline identified by the
         * provided DiscreteElement, this function returns the first knot in the support of the B-spline.
         * In other words it returns the lower bound of the support.
         *
         * @param[in] ix DiscreteElement identifying the B-spline.
         * @return Coordinate of the knot.
         */
```

include/ddc/kernels/splines/bsplines_non_uniform.hpp
```diff
@@ -108,92 +115,184 @@ class NonUniformBSplines : detail::NonUniformBSplinesBase
...
         {
             return ddc::coordinate(ddc::DiscreteElement<mesh_type>(ix.uid() + degree() + 1));
         }
 
+        /** @brief Returns the coordinate of the \f$n\f$-th knot associated to a DiscreteElement identifying a B-spline.
+         *
+         * @param[in] ix DiscreteElement identifying the B-spline.
+         * @param[in] n Integer indexing a knot in the support of the B-spline.
+         * @return Coordinate of the knot.
+         */
```
EmilyBourne, line 221:
```suggestion
        /** @brief Returns the coordinate of the first knot in the support of the identified B-spline.
         *
         * Each B-spline has a support defined over (degree+2) knots. For a B-spline identified by the
         * provided DiscreteElement, this function returns the (n+1)-th knot in the support of the B-spline.
         *
         * @param[in] ix DiscreteElement identifying the B-spline.
         * @param[in] n Index of the relevant  knot from the support of the B-spline.
         * @return Coordinate of the knot.
         */
```
Not sure about the description of `n`. Feel free to change if you have a better suggestion

include/ddc/kernels/splines/bsplines_non_uniform.hpp
```diff
@@ -108,92 +115,184 @@ class NonUniformBSplines : detail::NonUniformBSplinesBase
...
                 int n) const
         {
             return ddc::coordinate(ddc::DiscreteElement<mesh_type>(ix.uid() + n));
         }
 
+        /** @brief Returns the coordinate of the lower boundary knot.
+         *
+         * @return Coordinate of the lower boundary knot.
+         */
         KOKKOS_INLINE_FUNCTION ddc::Coordinate<Tag> rmin() const noexcept
```
EmilyBourne, line 233: This is a bad description. We are knot assured that the knots are all inside the domain. I use the knot to calculate this but I don't think they should appear in the description.
```suggestion
        /** @brief Returns the coordinate of the lower bound of the domain on which the B-splines are defined.
         *
         * @return Coordinate of the lower bound of the domain.
         */
        KOKKOS_INLINE_FUNCTION ddc::Coordinate<Tag> rmin() const noexcept
```

include/ddc/kernels/splines/bsplines_non_uniform.hpp
```diff
@@ -108,92 +115,184 @@ class NonUniformBSplines : detail::NonUniformBSplinesBase
...
         {
             return get_knot(0);
         }
 
+        /** @brief Returns the coordinate of the upper boundary knot.
+         *
+         * @return Coordinate of the upper boundary knot.
+         */
         KOKKOS_INLINE_FUNCTION ddc::Coordinate<Tag> rmax() const noexcept
```
EmilyBourne, line 242: See above

include/ddc/kernels/splines/bsplines_non_uniform.hpp
```diff
@@ -108,92 +115,184 @@ class NonUniformBSplines : detail::NonUniformBSplinesBase
...
         {
             return get_knot(ncells());
         }
 
+        /** @brief Returns the length of the domain supporting knots.
+         *
+         * @return The length of the domain supporting knots.
+         */
         KOKKOS_INLINE_FUNCTION double length() const noexcept
```
EmilyBourne, line 251: Again, not knots.
```suggestion
        /** @brief Returns the length of the domain.
         *
         * @return The length of the domain.
         */
        KOKKOS_INLINE_FUNCTION double length() const noexcept
```

include/ddc/kernels/splines/bsplines_non_uniform.hpp
EmilyBourne: Following comments below I have realised that there is information missing here:
```suggestion
        /** @brief Returns the coordinate of the knot corresponding to the given index for a B-spline.
         *
         * Returns the coordinate of the knot corresponding to the given index for a B-spline. The domain
         * over which the B-splines are defined is comprised of ncells+1 knots however there are a total of
         * ncells+1+2*degree knots. The additional knots which control the shape of the B-splines near the
         * boundary are added before and after the break points. The knot index is therefore in the interval [-degree, ncells+degree]
         *
         * @param[in] knot_idx Integer identifying index of the knot.
         * @return Coordinate of the knot.
         */
        KOKKOS_INLINE_FUNCTION ddc::Coordinate<Tag> get_knot(int knot_idx) const noexcept
        {
            return ddc::coordinate(ddc::DiscreteElement<mesh_type>(knot_idx + degree()));
        }
```

include/ddc/kernels/splines/bsplines_non_uniform.hpp
```diff
@@ -75,23 +85,20 @@ class NonUniformBSplines : detail::NonUniformBSplinesBase
...
-        template <class OriginMemorySpace>
-        explicit Impl(Impl<DDim, OriginMemorySpace> const& impl)
-            : m_domain(impl.m_domain)
-            , m_nknots(impl.m_nknots)
-        {
-        }
-
         /// @brief Construct a `Impl` using a brace-list, i.e. `Impl bsplines({0., 1.})`
         explicit Impl(std::initializer_list<ddc::Coordinate<Tag>> breaks)
             : Impl(breaks.begin(), breaks.end())
```
EmilyBourne, line 104: You need to specify what this brace-list represents

include/ddc/kernels/splines/bsplines_non_uniform.hpp
```diff
@@ -108,92 +115,184 @@ class NonUniformBSplines : detail::NonUniformBSplinesBase
...
         }
 
+        /** @brief TODO
+         *
+         * @return TODO.
+         */
         KOKKOS_INLINE_FUNCTION std::size_t size() const noexcept
         {
             return degree() + ncells();
         }
```
EmilyBourne, line 263:
```suggestion
        /** @brief Returns the number of elements necessary to construct a spline representation of a function.
         *
         * For a non-periodic domain the number of elements necessary to construct a spline representation of a function
         * is equal to the number of basis functions. However in the periodic case it additionally includes degree additional elements
         * which allow the first B-splines to be evaluated close to rmax (where they also appear due to the periodicity).
         *
         * @return The number of elements necessary to construct a spline representation of a function.
         */
        KOKKOS_INLINE_FUNCTION std::size_t size() const noexcept
        {
            return degree() + ncells();
        }
```

include/ddc/kernels/splines/bsplines_non_uniform.hpp
```diff
@@ -108,92 +115,184 @@ class NonUniformBSplines : detail::NonUniformBSplinesBase
...
 
-        /// Returns the discrete domain including ghost bsplines
+        /** @brief Returns the discrete domain including ghost bsplines
+         *
+         * @return The discrete domain including ghost bsplines.
+         */
         KOKKOS_INLINE_FUNCTION discrete_domain_type full_domain() const
```
EmilyBourne, line 269: Are ghost b-splines what I am describing in the docstring above? If so then one or both of these docstrings should define that term

include/ddc/kernels/splines/bsplines_non_uniform.hpp
```diff
@@ -108,92 +115,184 @@ class NonUniformBSplines : detail::NonUniformBSplinesBase
...
         {
             return discrete_domain_type(discrete_element_type(0), discrete_vector_type(size()));
         }
 
+        /** @brief TODO
+         *
+         * @return TODO
+         */
         KOKKOS_INLINE_FUNCTION std::size_t npoints() const noexcept
```
EmilyBourne, line 278:
```suggestion
        /** @brief The number of break points
         *
         * The number of break points or cell boundaries.
         *
         * @return The number of break points
         */
        KOKKOS_INLINE_FUNCTION std::size_t npoints() const noexcept
```

include/ddc/kernels/splines/bsplines_non_uniform.hpp
```diff
@@ -108,92 +115,184 @@ class NonUniformBSplines : detail::NonUniformBSplinesBase
...
         {
             return m_nknots - 2 * degree();
         }
 
+        /** @brief TODO
+         *
+         * @return TODO
+         */
         KOKKOS_INLINE_FUNCTION std::size_t nbasis() const noexcept
```
EmilyBourne, line 287:
```suggestion
        /** @brief Returns the number of basis functions.
         *
         * The number of functions in the spline basis.
         *
         * @return The number of basis functions.
         */
        KOKKOS_INLINE_FUNCTION std::size_t nbasis() const noexcept
```

include/ddc/kernels/splines/bsplines_non_uniform.hpp
```diff
@@ -209,6 +308,11 @@ struct is_non_uniform_bsplines : public std::is_base_of<detail::NonUniformBSplin
 {
 };
 
+/**
+ * @brief Indicates if a tag corresponds to uniform B-splines of not.
+ *
+ * @tparam The presumed non-uniform B-splines.
+ */
 template <class DDim>
 constexpr bool is_non_uniform_bsplines_v = is_non_uniform_bsplines<DDim>::value;
```
EmilyBourne, line 317:
```suggestion
 * @brief Indicates if a tag corresponds to non-uniform B-splines or not.
 *
 * @tparam The presumed non-uniform B-splines.
 */
template <class DDim>
constexpr bool is_non_uniform_bsplines_v = is_non_uniform_bsplines<DDim>::value;
```

include/ddc/kernels/splines/bsplines_non_uniform.hpp
EmilyBourne: I'm not sure. Maybe, this works if DiscreteDomain is properly documented :sweat_smile: but as the aim for docs is to make it easily readable for someone who is unfamiliar with DDC I think it would be better if there was also more information here.

By the way, do you mind not closing conversations that I start (auto-close because you applied the suggestion as-is is fine). I like to go back and compare my comment to the commits so I don't have to redo a complete review

EmilyBourne: :+1: perfect thanks

include/ddc/kernels/splines/bsplines_non_uniform.hpp
```diff
@@ -108,92 +115,217 @@ class NonUniformBSplines : detail::NonUniformBSplinesBase
...
         KOKKOS_INLINE_FUNCTION discrete_element_type eval_basis_and_n_derivs(
                 ddc::DSpan2D derivs,
                 ddc::Coordinate<Tag> const& x,
                 std::size_t n) const;
 
+        /** @brief Compute the integrals of the B-splines.
+         *
+         * The integral of each of the B-splines over their support within the domain on which this basis was defined.
+         *
+         * @param[out] int_vals The values of the integrals. It has to be a (nbasis) 1D mdspan.
```
EmilyBourne, line 193: I am not sure about this syntax `(nbasis) 1D` for me it is not clear that `nbasis` refers to the size

include/ddc/kernels/splines/bsplines_non_uniform.hpp
```diff
@@ -28,39 +28,49 @@ struct NonUniformBSplinesBase
...
     }
 
+    /// @brief Indicates if the B-splines are periodic or not.
     static constexpr bool is_periodic() noexcept
     {
         return Tag::PERIODIC;
     }
 
+    /// @brief Indicates if the B-splines are radial or not (should be deprecated soon because this concept is not in the scope of DDC).
     static constexpr bool is_radial() noexcept
```
tpadioleau, line 62: I agree, i don't think it used anywhere.
```suggestion
     [[deprecated]] static constexpr bool is_radial() noexcept
```

include/ddc/kernels/splines/bsplines_non_uniform.hpp
```diff
@@ -28,39 +28,49 @@ struct NonUniformBSplinesBase
...
         return false;
     }
 
+    /// @brief Indicates if the B-splines are uniform or not (this is not the case here).
     static constexpr bool is_uniform() noexcept
     {
         return false;
     }
 
+    /// @brief Impl TODO
```
tpadioleau, line 73:
```suggestion
    /// @brief Impl Storage class of the static attributes of the discrete dimension.
    /// @tparam DDim The name of the discrete dimension.
    /// @tparam MemorySpace The Kokkos memory space where the attributes are being stored.
```
What do you think ?

include/ddc/kernels/splines/bsplines_non_uniform.hpp
tpadioleau: I have answered above

include/ddc/kernels/splines/bsplines_non_uniform.hpp
blegouix: Thanks for that

include/ddc/kernels/splines/bsplines_non_uniform.hpp
```diff
@@ -108,93 +119,229 @@ class NonUniformBSplines : detail::NonUniformBSplinesBase
...
 
+        /** @brief Returns the coordinate of the first knot in the support of the identified B-spline.
+         *
+         * Each B-spline has a support defined over (degree+2) knots. For a B-spline identified by the
+         * provided DiscreteElement, this function returns the (n+1)-th knot in the support of the B-spline.
+         *
+         * @param[in] ix DiscreteElement identifying the B-spline.
+         * @param[in] n Integer indexing a knot in the support of the B-spline.
+         * @return Coordinate of the knot.
+         */
```
EmilyBourne, line 258: Sorry I seem to have given you the wrong brief description here.

```suggestion
        /** @brief Returns the coordinate of the (n+1)th knot in the support of the identified B-spline.
         *
         * Each B-spline has a support defined over (degree+2) knots. For a B-spline identified by the
         * provided DiscreteElement, this function returns the (n+1)-th knot in the support of the B-spline.
         *
         * @param[in] ix DiscreteElement identifying the B-spline.
         * @param[in] n Integer indexing a knot in the support of the B-spline.
         * @return Coordinate of the knot.
         */
```

include/ddc/kernels/splines/bsplines_non_uniform.hpp
blegouix: I think as we are inside DDC, we can assume the user is aware of what a DiscreteDomain is. So to me:

```
The type of a DiscreteDomain whose elements identify the B-splines.
```

looks ok.

include/ddc/kernels/splines/bsplines_non_uniform.hpp
blegouix: bae41cc7d22ed5464a816d64484655ad14d2bd10

include/ddc/kernels/splines/bsplines_non_uniform.hpp
```diff
@@ -75,125 +92,269 @@ class NonUniformBSplines : detail::NonUniformBSplinesBase
...
-            , m_nknots(impl.m_nknots)
-        {
-        }
-
-        /// @brief Construct a `Impl` using a brace-list, i.e. `Impl bsplines({0., 1.})`
+        /** @brief Constructs an Impl using a brace-list, i.e. `Impl bsplines({0., 1.})`
+         *
+         * The brace-list is usually the iterable list of knots. However, only the first and last
+         * indexes of the elements of the lists are effectively used by the constructor to build the Impl.
+         */
```
EmilyBourne, line 113: I think that `breaks` represents the break points not the knots.

> However, only the first and last indexes of the elements of the lists are effectively used by the constructor

This is not true. `breaks.begin()` and `breaks.end()` are not indexes they are iterators. `breaks.end()` is not the last element of the list, it is the one after (the stop condition for the iterator). This constructor calls another constructor below. It uses all the elements

include/ddc/kernels/splines/bsplines_non_uniform.hpp
```diff
@@ -75,125 +92,269 @@ class NonUniformBSplines : detail::NonUniformBSplinesBase
...
             : Impl(breaks.begin(), breaks.end())
         {
         }
 
-        /// @brief Construct a `Impl` using a C++20 "common range".
+        /** @brief Constructs an Impl using a std::vector.
+         *
+         * The std::vector is usually the iterable list of knots. However, only the first and last
+         * indexes of the elements of the lists are effectively used by the constructor to build the Impl.
+         */
```
EmilyBourne, line 123: Same comment and above

include/ddc/kernels/splines/bsplines_non_uniform.hpp
```diff
@@ -75,125 +92,269 @@ class NonUniformBSplines : detail::NonUniformBSplinesBase
...
         explicit Impl(std::vector<ddc::Coordinate<Tag>> const& breaks)
             : Impl(breaks.begin(), breaks.end())
         {
         }
 
-        /// @brief Construct a `Impl` using a pair of iterators.
+        /** @brief Construct a `Impl` using a pair of iterators.
+         *
+         * The (lower and upper) iterators are used to build the DiscreteDomain indexing the B-splines.
+         */
```
EmilyBourne, line 132: See e.g. https://cplusplus.com/reference/iterator/ for a description of iterators. I'm not sure it is relevant to talk about a "lower iterator" or "upper iterator", I'm also not clear what an "upper iterator" is. Given your comments above I think you think it is like a `ubound` in Fortran but this is not the case.

This constructor builds the knots from which the B-splines are constructed from the break points. This constructor should specify which method is used to do this. It does not build the DiscreteDomain indexing the B-splines. It builds the DiscreteDomain which describes the knots

EmilyBourne: @blegouix I disagree that this comment has been handled:

>This constructor builds the knots from which the B-splines are constructed from the break points. This constructor should specify which method is used to do this.

EmilyBourne: Here the first degree knots are set equal to the first break point and the last degree knots are set equal to the last break point. This is not the only solution possible (a different method is used in the uniform case)

EmilyBourne: Done

include/ddc/kernels/splines/bsplines_non_uniform.hpp
```diff
@@ -70,125 +95,301 @@ class NonUniformBSplines : detail::NonUniformBSplinesBase
...
-            : m_domain(impl.m_domain)
-            , m_nknots(impl.m_nknots)
-        {
-        }
-
-        /// @brief Construct a `Impl` using a brace-list, i.e. `Impl bsplines({0., 1.})`
+        /** @brief Constructs an Impl using a brace-list, i.e. `Impl bsplines({0., 1.})`
+         *
+         * The brace-list is the list of break points. It is used to build the DiscreteDomain indexing
+         * the knots and iterated over to build the knots coordinates list and initialize the associated DiscreteSpace.
```
EmilyBourne, line 115: Needs updating. Ideally if you know how I think the easiest is to reference the constructor which does the work

include/ddc/kernels/splines/bsplines_non_uniform.hpp
```diff
@@ -70,125 +95,301 @@ class NonUniformBSplines : detail::NonUniformBSplinesBase
...
         explicit Impl(std::initializer_list<ddc::Coordinate<Tag>> breaks)
             : Impl(breaks.begin(), breaks.end())
         {
         }
 
-        /// @brief Construct a `Impl` using a C++20 "common range".
+        /** @brief Constructs an Impl using a std::vector.
+         *
+         * The std::vector is the list of break points. It is used to build the DiscreteDomain indexing
+         * the knots and iterated over to build the knots coordinates list and initialize the associated DiscreteSpace.
```
EmilyBourne, line 127: Needs updating. Ideally if you know how I think the easiest is to reference the constructor which does the work

include/ddc/kernels/splines/bsplines_non_uniform.hpp
```diff
@@ -70,125 +95,301 @@ class NonUniformBSplines : detail::NonUniformBSplinesBase
...
+         */
         KOKKOS_INLINE_FUNCTION discrete_element_type
         eval_basis(DSpan1D values, ddc::Coordinate<Tag> const& x) const;
 
+        /** @brief Evaluates non-zero B-splines derivatives at a given coordinate
+         *
+         * The derivatives are computed for every B-splines with support at the given coordinate x. There are only (degree+1)
+         * B-splines which are non-zero at any given point. It is these B-splines which are derivated.
+         * A spline approximation of a derivative at coordinate x is a linear
+         * combination of those B-splines derivatives weigthed with splines coefficients of the spline-transformed
```
EmilyBourne, line 214:
```suggestion
         * combination of those B-splines derivatives weighted with splines coefficients of the spline-transformed
```

include/ddc/kernels/splines/bsplines_non_uniform.hpp
```diff
@@ -70,125 +95,301 @@ class NonUniformBSplines : detail::NonUniformBSplinesBase
...
+         * @param[out] derivs The derivatives of the B-splines evaluated at coordinate x. It has to be a 1D mdspan with (degree+1) elements.
+         * @param[in] x The coordinate where B-splines derivatives are evaluated.
+         * @return The index of the first B-spline which is derivated.
+         */
         KOKKOS_INLINE_FUNCTION discrete_element_type
         eval_deriv(DSpan1D derivs, ddc::Coordinate<Tag> const& x) const;
 
+        /** @brief Evaluates non-zero B-splines values and \f$n\f$ derivatives at a given coordinate
+         *
+         * The values and derivatives are computed for every B-splines with support at the given coordinate x. There are only (degree+1)
```
EmilyBourne, line 226:
```suggestion
         * The values and derivatives are computed for every B-spline with support at the given coordinate x. There are only (degree+1)
```

include/ddc/kernels/splines/bsplines_non_uniform.hpp
```diff
@@ -70,125 +95,301 @@ class NonUniformBSplines : detail::NonUniformBSplinesBase
...
+         * @param[out] values The values of the B-splines evaluated at coordinate x. It has to be a 1D mdspan with (degree+1) elements.
+         * @param[in] x The coordinate where B-splines are evaluated.
+         * @return The index of the first B-spline which is evaluated.
+         */
         KOKKOS_INLINE_FUNCTION discrete_element_type
         eval_basis(DSpan1D values, ddc::Coordinate<Tag> const& x) const;
 
+        /** @brief Evaluates non-zero B-splines derivatives at a given coordinate
+         *
+         * The derivatives are computed for every B-splines with support at the given coordinate x. There are only (degree+1)
```
EmilyBourne, line 211:
```suggestion
         * The derivatives are computed for every B-spline with support at the given coordinate x. There are only (degree+1)
```

include/ddc/kernels/splines/bsplines_non_uniform.hpp
```diff
@@ -70,125 +95,301 @@ class NonUniformBSplines : detail::NonUniformBSplinesBase
...
+         * @param[in] x The coordinate where B-splines derivatives are evaluated.
+         * @return The index of the first B-spline which is derivated.
+         */
         KOKKOS_INLINE_FUNCTION discrete_element_type
         eval_deriv(DSpan1D derivs, ddc::Coordinate<Tag> const& x) const;
 
+        /** @brief Evaluates non-zero B-splines values and \f$n\f$ derivatives at a given coordinate
+         *
+         * The values and derivatives are computed for every B-splines with support at the given coordinate x. There are only (degree+1)
+         * B-splines which are non-zero at any given point. It is these B-splines which are derivated.
```
EmilyBourne, line 227:
```suggestion
         * B-splines which are non-zero at any given point. It is these B-splines which are evaluated and derivated.
```

include/ddc/kernels/splines/bsplines_non_uniform.hpp
```diff
@@ -70,125 +95,301 @@ class NonUniformBSplines : detail::NonUniformBSplinesBase
...
+         * A spline approximation of a derivative at coordinate x is a linear
+         * combination of those B-splines derivatives weigthed with splines coefficients of the spline-transformed
```
EmilyBourne, line 229:
```suggestion
         * combination of those B-splines derivatives weighted with the spline coefficients of the spline-transformed
```

include/ddc/kernels/splines/bsplines_non_uniform.hpp
```diff
@@ -70,125 +95,301 @@ class NonUniformBSplines : detail::NonUniformBSplinesBase
...
 
-        /// @brief Construct a `Impl` using a pair of iterators.
+        /** @brief Constructs a Impl by iterating over a set of break points from begin to end.
+         *
+         * The provided break points describe the separation between the cells on which the polynomials
+         * comprising a spline are defined. They are used to build a set of knots. There are 2*degree more
+         * knots than break points. The knots are defined as follows:
+         * @f$ k_i = b_0 \forall 0 \leq i < d @f$
+         * @f$ k_{i+d} = b_i \forall 0 \leq i < n_b @f$
+         * @f$ k_{i+d+n_b} = b_{n_b} \forall 0 \leq i < d @f$
```
EmilyBourne, line 143: I have used `@f$` as this is the convention in Gysela but below you use `\f$`. I don't know what the usual convention is in DDC please use the correct version

include/ddc/kernels/splines/bsplines_non_uniform.hpp
```diff
@@ -70,125 +95,301 @@ class NonUniformBSplines : detail::NonUniformBSplinesBase
...
+         *
+         * @return The number of elements necessary to construct a spline representation of a function.
+         */
         KOKKOS_INLINE_FUNCTION std::size_t size() const noexcept
         {
             return degree() + ncells();
         }
 
-        /// Returns the discrete domain including ghost bsplines
+        /** @brief Returns the discrete domain including eventual additionnal bsplines in the periodic case. See size().
```
EmilyBourne, line 355:
```suggestion
        /** @brief Returns the discrete domain including eventual additional B-splines in the periodic case. See size().
```

include/ddc/kernels/splines/bsplines_non_uniform.hpp
```diff
@@ -70,125 +95,301 @@ class NonUniformBSplines : detail::NonUniformBSplinesBase
...
+         *
+         * @return The discrete domain including eventual additionnal bsplines.
```
EmilyBourne, line 357:
```suggestion
         * @return The discrete domain including eventual additional B-splines.
```

include/ddc/kernels/splines/bsplines_non_uniform.hpp
blegouix: I added the @see but I am not sure if it fulfills your asking.

include/ddc/kernels/splines/bsplines_non_uniform.hpp
EmilyBourne: "splines coefficients" still present

EmilyBourne: https://github.com/CExA-project/ddc/pull/415#discussion_r1572057422

include/ddc/kernels/splines/bsplines_non_uniform.hpp
EmilyBourne: Yes this looks good. It correctly generates a link to the more complete docstring. This still needs updating though as it contains unnecessary information. I would recommend something like:
```suggestion
         * Constructs a Impl by iterating over a list of break points. Internally this constructor calls the constructor
         * ... . More details of the implementation can be found there.
```
Or just repeat the information from the other docstring here

include/ddc/kernels/splines/bsplines_non_uniform.hpp
blegouix: Sorry I think I restored the Ss instead of removing ^^'

include/ddc/kernels/splines/bsplines_uniform.hpp
```diff
@@ -101,88 +126,233 @@ class UniformBSplines : detail::UniformBSplinesBase
...
             assert(values.size() == degree() + 1);
             return eval_basis(values, x, degree());
         }
 
+        /** @brief Evaluates non-zero B-splines derivatives at a given coordinate
+         *
+         * The derivatives are computed for every B-spline with support at the given coordinate x. There are only (degree+1)
+         * B-splines which are non-zero at any given point. It is these B-splines which are derivated.
+         * A spline approximation of a derivative at coordinate x is a linear
+         * combination of those B-splines derivatives weighted with the splines coefficients of the spline-transformed
```
EmilyBourne, line 191: splines coefficients

include/ddc/kernels/splines/bsplines_uniform.hpp
```diff
@@ -101,88 +126,233 @@ class UniformBSplines : detail::UniformBSplinesBase
...
+         */
         KOKKOS_INLINE_FUNCTION discrete_element_type
         eval_deriv(DSpan1D derivs, ddc::Coordinate<Tag> const& x) const;
 
+        /** @brief Evaluates non-zero B-splines values and \f$n\f$ derivatives at a given coordinate
+         *
+         * The values and derivatives are computed for every B-spline with support at the given coordinate x. There are only (degree+1)
+         * B-splines which are non-zero at any given point. It is these B-splines which are evaluated and derivated.
+         * A spline approximation of a derivative at coordinate x is a linear
+         * combination of those B-splines derivatives weighted with splines coefficients of the spline-transformed
```
EmilyBourne, line 206: splines coefficients

EmilyBourne:
Something needs to be done for the integrals I see multiple possible solutions:
1. Do nothing. The error will be raised every time someone touches one of these files. This is a bad solution.
2. Change the result type to be explicit. The types of the arguments can still be things like `discrete_domain_type` as these follow the class specification so they still work
3. Move the definition into the class.

We should choose 2 or 3. My preference would be 2 as it is a minor change and keeps the class declaration easy to read. If you prefer 3 then I think all definitions should be moved not just this one

blegouix:
I propose to keep the `integrals` fix for another MR

EmilyBourne:
> I propose to keep the `integrals` fix for another MR to focus on documentation. If you agree I'll open the issue

I think it depends on the chosen solution. Solution 2 is one line and could/should be done here, but solution 3 is bigger and should indeed be its own issue

blegouix:
`error: identifier "discrete_domain_type" is undefined`

I gave it a try :)

EmilyBourne:
> `error: identifier "discrete_domain_type" is undefined`
> 
> I gave it a try :)

It compiles fine on my computer. Maybe you misunderstood. My suggestion is to use:
```cpp
        template <class Layout, class MemorySpace2>
        KOKKOS_INLINE_FUNCTION ddc::ChunkSpan<double, ddc::DiscreteDomain<DDim>, Layout, MemorySpace2>
        integrals(
                ddc::ChunkSpan<double, discrete_domain_type, Layout, MemorySpace2> int_vals) const;
```
for the declaration and:
```cpp
template <class Tag, std::size_t D>
template <class DDim, class MemorySpace>
template <class Layout, class MemorySpace2>
KOKKOS_INLINE_FUNCTION ddc::ChunkSpan<double, ddc::DiscreteDomain<DDim>, Layout, MemorySpace2>
NonUniformBSplines<Tag, D>::Impl<DDim, MemorySpace>::integrals(
        ddc::ChunkSpan<double, discrete_domain_type, Layout, MemorySpace2> int_vals) const
```
for the definition

include/ddc/kernels/splines/bsplines_non_uniform.hpp
```diff
@@ -70,125 +95,302 @@ class NonUniformBSplines : detail::NonUniformBSplinesBase
...
-        explicit Impl(Impl<DDim, OriginMemorySpace> const& impl)
-            : m_domain(impl.m_domain)
-            , m_nknots(impl.m_nknots)
-        {
-        }
-
-        /// @brief Construct a `Impl` using a brace-list, i.e. `Impl bsplines({0., 1.})`
+        /** @brief Constructs an Impl using a brace-list, i.e. `Impl bsplines({0., 1.})`
+         *
+         * Constructs a Impl by iterating over a list of break points. Internally this constructor calls the constructor
```
EmilyBourne, line 114:
```suggestion
         * Constructs an Impl by iterating over a list of break points. Internally this constructor calls the constructor
```

include/ddc/kernels/splines/bsplines_non_uniform.hpp
```diff
@@ -70,125 +95,302 @@ class NonUniformBSplines : detail::NonUniformBSplinesBase
...
+         */
         explicit Impl(std::initializer_list<ddc::Coordinate<Tag>> breaks)
             : Impl(breaks.begin(), breaks.end())
         {
         }
 
-        /// @brief Construct a `Impl` using a C++20 "common range".
+        /** @brief Constructs an Impl using a std::vector.
+         *
+         * Constructs a Impl by iterating over a list of break points. Internally this constructor calls the constructor
```
EmilyBourne, line 126:
```suggestion
         * Constructs an Impl by iterating over a list of break points. Internally this constructor calls the constructor
```

include/ddc/kernels/splines/bsplines_non_uniform.hpp
```diff
@@ -70,125 +95,302 @@ class NonUniformBSplines : detail::NonUniformBSplinesBase
...
+         *
+         * @param breaks The std::vector of the coordinates of break points.
+         */
         explicit Impl(std::vector<ddc::Coordinate<Tag>> const& breaks)
             : Impl(breaks.begin(), breaks.end())
         {
         }
 
-        /// @brief Construct a `Impl` using a pair of iterators.
+        /** @brief Constructs a Impl by iterating over a set of break points from begin to end.
```
EmilyBourne, line 136:
```suggestion
        /** @brief Constructs an Impl by iterating over a set of break points from begin to end.
```

include/ddc/kernels/splines/bsplines_non_uniform.hpp
```diff
@@ -70,125 +95,302 @@ class NonUniformBSplines : detail::NonUniformBSplinesBase
...
+         *
+         * This constructor makes the knots accessible via a DiscreteSpace.
+         *
+         * @param breaks_begin The iterator which points at the beginning of the break points.
+         * @param breaks_end The iterator which points at the end of the break points.
+         */
         template <class RandomIt>
         Impl(RandomIt breaks_begin, RandomIt breaks_end);
 
+        /** @brief Copy-constructs from another Impl with different Kokkos memory space
```
EmilyBourne, line 154:
```suggestion
        /** @brief Copy-constructs from another Impl with a different Kokkos memory space
```

include/ddc/kernels/splines/bsplines_non_uniform.hpp
```diff
@@ -70,125 +95,302 @@ class NonUniformBSplines : detail::NonUniformBSplinesBase
...
+         */
         Impl(Impl&& x) = default;
 
+        /// @brief Destructs
         ~Impl() = default;
 
+        /** @brief Copy-assigns
+         *
+         * @param x A reference to another Impl
+         * @return A reference to the copy Impl
```
EmilyBourne, line 183:
```suggestion
         * @return A reference to the copied Impl
```

include/ddc/kernels/splines/bsplines_non_uniform.hpp
```diff
@@ -70,125 +95,302 @@ class NonUniformBSplines : detail::NonUniformBSplinesBase
...
+         * initial discrete function.
+         *
+         * @param[out] values The values of the B-splines evaluated at coordinate x. It has to be a 1D mdspan with (degree+1) elements.
+         * @param[in] x The coordinate where B-splines are evaluated.
+         * @return The index of the first B-spline which is evaluated.
+         */
         KOKKOS_INLINE_FUNCTION discrete_element_type
         eval_basis(DSpan1D values, ddc::Coordinate<Tag> const& x) const;
 
+        /** @brief Evaluates non-zero B-splines derivatives at a given coordinate
```
EmilyBourne, line 209:
```suggestion
        /** @brief Evaluates non-zero B-spline derivatives at a given coordinate
```

include/ddc/kernels/splines/bsplines_non_uniform.hpp
```diff
@@ -70,125 +95,302 @@ class NonUniformBSplines : detail::NonUniformBSplinesBase
...
+         *
+         * The derivatives are computed for every B-spline with support at the given coordinate x. There are only (degree+1)
+         * B-splines which are non-zero at any given point. It is these B-splines which are derivated.
+         * A spline approximation of a derivative at coordinate x is a linear
+         * combination of those B-splines derivatives weighted with the spline coefficients of the spline-transformed
```
EmilyBourne, line 214:
```suggestion
         * combination of those B-spline derivatives weighted with the spline coefficients of the spline-transformed
```

include/ddc/kernels/splines/bsplines_non_uniform.hpp
```diff
@@ -70,125 +95,302 @@ class NonUniformBSplines : detail::NonUniformBSplinesBase
...
+         * initial discrete function.
+         *
+         * @param[out] derivs The derivatives of the B-splines evaluated at coordinate x. It has to be a 1D mdspan with (degree+1) elements.
+         * @param[in] x The coordinate where B-splines derivatives are evaluated.
+         * @return The index of the first B-spline which is derivated.
+         */
         KOKKOS_INLINE_FUNCTION discrete_element_type
         eval_deriv(DSpan1D derivs, ddc::Coordinate<Tag> const& x) const;
 
+        /** @brief Evaluates non-zero B-splines values and \f$n\f$ derivatives at a given coordinate
```
EmilyBourne, line 224:
```suggestion
        /** @brief Evaluates non-zero B-spline values and \f$n\f$ derivatives at a given coordinate
```

include/ddc/kernels/splines/bsplines_non_uniform.hpp
```diff
@@ -70,125 +95,302 @@ class NonUniformBSplines : detail::NonUniformBSplinesBase
...
+        KOKKOS_INLINE_FUNCTION ddc::
+                ChunkSpan<double, ddc::DiscreteDomain<DDim>, Layout, MemorySpace2>
+                integrals(ddc::ChunkSpan<double, discrete_domain_type, Layout, MemorySpace2>
+                                  int_vals) const;
+
+        /** @brief Returns the coordinate of the knot corresponding to the given index.
+         *
+         * Returns the coordinate of the knot corresponding to the given index. The domain
+         * over which the B-splines are defined is comprised of ncells+1 knots however there are a total of
+         * ncells+1+2*degree knots. The additional knots which control the shape of the B-splines near the
```
EmilyBourne, line 259:
```suggestion
         * over which the B-splines are defined is comprised of ncells+1 break points however there are a total of
         * ncells+1+2*degree knots. The additional knots which control the shape of the B-splines near the
```

include/ddc/kernels/splines/bsplines_uniform.hpp
```diff
@@ -28,33 +28,59 @@ struct UniformBSplinesBase
 
 } // namespace detail
 
+/**
+ * The type of a uniform B-splines 1D basis.
```
EmilyBourne, line 32: See above

include/ddc/kernels/splines/bsplines_uniform.hpp
```diff
@@ -28,33 +28,59 @@ struct UniformBSplinesBase
...
+ *
+ * Knots for uniform B-splines basis are uniformly distributed (the associated discrete dimension
```
EmilyBourne, line 34:
```suggestion
 * Knots for uniform B-splines are uniformly distributed (the associated discrete dimension 
```

include/ddc/kernels/splines/bsplines_uniform.hpp
```diff
@@ -101,88 +126,234 @@ class UniformBSplines : detail::UniformBSplinesBase
...
+         */
         KOKKOS_INLINE_FUNCTION discrete_element_type
         eval_deriv(DSpan1D derivs, ddc::Coordinate<Tag> const& x) const;
 
+        /** @brief Evaluates non-zero B-splines values and \f$n\f$ derivatives at a given coordinate
+         *
+         * The values and derivatives are computed for every B-spline with support at the given coordinate x. There are only (degree+1)
+         * B-splines which are non-zero at any given point. It is these B-splines which are evaluated and derivated.
+         * A spline approximation of a derivative at coordinate x is a linear
+         * combination of those B-splines derivatives weighted with spline coefficients of the spline-transformed
```
EmilyBourne, line 206:
```suggestion
         * combination of those B-spline derivatives weighted with spline coefficients of the spline-transformed
```

EmilyBourne (changes requested):
These explanations seem clear to me. Well done. I am happy to approve once my grammar comments have been handled.

It could be useful to have someone less familiar with the code look over the docs though as I am probably not the target audience for these

EmilyBourne (dismissed):
Looks good

include/ddc/kernels/splines/bsplines_non_uniform.hpp
```diff
@@ -70,125 +95,302 @@ class NonUniformBSplines : detail::NonUniformBSplinesBase
...
+         *
+         * Returns the coordinate of the knot corresponding to the given index. The domain
+         * over which the B-splines are defined is comprised of ncells+1 break points however there are a total of
+         * ncells+1+2*degree knots. The additional knots which control the shape of the B-splines near the
+         * boundary are added before and after the break points. The knot index is therefore in the interval [-degree, ncells+degree]
+         *
+         * @param[in] knot_idx Integer identifying index of the knot.
+         * @return Coordinate of the knot.
+         */
         KOKKOS_INLINE_FUNCTION ddc::Coordinate<Tag> get_knot(int knot_idx) const noexcept
```
tpadioleau, line 265: For a future MR, I guess these `get_knot*` functions can be simplified. A single function making the conversion between a `DiscreteElement<BSplines>` and `DiscreteElement<Knots>` should be enough.

include/ddc/kernels/splines/bsplines_non_uniform.hpp
```diff
@@ -70,125 +95,302 @@ class NonUniformBSplines : detail::NonUniformBSplinesBase
...
+         * @return Coordinate of the knot.
+         */
         KOKKOS_INLINE_FUNCTION ddc::Coordinate<Tag> get_support_knot_n(
                 discrete_element_type const& ix,
                 int n) const
         {
             return ddc::coordinate(ddc::DiscreteElement<mesh_type>(ix.uid() + n));
         }
 
+        /** @brief Returns the coordinate of the lower bound of the domain on which the B-splines are defined.
```
tpadioleau, line 316:
```suggestion
        /** @brief Returns the coordinate of the first break point of the domain on which the B-splines are defined.
```

include/ddc/kernels/splines/bsplines_non_uniform.hpp
```diff
@@ -70,125 +95,302 @@ class NonUniformBSplines : detail::NonUniformBSplinesBase
...
+         *
+         * @return Coordinate of the lower bound of the domain.
+         */
         KOKKOS_INLINE_FUNCTION ddc::Coordinate<Tag> rmin() const noexcept
         {
             return get_knot(0);
         }
 
+        /** @brief Returns the coordinate of the upper bound of the domain on which the B-splines are defined.
```
tpadioleau, line 325:
```suggestion
        /** @brief Returns the coordinate of the last break point of the domain on which the B-splines are defined.
```

include/ddc/kernels/splines/bsplines_non_uniform.hpp
```diff
@@ -70,125 +95,302 @@ class NonUniformBSplines : detail::NonUniformBSplinesBase
...
+        /** @brief Evaluates non-zero B-splines at a given coordinate.
+         *
+         * The values are computed for every B-spline with support at the given coordinate x. There are only (degree+1)
+         * B-splines which are non-zero at any given point. It is these B-splines which are evaluated.
+         * This can be useful to calculate a spline approximation of a function. A spline approximation at coordinate x
+         * is a linear combination of these B-spline evaluations weighted with the spline coefficients of the spline-transformed
+         * initial discrete function.
+         *
+         * @param[out] values The values of the B-splines evaluated at coordinate x. It has to be a 1D mdspan with (degree+1) elements.
+         * @param[in] x The coordinate where B-splines are evaluated.
```
tpadioleau, line 203: Shall we mention the range of validity for `x` ? that is the range of break points right ?

include/ddc/kernels/splines/bsplines_non_uniform.hpp
```diff
@@ -70,125 +95,302 @@ class NonUniformBSplines : detail::NonUniformBSplinesBase
...
             : Impl(breaks.begin(), breaks.end())
         {
         }
 
-        /// @brief Construct a `Impl` using a pair of iterators.
+        /** @brief Constructs an Impl by iterating over a set of break points from begin to end.
+         *
+         * The provided break points describe the separation between the cells on which the polynomials
+         * comprising a spline are defined. They are used to build a set of knots. There are 2*degree more
+         * knots than break points. The knots are defined as follows:
```
tpadioleau, line 140:
```suggestion
         * knots than break points. In the non-periodic case the knots are defined as follows:
```

tpadioleau (commented):
@blegouix @EmilyBourne thanks a lot for this work! I have minor suggestions/questions.

If you apply some suggestions, please make also the change in the file `bsplines_uniform.hpp` if they are meaningful.

include/ddc/kernels/splines/bsplines_non_uniform.hpp
EmilyBourne: Indeed. This is part of #416

EmilyBourne: I don't think we can convert from a `DiscreteElement<BSplines>` to a `DiscreteElement<Knots>` as there are d+2 knots associated with each bspline but the knot samplings should be used to replace this function by a call to `coordinate`

include/ddc/kernels/splines/bsplines_non_uniform.hpp
EmilyBourne: That is the correct range. I agree that this could be useful :+1:

include/ddc/kernels/splines/bsplines_non_uniform.hpp
EmilyBourne: I agree. 
```
         * In the periodic case the knots are defined as:
         * \f$ k_i = b_{n_b-d+i} \forall 0 \leq i < d \f$
         * \f$ k_{i+d} = b_i \forall 0 \leq i \leq n_b \f$
         * \f$ k_{i+d+n_b} = b_{i+1} \forall 0 \leq i < d \f$
```

include/ddc/kernels/splines/bsplines_non_uniform.hpp
```diff
@@ -70,125 +95,302 @@ class NonUniformBSplines : detail::NonUniformBSplinesBase
...
+         * \f$ k_i = b_0 \forall 0 \leq i < d \f$
+         * \f$ k_{i+d} = b_i \forall 0 \leq i < n_b \f$
+         * \f$ k_{i+d+n_b} = b_{n_b} \forall 0 \leq i < d \f$
```
EmilyBourne, line 143:
```suggestion
         * \f$ k_{i+d+n_b} = b_{n_b-1} \forall 0 \leq i < d \f$
```
Just realised that there is an index problem here

include/ddc/kernels/splines/bsplines_non_uniform.hpp
tpadioleau: I was thinking of keeping only `get_first_knot` that would return the first knot `DiscreteElement<Knots>` in the support of the b-spline ? The call to `coordinate` would be up to the user ?

EmilyBourne: I think `get_last_support_knot` can also be useful. I agree that it makes sense for these 2 functions to return `DiscreteElement`s. `get_knot` is equivalent to `coordinate` and `get_support_knot_n` is then equivalent to `get_first_support_knot() + DiscreteVector<Knots>(n)`

include/ddc/kernels/splines/bsplines_non_uniform.hpp
```diff
@@ -28,34 +28,59 @@ struct NonUniformBSplinesBase
...
+/**
+ * The type of a non-uniform 1D spline basis (B-spline).
+ *
+ * Knots for non-uniform B-splines are non-uniformly distributed (no assumption is made on the uniformity of their distribution,
+ * the associated discrete dimension is a NonUniformPointSampling).
+ *
+ * @tparam Tag The tag identifying the continuous dimension on which the support of the B-spline functions are defined.
+ * @tparam D The degree of the B-splines.
+ */
 template <class Tag, std::size_t D>
```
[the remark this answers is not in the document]
blegouix: Yeah looks good

tpadioleau (approved):
Great work!

PR #420, opened 2024-04-19
Allow the reordering of a DiscreteDomain

### LINKED ISSUE #419: Allow the reordering of a DiscreteDomain
DDC contains a reordering constructor for a `DiscreteDomain`:
https://github.com/CExA-project/ddc/blob/9281e03ae9bf5b292aea65fb7d46fb3765c730da/include/ddc/discrete_domain.hpp#L260-L265

however the assignment operator is missing.

EmilyBourne:
Add a function to initialise a `DiscreteDomain` from a `DiscreteDomain` with a different ordering. Fixes #419

### DIFF
--- a/include/ddc/discrete_domain.hpp
+++ b/include/ddc/discrete_domain.hpp
@@ -101,7 +101,7 @@ public:
     KOKKOS_DEFAULTED_FUNCTION DiscreteDomain& operator=(DiscreteDomain&& x) = default;
 
     /**
-     * @brief Copy a DiscreteDomain by reordering.
+     * @brief Copy a DiscreteDomain by reordering and slicing.
      *
      * An assign operator to build a DiscreteDomain from another compatible domain.
      * A domain is compatible if it either contains the same dimensions as this
--- a/include/ddc/discrete_domain.hpp
+++ b/include/ddc/discrete_domain.hpp
@@ -100,6 +100,25 @@ public:
 
     KOKKOS_DEFAULTED_FUNCTION DiscreteDomain& operator=(DiscreteDomain&& x) = default;
 
+    /**
+     * @brief Copy a DiscreteDomain by reordering.
+     *
+     * An assign operator to build a DiscreteDomain from another compatible domain.
+     * A domain is compatible if it either contains the same dimensions as this
+     * domain (even if they are in a different order) or if it contains at
+     * the dimensions of this domain plus some additional dimensions which will
+     * be unused here.
+     *
+     * @param domain A compatible domain.
+     */
+    template <class... ODDims>
+    DiscreteDomain& KOKKOS_FUNCTION operator=(DiscreteDomain<ODDims...> const& domain)
+    {
+        m_element_begin = DiscreteElement<DDims...>(domain.front());
+        m_element_end = DiscreteElement<DDims...>(domain.back());
+        return *this;
+    }
+
     template <class... ODims>
     KOKKOS_FUNCTION constexpr bool operator==(DiscreteDomain<ODims...> const& other) const
     {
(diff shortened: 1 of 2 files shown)


tpadioleau:
I think it should already be addressed by existing constructors. Can you provide an example that does not compile ?

tpadioleau:
> > I think it should already be addressed by existing constructors. Can you provide an example that does not compile ?
> 
> Hmm, it seems that you are right. The tests I have added already work on the main branch. The assignment operator doesn't work:
> ```cpp
> DDomZYX const dom_z_y_x = dom_x_y_z;
> ```
> but maybe this is intentional?

I see, I do not see a good reason to not support this assignment operator, it seems to be an equivalent of the reordering & slicing constructor.

tpadioleau:
Actually this syntax is not related to the assignment operator. It only involves either the constructor or a cast operator.

That being said, I think it is worth having an equivalent assignement operator relative to the slicing and reordering constructor.

tests/discrete_domain.cpp
```diff
@@ -234,3 +237,80 @@ TEST(ProductMDomainTest, SliceDomainXToolate)
...
+    DDomZ const dom_z(lbound_z, nelems_z);
+    DDomXYZ const dom_x_y_z(dom_x, dom_y, dom_z);
+    DDomZYX const dom_z_y_x = dom_x_y_z;
+    EXPECT_EQ(ddc::select<DDimX>(dom_x_y_z.front()), ddc::select<DDimX>(dom_z_y_x.front()));
+    EXPECT_EQ(ddc::select<DDimY>(dom_x_y_z.front()), ddc::select<DDimY>(dom_z_y_x.front()));
+    EXPECT_EQ(ddc::select<DDimZ>(dom_x_y_z.front()), ddc::select<DDimZ>(dom_z_y_x.front()));
+    EXPECT_EQ(ddc::select<DDimX>(dom_x_y_z.back()), ddc::select<DDimX>(dom_z_y_x.back()));
+    EXPECT_EQ(ddc::select<DDimY>(dom_x_y_z.back()), ddc::select<DDimY>(dom_z_y_x.back()));
+    EXPECT_EQ(ddc::select<DDimZ>(dom_x_y_z.back()), ddc::select<DDimZ>(dom_z_y_x.back()));
+}
```
EmilyBourne, line 289: @tpadioleau These 2 tests don't compile however this is because the transpose constructor is marked `explicit`. I think this is something that should be kept so these tests should be removed. Do you agree?

tpadioleau: Should work with `DDomZYX const dom_z_y_x(dom_x_y_z)`

EmilyBourne: Yes. That is the test that I added above

tpadioleau: Sorry I did not see the other tests. Yes I would remove them.

EmilyBourne: Done [6ff8339](https://github.com/CExA-project/ddc/pull/420/commits/6ff8339308a8f2bda1c85e65a13695276cb805ee)

include/ddc/discrete_domain.hpp
```diff
@@ -100,6 +100,25 @@ class DiscreteDomain
 
     KOKKOS_DEFAULTED_FUNCTION DiscreteDomain& operator=(DiscreteDomain&& x) = default;
 
+    /**
+     * @brief Copy a DiscreteDomain by reordering.
```
tpadioleau, line 104:
```suggestion
     * @brief Copy a DiscreteDomain by reordering and slicing.
```

tpadioleau (changes requested):
Thanks, do we really need both the 2D and 3D versions then ?

EmilyBourne:
> Thanks, do we really need both the 2D and 3D versions then ?

I don't think so, 3D should be sufficient. I will remove the 2D version

PR #428, opened 2024-04-24
[Splines] Remove NATURAL boundary condition

blegouix:
`ddc::BoundCond::NATURAL` is used nowhere, not tested and even not fully implement.

`n_user_input()` is also removed because the returned value was different from `n_boundary_equations()`


EmilyBourne (approved):

tpadioleau (approved):
If @EmilyBourne agrees then ok for me

PR #430, opened 2024-04-25
SplineEvaluator doc

blegouix:


blegouix:
I remove the draft status but only for the SplineEvaluator (not for 2D). The reason is the remaining functions I have to document are closely linked to their 1D pendant but they are twice more numerous. So I prefer to validate the 1D functions before doing the copy-pastes.

include/ddc/kernels/splines/spline_evaluator.hpp
```diff
@@ -111,26 +150,79 @@ class SplineEvaluator
...
+     * @brief Evaluate 1D spline function (described by its spline coefficients) at a given coordinate.
+     *
+     * The spline coefficients represent a 1D spline function in a B-splines (basis splines). They can be obtained using a SplineBuilder. 
+     *
+     * Remark: calling SplineBuilder then SplineEvaluator corresponds to a spline interpolation.
+     *
+     * @param coord_eval The coordinate where the spline is evaluated. Note that only the component along the dimension of interest is used.
+     * @param spline_coef A Chunkspan storing the 1D spline coefficients.
+     *
+     * @return The value of the spline function at the desired coordinate.
```
EmilyBourne, line 224:
```suggestion
     * @brief Evaluate a 1D spline function (described by its spline coefficients) at a given coordinate.
     *
     * The spline coefficients represent a 1D spline function defined on B-splines (basis splines).
     * They can be obtained via various methods, such as using a SplineBuilder. 
     *
     * Remark: calling SplineBuilder then SplineEvaluator corresponds to a spline interpolation.
     *
     * @param coord_eval The coordinate where the spline is evaluated. Note that only the component along the dimension of interest is used.
     * @param spline_coef A ChunkSpan storing the 1D spline coefficients.
     *
     * @return The value of the spline function at the desired coordinate. 
```

include/ddc/kernels/splines/spline_evaluator.hpp
```diff
@@ -140,6 +232,25 @@ class SplineEvaluator
         return eval(coord_eval, spline_coef);
     }
 
+    /**
+     * @brief Evaluate spline function (described by its spline coefficients) on a mesh.
+     *
+     * The spline coefficients represent a spline function in a B-splines (basis splines). They can be obtained using a SplineBuilder.
```
EmilyBourne, line 238: Comments from above also apply here

include/ddc/kernels/splines/spline_evaluator.hpp
```diff
@@ -140,6 +232,25 @@ class SplineEvaluator
...
+     *
+     * This is not a multidimensional evaluation. This is a batched 1D evaluation. It means for each coordinate of coords_eval, the evaluation is performed with the 1D set of spline coefficents of spline_coef identified with the same batch_domain_type::discrete_element_type which identifies the given coordinate of coords_eval (or the corresponding value of spline_eval which is computed).
```
EmilyBourne, line 240: As discussed in #409 please break up this line.

> It means for each coordinate of coords_eval, the evaluation is performed with the 1D set of spline coefficents of spline_coef identified with the same batch_domain_type::discrete_element_type which identifies the given coordinate of coords_eval (or the corresponding value of spline_eval which is computed).

This sentence is huge and hard to follow. I'm going to try and make it easier to read but I would appreciate a second opinion

> This means that for each slice of coordinates identified by a batch_domain_type::discrete_element_type, the evaluation is performed with the 1D set of spline coefficients identified by the same batch_domain_type::discrete_element_type.

include/ddc/kernels/splines/spline_evaluator.hpp
```diff
@@ -140,6 +232,25 @@ class SplineEvaluator
...
+     *
+     * @param[out] spline_eval The values of the spline function at the desired coordinates. For practical reasons those are
+     * stored in a Chunkspan defined on a batched_interpolation_domain_type. Note that the coordinates of the
+     * points represented by this domains are unused and irrelevant (but the points themselves (DiscreteElement) are used to select
+     * the set of 1D spline coefficients retained to perform the evaluation).
+     * @param[in] coords_eval The coordinates where the spline is evaluated. Those are
+     * stored in a Chunkspan defined on a batched_interpolation_domain_type. Note that the coordinates of the
+     * points represented by this domains are unused and irrelevant (but the points themselves (DiscreteElement) are used to select
+     * the set of 1D spline coefficients retained to perform the evaluation).
+     * @param[in] spline_coef A Chunkspan storing the spline coefficients.
```
EmilyBourne, line 252:
```suggestion
     * @param[out] spline_eval The values of the spline function at the desired coordinates. For practical reasons those are
     * stored in a ChunkSpan defined on a batched_interpolation_domain_type. Note that the coordinates of the
     * points represented by this domains are unused and irrelevant (but the points themselves (DiscreteElement) are used to select
     * the set of 1D spline coefficients retained to perform the evaluation).
     * @param[in] coords_eval The coordinates where the spline is evaluated. Those are
     * stored in a ChunkSpan defined on a batched_interpolation_domain_type. Note that the coordinates of the
     * points represented by this domain are unused and irrelevant (but the points themselves (DiscreteElement) are used to select
     * the set of 1D spline coefficients retained to perform the evaluation).
     * @param[in] spline_coef A ChunkSpan storing the spline coefficients.
```

include/ddc/kernels/splines/spline_evaluator.hpp
```diff
@@ -168,6 +279,16 @@ class SplineEvaluator
...
 
+    /**
+     * @brief Derivate 1D spline function (described by its spline coefficients) at a given coordinate.
+     *
+     * The spline coefficients represent a 1D spline function in a B-splines (basis splines). They can be obtained using a SplineBuilder. 
+     *
+     * @param coord_eval The coordinate where the spline is derivated. Note that only the component along the dimension of interest is used.
+     * @param spline_coef A Chunkspan storing the 1D spline coefficients.
+     *
+     * @return The derivative of the spline function at the desired coordinate.
```
EmilyBourne, line 290:
```suggestion
     * @brief Differentiate a 1D spline function (described by its spline coefficients) at a given coordinate.
     *
     * The spline coefficients represent a 1D spline function in a B-splines (basis splines). They can be obtained using a SplineBuilder.
     *
     * @param coord_eval The coordinate where the spline is differentiated. Note that only the component along the dimension of interest is used.
     * @param spline_coef A Chunkspan storing the 1D spline coefficients.
     *
     * @return The derivative of the spline function at the desired coordinate. 
```

> They can be obtained using a SplineBuilder.

This sentence is repeated multiple times but a `SplineBuilder` is not the only way to obtain spline coefficients. I think that either this comment should be removed or it should be clear that this is not the only option.

include/ddc/kernels/splines/spline_evaluator.hpp
```diff
@@ -177,6 +298,23 @@ class SplineEvaluator
         return eval_no_bc<eval_deriv_type>(coord_eval, spline_coef);
     }
 
+    /**
+     * @brief Derivate spline function (described by its spline coefficients) on a mesh.
```
EmilyBourne, line 302: Recurring error in this file:
image

The verb you are looking for is differentiate
image

The noun is indeed derivative though

include/ddc/kernels/splines/spline_evaluator.hpp
EmilyBourne: A similar sentence appears below

include/ddc/kernels/splines/spline_evaluator.hpp
```diff
@@ -206,6 +344,17 @@ class SplineEvaluator
                 });
     }
 
+    /** @brief Perform batched 1D integrations of a spline function (described by its spline coefficients) along the dimension of interest and store results on a subdomain of batch_domain.
```
EmilyBourne, line 347:
```suggestion
    /** @brief Perform batched 1D integrations of spline functions (described by their spline coefficients) along the dimension of interest and store results on a subdomain of batch_domain.
```

include/ddc/kernels/splines/spline_evaluator.hpp
```diff
@@ -206,6 +344,17 @@ class SplineEvaluator
...
+     *
+     * The spline coefficients represent a spline function in a B-splines (basis splines). They can be obtained using a SplineBuilder.
+     *
+     * The integration is not performed in a multidimensional way (in any sense). This is a batched 1D integration. It means for each element of integrals, the integration is performed with the 1D set of spline coefficents of spline_coef identified with the same DiscreteElement.
+     *
+     * @param[out] integrals The integrals of the spline function on the subdomain of batch_domain. For practical reasons those are
+     * stored in a Chunkspan defined on a batch_domain_type. Note that the coordinates of the
+     * points represented by this domains are unused and irrelevant.
+     * @param[in] spline_coef A Chunkspan storing the spline coefficients.
```
EmilyBourne, line 356:
```suggestion
     * stored in a ChunkSpan defined on a batch_domain_type. Note that the coordinates of the
     * points represented by this domains are unused and irrelevant.
     * @param[in] spline_coef A ChunkSpan storing the spline coefficients.
```

`ChunkSpan` not `Chunkspan` as this is a code keyword

EmilyBourne (changes requested):

include/ddc/kernels/splines/spline_evaluator.hpp
```diff
@@ -13,10 +13,23 @@
 
 namespace ddc {
 
+/**
+ * @brief A class to evaluate, derivate or integrate a spline function.
+ *
+ * A class which contains an operator () which can be used to evaluate, derivate or integrate a spline function.
+ *
+ * @tparam ExecSpace The Kokkos execution space on which the spline transform is performed.
+ * @tparam MemorySpace The Kokkos memory space on which the data (interpolation function and splines coefficients) is stored.
```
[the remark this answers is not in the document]
blegouix: I don't understand the question, the data stored to describe the "interpolation function" is the interpolation function.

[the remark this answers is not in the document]
EmilyBourne: Then maybe I have misunderstood what you are trying to explain. What is the "interpolation function"? Is it not the spline? Presumably not

include/ddc/kernels/splines/spline_evaluator.hpp
```diff
@@ -13,10 +13,23 @@
...
+ * @tparam BSplines The discrete dimension representing the B-splines.
+ * @tparam InterpolationMesh The discrete dimension on which interpolation points are defined.
```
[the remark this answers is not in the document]
blegouix: You want to call it evaluation points ?

```
In [engineering](https://en.wikipedia.org/wiki/Engineering) and [science](https://en.wikipedia.org/wiki/Science), one often has a number of data points, obtained by [sampling](https://en.wikipedia.org/wiki/Sampling_(statistics)) or [experimentation](https://en.wikipedia.org/wiki/Experimentation), which represent the values of a function for a limited number of values of the [independent variable](https://en.wikipedia.org/wiki/Dependent_and_independent_variables). It is often required to interpolate; that is, estimate the value of that function for an intermediate value of the independent variable.
```

Maybe this is because we have some problem to agree for a terminology, maybe we do not have the same definiton for an interpolation.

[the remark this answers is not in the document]
blegouix: To me `use a spline builder + spline evaluator to approximate a function at a different set of points` is exactly what an interpolation is.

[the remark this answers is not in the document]
EmilyBourne: I think you are ignoring the case where we have a spline but are not interpolating a known function. For example when we obtain a spline by solving a FEM equation. A spline is itself a function and it is not only useful as an approximation of another function.

In your definition you have:
> one often has a number of data points which represent the values of a function for a limited number of values of the [independent variable](https://en.wikipedia.org/wiki/Dependent_and_independent_variables).

In the case where the spline is not obtained via the SplineBuilder this section does not apply to splines. This is why I don't think the terminology is relevant in this file.

Further even supposing we are using the builder your implementation doesn't force us to evaluate the function at the points that were used to create the spline.


> To me use a spline builder + spline evaluator to approximate a function at a different set of points is exactly what an interpolation is.

I agree but this is not an interpolator it is a spline evaluator and an interpolator is only one way of using the evaluator. I also suspect that it is only an interpolator if these "evaluation_points" match the "interpolation_points". So I think the current definition is accurate in the context of an interpolator but not accurate in any other context

include/ddc/kernels/splines/spline_evaluator.hpp
```diff
@@ -33,37 +46,55 @@ class SplineEvaluator
...
+    /// @brief The type of the Kokkos memory space used by this class.
     using memory_space = MemorySpace;
 
-    using bsplines_type = BSplinesType;
-
-    using left_extrapolation_rule_type = LeftExtrapolationRule;
-    using right_extrapolation_rule_type = RightExtrapolationRule;
-
+    /// @brief The type of the interpolation discrete dimension (discrete dimension of interest) used by this class.
     using interpolation_mesh_type = InterpolationMesh;
```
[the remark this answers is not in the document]
blegouix: This I agree. But then we are going to have different names for the `interpolation_domain_type` in the SplineBuilder and `evaluation_domain_type` in the SplineEvaluator, which looks very confusing to me. This looks like trying hard to "forget" the existence of a symmetry which exists anyway. That's why since the beginning I wanted to stop using the "interpolation" terminology.

I think this is already complicated enough and we should not use two different names for the same thing in the splines.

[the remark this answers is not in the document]
EmilyBourne: > I think this is already complicated enough and we should not use two different names for the same thing in the splines.

The problem is that these are not at all the same thing. They are only the same thing in the context of a semi-Lagrangian interpolation.

In the context of an interpolation where I want to use a spline approximation to refine my grid they are different.

In the context of a spline solution to a FEM equation they are unrelated.

include/ddc/kernels/splines/spline_evaluator.hpp
blegouix: Yes I had some problems to make this part simple while giving the important idea to the reader.

include/ddc/kernels/splines/spline_evaluator.hpp
blegouix: What I designate here with "spline function" is a multidimensional function which is defined on B-splines along dimension of interest and batch dimensions on batch domain. This is why in 1D methods I make a distinction with "1D spline function". Maybe I can add an explaination somewhere, but changing this would require a lot of change for consistency.

EmilyBourne: > Maybe I can add an explaination somewhere

That would be great. I think it is needed as this "multidimensional function" is discrete in most of its dimensions but continuous in one. That is an odd convention which should be explained.

include/ddc/kernels/splines/spline_evaluator.hpp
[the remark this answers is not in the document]
blegouix: It is the evaluated function

include/ddc/kernels/splines/spline_evaluator.hpp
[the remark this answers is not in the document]
blegouix: `In the context of a spline solution to a FEM equation they are unrelated` I don't think I agree because for any use-case they are related in the sense that:

SplineBuilder acts on a function defined on X and returns a function defined on BSplineX.
SplineEvaluator acts on a function defined on BSplineX and returns a function defined on X.

Sure the physical units can be different, but this is the same when we do `iFFT(FFT(f)/k^2)`.

Maybe I miss something or this is not what you had in mind ?

include/ddc/kernels/splines/spline_evaluator.hpp
blegouix: I agree your sentence is better

include/ddc/kernels/splines/spline_evaluator.hpp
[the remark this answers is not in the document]
EmilyBourne: > SplineBuilder acts on a function defined on X and returns a function defined on BSplineX.
> SplineEvaluator acts on a function defined on BSplineX and returns a function defined on X.

This is true on a continuous domain but if we stick to objects in DDC a discrete definition would be:

> SplineBuilder acts on a function defined on IDimX1 and returns a function defined on BSplineX.
> SplineEvaluator acts on a function defined on BSplineX and returns a function defined on IDimX2.


> this is the same when we do `iFFT(FFT(f)/k^2)`

I think this is explains why I think `SplineEvaluator` and `iFFT` are not equivalent. The inverse Fourier transform is not implemented as an evaluation of the Fourier approximation of the function. The Fourier approximation is defined as:
$f(x) = {\sum_{k=-\infty}}^{\infty} c_k \exp(i2\pi x k/P)$
But the Fourier transform is the (continuous) coefficients $c_n$. The inverse Fourier transform is calculated using:
$f(x_n) = 1/N {\sum_{k=0}}^{N-1} c_k \exp(i 2\pi n k/N)$
Because $x_n$ doesn't appear in the RHS you can't use the iFFT to evaluate the function at an arbitrary point.

For splines we have something similar in that:
- the builder is equivalent to `FFT`
    - Builder : $c_k = (B^-1f(x_j))_k$
    - FFT : $c_k = {\sum_{j=0}}^{N-1} f(x_j) \exp(i 2\pi k j/N)$
- the spline coefficients are equivalent to the Fourier transforms
     - Builder : $c_k$
     - FFT : $c_k$
- the evaluator is equivalent to evaluating the approximation
     - Spline Evaluator : $f(x) = {\sum_{i=0}}^N c_n b_i(x)$
     - Fourier Evaluator : $f(x) = {\sum_{k=-\infty}}^{\infty} c_k \exp(i2\pi x k/P)$
- Multiplying the matrix from the builder by a vector of spline coefficients is equivalent to `iFFT`. 
     - Spline inverse transform : $f(x_j) = (B c_i)_j
     - iFFT : $f(x_j) = 1/N {\sum_{k=0}}^{N-1} c_k \exp(i 2\pi j k/N)

It would be possible to implement an inverse Spline Builder just as it would be possible to implement a Fourier evaluator but this is not commonly done because these 2 tools are rarely used to solve the same problem.

In the majority of cases Fourier is used to manipulate values at fixed points. This is useful for (amongst others):
- derivatives (thanks to the properties of Fourier space)
- noise removal (by eliminating coefficients that correspond to highly oscillating modes)

It would be possible to use a spline to calculate a derivative using a Builder(iBuilder(f)) method. From the coefficients $c_{k,d}$ of a d-th degree spline representation you can obtain the coefficients $c'_{k,d-1}$ of the (d-1)-th spline representation of the derivative using $d_x b_{i,d}(x) = d (b_{i,d-1}(x)/(k_{i+d}-k_i) - b_{i+1, d-1}(x)/(k_{i+d+1}-k_{i+1}))$ You would then have to find the coefficients $c'_{k,d}$ somehow. As you can see this is significantly more complicated than just evaluating the function.

Suppressing a signal does make sense using splines. You can set one of the coefficients to 0 to damp the function in that region, but it is not very useful.

This is why we don't usually apply the matrix in an inverse Spline Builder.

On the spline side, spline approximations are usually used to approximate functions. This is useful for (amongst other things):
- interpolation to determine values of functions at points where they are not already known
- creating smooth functions which minimise certain properties (e.g. splines are often the solution to energy minimisation problems)

For interpolation we use an "interpolating spline" which "interpolates" a function f(x) at known "interpolation points":
image

The interpolating spline (in red) is used to approximate the function at arbitrary points x. It is possible to do this with a Fourier approximation by evaluating the equation above, however while a spline approximation will tend to be smoother than the original function, the Fourier approximation will tend to be noisier:
image

This is why we don't usually use a Fourier approximation to interpolate a function.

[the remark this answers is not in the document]
EmilyBourne: I guess I reached the limit for the maximum number of equations in a comment :laughing:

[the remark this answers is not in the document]
EmilyBourne: > I think this is already complicated enough and we should not use two different names for the same thing in the splines.

I think it is significantly more complicated to use the same name for two similar but slightly different concepts than to use 2 names. Having different names makes it easier to explain the difference clearly.

The mathematical convention is:
image

You could argue that every point of every function is an interpolation point as every function interpolates itself but I think this is just proof that the concept of "interpolation point" is not relevant for every function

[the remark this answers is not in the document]
blegouix: #435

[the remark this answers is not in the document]
EmilyBourne: Yes I would be happy to use the term `interpolation points` for an FFT. But if you are aiming for a common API I would recommend using terminology from the splines as it is used in more contexts. The Fourier terminology tends to be quite specific to Fourier. But if for example you added a Lagrange interpolator then the builder/evaluator would contain the same concepts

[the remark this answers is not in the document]
blegouix: Thx sorry you were right from the beginning, As an evaluation can contain less or more information (depending on the "arrival" mesh) than the spline coefficients this is not an (inverse) transformation. This is the case in semi-Lagrangian though but this is not the general case.

Should I "`sed -i s/interpolation/evaluation/g`" the SplineEvaluator and SplineEvaluator2D in another branch ?

include/ddc/kernels/splines/spline_evaluator.hpp
blegouix: I precised it with " spline function defined on a cartesian product of batch_domain and B-splines"

include/ddc/kernels/splines/spline_evaluator.hpp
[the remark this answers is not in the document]
blegouix: Is:
`@tparam MemorySpace The Kokkos memory space on which the data (spline coefficients and evaluation) is stored.`
ok ?

include/ddc/kernels/splines/spline_evaluator.hpp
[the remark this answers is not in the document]
blegouix: This can be closed now that interpolation is renamed evaluation ?

include/ddc/kernels/splines/spline_evaluator.hpp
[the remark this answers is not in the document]
EmilyBourne: Yes that's much clearer. Thanks :)

include/ddc/kernels/splines/spline_evaluator.hpp
```diff
@@ -113,26 +157,79 @@ class SplineEvaluator
...
+     * @param x A reference to another SplineEvaluator
+     * @return A reference to the copied SplineEvaluator
+     */
     SplineEvaluator& operator=(SplineEvaluator const& x) = default;
 
+    /**
+     * @brief Move-assigns
+     *
+     * @param x An rvalue to another SplineEvaluator
+     * @return A reference to the moved SplineEvaluator
```
blegouix, line 189:
```suggestion
     * @return A reference to this object.
```

include/ddc/kernels/splines/spline_evaluator.hpp
```diff
@@ -113,26 +157,79 @@ class SplineEvaluator
     {
     }
 
+    /**
+     * @brief Copy-constructs
+     *
+     * @param x A reference to another SplineEvaluator
```
blegouix, line 163:
```suggestion
     * @param x A reference to another SplineEvaluator.
```

include/ddc/kernels/splines/spline_evaluator.hpp
```diff
@@ -113,26 +157,79 @@ class SplineEvaluator
     {
     }
 
+    /**
+     * @brief Copy-constructs
```
blegouix, line 161:
```suggestion
     * @brief Copy-constructs.
```

include/ddc/kernels/splines/spline_evaluator.hpp
```diff
@@ -113,26 +157,79 @@ class SplineEvaluator
...
+     *
+     * @param x A reference to another SplineEvaluator
+     */
     SplineEvaluator(SplineEvaluator const& x) = default;
 
+    /**
+     * @brief Move-constructs
+     *
+     * @param x An rvalue to another SplineEvaluator
```
blegouix, line 170:
```suggestion
     * @brief Move-constructs.
     *
     * @param x An rvalue to another SplineEvaluator.
```

include/ddc/kernels/splines/spline_evaluator.hpp
```diff
@@ -113,26 +157,79 @@ class SplineEvaluator
...
     SplineEvaluator(SplineEvaluator&& x) = default;
 
+    /// @brief Destructs
     ~SplineEvaluator() = default;
 
+    /**
+     * @brief Copy-assigns
+     *
+     * @param x A reference to another SplineEvaluator
+     * @return A reference to the copied SplineEvaluator
```
blegouix, line 181:
```suggestion
     * @brief Copy-assigns.
     *
     * @param x A reference to another SplineEvaluator.
     * @return A reference to the copied SplineEvaluator.
```

include/ddc/kernels/splines/spline_evaluator.hpp
```diff
@@ -113,26 +157,79 @@ class SplineEvaluator
...
+     */
     SplineEvaluator& operator=(SplineEvaluator const& x) = default;
 
+    /**
+     * @brief Move-assigns
+     *
+     * @param x An rvalue to another SplineEvaluator
```
blegouix, line 188:
```suggestion
     * @brief Move-assigns.
     *
     * @param x An rvalue to another SplineEvaluator.
```

include/ddc/kernels/splines/spline_evaluator_2d.hpp
```diff
@@ -193,76 +235,105 @@ class SplineEvaluator2D
     }
 
     /**
-     * @brief Instantiate a SplineEvaluator2D from another
-     * SplineEvaluator2D (lvalue).
+     * @brief Copy-constructs
```
blegouix, line 238:
```suggestion
     * @brief Copy-constructs.
```

include/ddc/kernels/splines/spline_evaluator_2d.hpp
```diff
@@ -193,76 +235,105 @@ class SplineEvaluator2D
...
      *
-     * @param[in] x
-     * 		SplineEvaluator2D evaluator used to instantiate the new one.
+     * @param x A reference to another SplineEvaluator
```
blegouix, line 240:
```suggestion
     * @param x A reference to another SplineEvaluator.
```

include/ddc/kernels/splines/spline_evaluator.hpp
```diff
@@ -179,6 +309,26 @@ class SplineEvaluator
...
+     * (basis splines). They can be obtained via various methods, such as using a SplineBuilder.
+     *
+     * The derivation is not performed in a multidimensional way (in any sense). This is a batched 1D derivation.
+     * This means that for each slice of coordinates identified by a batch_domain_type::discrete_element_type,
+     * the derivation is performed with the 1D set of spline coefficients identified by the same batch_domain_type::discrete_element_type.
+     *
+     * @param[out] spline_eval The derivatives of the spline function at the desired coordinates. For practical reasons those are
+     * stored in a ChunkSpan defined on a batched_evaluation_domain_type. Note that the coordinates of the
+     * points represented by this domain are unused and irrelevant (but the points themselves (DiscreteElement) are used to select
+     * the set of 1D spline coefficients retained to perform the evaluation).
```
tpadioleau, line 325:
```suggestion
     * @param[out] spline_eval The derivatives of the spline function at the desired coordinates. For practical reasons those are
     * stored in a ChunkSpan defined on a batched_evaluation_domain_type.
```

include/ddc/kernels/splines/spline_evaluator.hpp
```diff
@@ -142,6 +239,28 @@ class SplineEvaluator
...
+     * This is not a multidimensional evaluation. This is a batched 1D evaluation. This means that for each slice of coordinates
+     * identified by a batch_domain_type::discrete_element_type, the evaluation is performed with the 1D set of
+     * spline coefficients identified by the same batch_domain_type::discrete_element_type.
+     *
+     * Remark: calling SplineBuilder then SplineEvaluator corresponds to a spline interpolation.
+     *
+     * @param[out] spline_eval The values of the spline function at the desired coordinates. For practical reasons those are
+     * stored in a ChunkSpan defined on a batched_evaluation_domain_type. Note that the coordinates of the
+     * points represented by this domain are unused and irrelevant (but the points themselves (DiscreteElement) are used to select
+     * the set of 1D spline coefficients retained to perform the evaluation).
```
tpadioleau, line 257:
```suggestion
     * @param[out] spline_eval The values of the spline function at the desired coordinates. For practical reasons those are
     * stored in a ChunkSpan defined on a batched_evaluation_domain_type.
```
I would remove the comment on `spline_eval`, I feel it is redundant with the comment in the description of `coords_eval`.

include/ddc/kernels/splines/spline_evaluator.hpp
```diff
@@ -208,6 +358,19 @@ class SplineEvaluator
                 });
     }
 
+    /** @brief Perform batched 1D integrations of a spline function (described by its spline coefficients) along the dimension of interest and store results on a subdomain of batch_domain.
+     *
+     * The spline coefficients represent a spline function defined on a B-splines (basis splines). They can be obtained via various methods, such as using a SplineBuilder.
```
tpadioleau, line 363:
```suggestion
     * The spline coefficients represent a spline function defined on a B-splines (basis splines). They can be obtained via the SplineBuilder.
```
I would either give examples of the various methods either not mention it

include/ddc/kernels/splines/spline_evaluator_2d.hpp
```diff
@@ -193,76 +235,105 @@ class SplineEvaluator2D
...
     /**
-     * @brief Assign a SplineEvaluator2D from another SplineEvaluator2D (lvalue).
-     *
-     * @param[in] x
-     * 		SplineEvaluator2D mapping used to assign.
+     * @brief Copy-assigns.
      *
-     * @return The SplineEvaluator2D assigned.
+     * @param x A reference to another SplineEvaluator.
+     * @return A reference to the copied SplineEvaluator.
```
tpadioleau, line 258:
```suggestion
     * @return A reference to this object.
```

include/ddc/kernels/splines/spline_evaluator_2d.hpp
```diff
@@ -16,15 +16,29 @@
...
+ * @tparam ExecSpace The Kokkos execution space on which the spline evaluation is performed.
+ * @tparam MemorySpace The Kokkos memory space on which the data (spline coefficients and evaluation) is stored.
+ * @tparam BSplines1 The discrete dimension representing the B-splines along the first dimension of interest.
+ * @tparam BSplines2 The discrete dimension representing the B-splines along the second dimension of interest.
+ * @tparam EvaluationMesh1 The first discrete dimension on which evaluation points are defined.
+ * @tparam EvaluationMesh2 The second discrete dimension on which evaluation points are defined.
+ * @tparam LeftExtrapolationRule1 The lower extrapolation rule type along first dimension of interest.
+ * @tparam RightExtrapolationRule1 The upper extrapolation rule type along first dimension of interest.
+ * @tparam LeftExtrapolationRule2 The lower extrapolation rule type along second dimension of interest.
+ * @tparam RightExtrapolationRule2 The upper extrapolation rule type along second dimension of interest.
```
tpadioleau, line 32: same comment for left/right

include/ddc/kernels/splines/spline_evaluator_2d.hpp
```diff
@@ -273,6 +344,28 @@ class SplineEvaluator2D
...
+     * This is not a nD evaluation. This is a batched 2D evaluation. This means that for each slice of coordinates
+     * identified by a batch_domain_type::discrete_element_type, the evaluation is performed with the 2D set of
+     * spline coefficients identified by the same batch_domain_type::discrete_element_type.
+     *
+     * Remark: calling SplineBuilder2D then SplineEvaluator2D corresponds to a 2D spline interpolation.
+     *
+     * @param[out] spline_eval The values of the 2D spline function at the desired coordinates. For practical reasons those are
+     * stored in a ChunkSpan defined on a batched_evaluation_domain_type. Note that the coordinates of the
+     * points represented by this domain are unused and irrelevant (but the points themselves (DiscreteElement) are used to select
+     * the set of 2D spline coefficients retained to perform the evaluation).
```
tpadioleau, line 362:
```suggestion
     * @param[out] spline_eval The values of the 2D spline function at the desired coordinates. For practical reasons those are
     * stored in a ChunkSpan defined on a batched_evaluation_domain_type.
```

include/ddc/kernels/splines/spline_evaluator_2d.hpp
```diff
@@ -443,14 +578,24 @@ class SplineEvaluator2D
...
+     * (basis splines). They can be obtained via various methods, such as using a SplineBuilder2D.
+     *
+     * This is not a nD differentiation. This is a batched 2D differentiation.
+     * This means that for each slice of coordinates identified by a batch_domain_type::discrete_element_type,
+     * the differentiation is performed with the 2D set of spline coefficients identified by the same batch_domain_type::discrete_element_type.
+     *
+     * @param[out] spline_eval The derivatives of the 2D spline function at the desired coordinates. For practical reasons those are
+     * stored in a ChunkSpan defined on a batched_evaluation_domain_type. Note that the coordinates of the
+     * points represented by this domain are unused and irrelevant (but the points themselves (DiscreteElement) are used to select
+     * the set of 2D spline coefficients retained to perform the evaluation).
```
tpadioleau, line 593:
```suggestion
     * @param[out] spline_eval The derivatives of the 2D spline function at the desired coordinates. For practical reasons those are
     * stored in a ChunkSpan defined on a batched_evaluation_domain_type.
```

include/ddc/kernels/splines/spline_evaluator_2d.hpp
```diff
@@ -485,14 +630,24 @@ class SplineEvaluator2D
...
+     * (basis splines). They can be obtained via various methods, such as using a SplineBuilder2D.
+     *
+     * This is not a nD cross-differentiation. This is a batched 2D cross-differentiation.
+     * This means that for each slice of coordinates identified by a batch_domain_type::discrete_element_type,
+     * the cross-differentiation is performed with the 2D set of spline coefficients identified by the same batch_domain_type::discrete_element_type.
+     *
+     * @param[out] spline_eval The cross-derivatives of the 2D spline function at the desired coordinates. For practical reasons those are
+     * stored in a ChunkSpan defined on a batched_evaluation_domain_type. Note that the coordinates of the
+     * points represented by this domain are unused and irrelevant (but the points themselves (DiscreteElement) are used to select
+     * the set of 2D spline coefficients retained to perform the evaluation).
```
tpadioleau, line 645:
```suggestion
     * @param[out] spline_eval The cross-derivatives of the 2D spline function at the desired coordinates. For practical reasons those are
     * stored in a ChunkSpan defined on a batched_evaluation_domain_type.
```

include/ddc/kernels/splines/spline_evaluator_2d.hpp
```diff
@@ -526,6 +681,27 @@ class SplineEvaluator2D
...
+     *
+     * This is not a nD evaluation. This is a batched 2D differentiation.
+     * This means that for each slice of coordinates identified by a batch_domain_type::discrete_element_type,
+     * the differentiation is performed with the 2D set of spline coefficients identified by the same batch_domain_type::discrete_element_type.
+     *
+     * @tparam InterestDim Dimension along which differentiation is performed.
+     * @param[out] spline_eval The derivatives of the 2D spline function at the desired coordinates. For practical reasons those are
+     * stored in a ChunkSpan defined on a batched_evaluation_domain_type. Note that the coordinates of the
+     * points represented by this domain are unused and irrelevant (but the points themselves (DiscreteElement) are used to select
+     * the set of 2D spline coefficients retained to perform the evaluation).
```
tpadioleau, line 698:
```suggestion
     * @param[out] spline_eval The derivatives of the 2D spline function at the desired coordinates. For practical reasons those are
     * stored in a ChunkSpan defined on a batched_evaluation_domain_type.
```

include/ddc/kernels/splines/spline_evaluator_2d.hpp
```diff
@@ -554,6 +730,31 @@ class SplineEvaluator2D
...
+     *
+     * Note: double-differentiation other than cross-differentiation is not supported atm. See #440
+     *
+     * @tparam InterestDim1 First dimension along which differentiation is performed.
+     * @tparam InterestDim2 Second dimension along which differentiation is performed.
+     *
+     * @param[out] spline_eval The derivatives of the 2D spline function at the desired coordinates. For practical reasons those are
+     * stored in a ChunkSpan defined on a batched_evaluation_domain_type. Note that the coordinates of the
+     * points represented by this domain are unused and irrelevant (but the points themselves (DiscreteElement) are used to select
+     * the set of 2D spline coefficients retained to perform the evaluation).
```
tpadioleau, line 751:
```suggestion
     * @param[out] spline_eval The derivatives of the 2D spline function at the desired coordinates. For practical reasons those are
     * stored in a ChunkSpan defined on a batched_evaluation_domain_type.
```

tpadioleau (commented):
Great job only minor modifications

include/ddc/kernels/splines/spline_evaluator.hpp
```diff
@@ -15,57 +15,93 @@
...
+ * @brief A class to evaluate, differentiate or integrate a spline function.
+ *
+ * A class which contains an operator () which can be used to evaluate, differentiate or integrate a spline function.
+ *
+ * @tparam ExecSpace The Kokkos execution space on which the spline evaluation is performed.
+ * @tparam MemorySpace The Kokkos memory space on which the data (spline coefficients and evaluation) is stored.
+ * @tparam BSplines The discrete dimension representing the B-splines.
+ * @tparam EvaluationMesh The discrete dimension on which evaluation points are defined.
+ * @tparam LeftExtrapolationRule The lower extrapolation rule type.
+ * @tparam RightExtrapolationRule The upper extrapolation rule type.
```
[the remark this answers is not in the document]
blegouix: #503

PR #431, opened 2024-04-29
Splines small classes doc

blegouix:


include/ddc/kernels/splines/spline_boundary_conditions.hpp
```diff
@@ -8,16 +8,22 @@
...
-    // Periodic boundary condition u(1)=u(n)
-    PERIODIC,
-    // Hermite boundary condition
-    HERMITE,
-    // Use Greville points instead of conditions on derivative for B-Spline
-    // interpolation
-    GREVILLE,
+    PERIODIC, ///< Periodic boundary condition u(1)=u(n)
+    HERMITE, ///< Hermite boundary condition
+    GREVILLE, ///< Use Greville points instead of conditions on derivative for B-Spline interpolation
```
tpadioleau, line 16: I think for these two boundary conditions we will need at least a reference to explain what they are ?

include/ddc/kernels/splines/spline_boundary_conditions.hpp
```diff
@@ -32,6 +38,14 @@ static inline std::ostream& operator<<(std::ostream& out, ddc::BoundCond const b
...
 
+/**
+ * @brief Return the number of equations needed to describe a given boundary condition.
+ *
+ * @param bc The boundary condition.
+ * @param degree The degree of the spline.
+ *
+ * @return The number of equations.
+ **/
 constexpr int n_boundary_equations(ddc::BoundCond const bc, std::size_t const degree)
```
tpadioleau, line 49: I believe `constexpr` and exceptions are not compatible.

tpadioleau (approved):
Good to me

include/ddc/kernels/splines/spline_boundary_conditions.hpp
blegouix: In #437 there is a link to @EmilyBourne thesis. Maybe we can write it there too ? Or Emily has another proposal ?

tpadioleau: Yes seems good to me

tpadioleau (approved):

PR #437, opened 2024-05-03
Transformation general doc

### LINKED ISSUE #435: Document ddc changes of basis from a general pov
See https://github.com/CExA-project/ddc/pull/430#discussion_r1587601447

I think it should be included in the documentation (not in the API part but in the general presentation of DDC features)

blegouix:
Closes #435


docs/transformations.md
```diff
@@ -0,0 +1,91 @@
...
+Those methods are performance portable (supported with CPU Serial, CPU OpenMP, CUDA and HIP backends). They are used
+in *examples/heat_equation_spectral.cpp* (for DFT) and *examples/characteristics_advection.cpp* (for splines).
+A particular use-case of splines is benchmarked in *benchmarks/splines.cpp*.
+
+## General presentation of transformations
+
+### Computation kernels dedicated to transformations in DDC
+
+A transformation is a change of basis, going from a function space to another. Because the available memory is finite,
+the basis are necessarily finite (which means it contains a finite set of basis functions, but those basis functions
```
EmilyBourne, line 21: The plural of "basis" is bases
```suggestion
the bases are necessarily finite (which means it contains a finite set of basis functions, but those basis functions
```

docs/transformations.md
```diff
@@ -0,0 +1,91 @@
...
+may be continuous).
+
+The use-cases of transformations for simulation or signal processing needs are plethoric: interpolation, spectral methods,
```
EmilyBourne, line 24: image
"plethoric" is correct but archaic. I wasn't sure if it was actually a word. I suggest "multitude" or the more straightforward:
```suggestion
There are many use-cases of transformations for simulation or signal processing needs: interpolation, spectral methods,
```

docs/transformations.md
```diff
@@ -0,0 +1,91 @@
...
+finite element methods, filtering (compression), signal analysis (post-process)...
+
+Every basis have their own specificities (ie. orthogonality, n-derivability, etc...) but as they belong to the same frame
```
EmilyBourne, line 27:
```suggestion
Every basis has its own specificities (ie. orthogonality, n-derivability, etc...) but as they belong to the same frame
```

docs/transformations.md
```diff
@@ -0,0 +1,91 @@
...
+of set of functions which form a basis, they have in common a formalism and a terminology. However, this is currently
```
EmilyBourne, line 28:
```suggestion
or set of functions which form a basis, they have in common a formalism and a terminology. However, this is currently
```

docs/transformations.md
```diff
@@ -0,0 +1,91 @@
...
+It must be explained:
+
+- Fourier requires the periodicity of the represented function, thus boundary conditions does not need to be provided.
+- Fourier basis functions are indexed with the wave vector k, whose possible values form a set of coordinates which generates
+the Fourier space. Mesh of "real" space and mesh of Fourier space are in bijection one-to-the-other.
+The situation is more complicated for Splines.
+- Fourier and Splines basis use-cases are quite different. Thus - as shown below - the so-called "evaluator" is not
+implemented for Fourier (while inverse DFT is) and inverse Spline transform is not implemented (while Spline evaluation is).
+- FFT backends support 1D to 3D Fourier transforms. General n-dimensional Spline transform would be a mess to implement for all supported boundary conditions so we limit to the 2D case.
+- DFT is not currently supported in batched configuration (but it may be the subject of futur development). Batched splines are supported.
```
EmilyBourne, line 39:
```suggestion
- DFT is not currently supported in batched configuration (but it may be the subject of future development). Batched splines are supported.
```

docs/transformations.md
```diff
@@ -0,0 +1,91 @@
...
+\f[
+f(x_i) = \sum_j c_j \phi_j(x_i)
+\f]
+
+Note: the first method is covered by the second one in the particular case of \f$ \phi \f$ being the Dirac basis functions.
+
+The aim of a computation kernel dedicated to a transformation is to determine the values of coefficients \f$ c_j \f$
+knowing \f$ f(x_i) \f$ (and for the inverse transformation, \f$ f(x_i) \f$ knowing \f$ c_j \f$). 
+
+Additionally to the *transformation* and the *inverse transformation*, there is a third operation which can be usefull:
```
EmilyBourne, line 67:
```suggestion
Additionally to the *transformation* and the *inverse transformation*, there is a third operation which can be useful:
```

docs/transformations.md
```diff
@@ -0,0 +1,91 @@
...
+the *evaluation*. Indeed, more than a discrete function, coefficients \f$ c_j \f$ generate an interpolating (continuous)
+function which can be evaluated anywhere between the points of the mesh.
+
+Note: performing a (direct) transformation then an evaluation is an interpolation.
+
+Basis of function spaces can have additional properties like orthogonality
```
EmilyBourne, line 73:
```suggestion
Bases of function spaces can have additional properties like orthogonality
```

docs/transformations.md
blegouix: but as they belong to the same frame of "set of functions which form a basis"

En français on a "base de fonctions" mais en anglais je trouve pas d'équivalent plus compact

EmilyBourne: Ah ok. In that case I think you can remove "frame of" it doesn't mean anything here.

"base de fonctions" = function space basis ?

docs/transformations.md
```diff
@@ -0,0 +1,106 @@
...
+### Lexical 
+
+Here is a summary of different general terms used in the context of DDC transformations:
+
+| Term     | Meaning & commentary |
+|----------|-----------------|
+| Discrete function | Function defined on a mesh (via its values at each point of via its coefficients in a particular finite basis representation). |
+| Dimension(s) of interest | Dimension(s) along which the transformation is performed |
+| Batch dimension(s) | Dimension(s) unaffected by the transformation. It leads to a set of transformations performed independantly (along dimensions of interest). Those are embarassingly parallelizable. |
+| Interpolant functions | Same as basis functions \f$ \phi_j \f$ |
```
EmilyBourne, line 87: Are you sure about this definition?

I found https://mathworld.wolfram.com/Interpolant.html
image

I interpret this as meaning "interpolating function" = "interpolant function". I don't think it is the same as a basis function. The basis function does not have to agree with the function being interpolated

docs/transformations.md
blegouix: Probably yes, I find the "space" here a bit confusing because in my understanding ie. Function space of Fourier transform (abstract high-dimensional space generated from all e^ikx) is not the same as Fourier space (same dimensionality as real space whose points are identified with wavevector). That's why i'm trying to avoid using the word "space". What do you think ?

docs/transformations.md
blegouix: You I right it looks I completly invented this

docs/transformations.md
EmilyBourne: Function space is a term from linear algebra (https://en.wikipedia.org/wiki/Vector_space#Function_spaces). It is a kind of vector space.

> A vector space (also called a linear space) is a set whose elements, often called vectors, may be added together and multiplied ("scaled") by numbers called scalars.

Usually when we think of vector spaces we think of a finite-dimensional vector space. I.e. a vector space with a finite number of basis elements. However it is also possible to have an infinite-dimensional vector space.

The function space of Fourier transform is a finite-dimensional function space whose basis elements are
image

while the Fourier space is an infinite-dimensional function space with an infinite number of basis elements
image $n \in [-\infty, ..., \infty]$

EmilyBourne: A Discrete Fourier transform takes you from the space of real functions to a subset of the Fourier space so I don't think this is necessarily misleading for those who are familiar with the theory of Fourier transforms and DFT.

Also beware, in IT most of the time when we talk about "Fourier transforms" what we mean is "Discrete Fourier Transform". Mathematically speaking, a Fourier transform is calculated via an integral and the result is indeed something in Fourier space. However when we calculate this on a computer we necessarily have a "Discrete Fourier Transform" and therefore map to a subset of this space

EmilyBourne: >  ie. Function space of Fourier transform (abstract high-dimensional space generated from all e^ikx) is not the same as Fourier space (same dimensionality as real space whose points are identified with wavevector)

If by "high-dimensional space" you mean "infinitely high-dimensional space" then I don't think I agree with this statement

EmilyBourne: Otherwise maybe this is clearer:

> Every basis has its own specificities (ie. orthogonality, n-derivability, etc...) but as they are all basis functions,
> they have in common a formalism and a terminology. However, this is currently

blegouix: By "high-dimensional space" I mean space with as much dimensions as their are basis functions (N in your formula). Infinite-dimensional only in the continuous case then.

I think I do not agree with this:
"
the Fourier space is an infinite-dimensional function space with an infinite number of basis elements
image 
"

To me the Fourier space is just the space in which $\tilde{f}(\vec(k))$ lies:

"Fourier space (or frequency domain or Fourier domain) is an abstract "space" into which the [Fourier transform] maps a function, consisting of the [amplitude] and phase of the sine function at various [frequencies] that sum to produce the same shape" http://astro.vaporia.com/start/fourierspace.html

So Fourier space maps the Function space of Fourier transform, but they are not the same.

EmilyBourne: I see what you mean `k` is not really a function it is just an object in Fourier space. So Fourier space is a Vector space not a Function space.

The basis doesn't belong to Fourier space it is in a Function space. I agree that these concepts are hard to keep straight but I'm not sure it is the very generic word "space" that causes these problems. In maths "space" just means that you can do `a*b+c` for elements `b` and `c` which belong to the space. This is true for Function space bases and for elements of Fourier space.

What term are you using for the basis elements  image ?
I think that it is this that is hard to define as it is unrelated to the Fourier transform

EmilyBourne: I've just realised that you mean "framework"

EmilyBourne: bases plural

docs/transformations.md
```diff
@@ -0,0 +1,105 @@
...
+
+A transformation is a change of basis, going from a function space to another. Because the available memory is finite,
+the bases are necessarily finite (which means it contains a finite set of basis functions, but those basis functions
+may be continuous).
+
+There are many use-cases of transformations for simulation or signal processing needs: interpolation, spectral methods,
+finite element methods, filtering (compression), signal analysis (post-process)...
+
+Every basis has its own specificities (ie. orthogonality, n-derivability, etc...) but as they all are
+function space basis, they have in common a formalism and a terminology. However, this is currently
```
EmilyBourne, line 28:
```suggestion
function space bases, they have in common a formalism and a terminology. However, this is currently
```

docs/transformations.md
```diff
@@ -0,0 +1,105 @@
...
+not very manifest when looking at the API in the DDC implementations (DFT and Splines API are very different).
+It must be explained:
+
+- Fourier requires the periodicity of the represented function, thus boundary conditions does not need to be provided.
```
EmilyBourne, line 32:
```suggestion
- Fourier requires the periodicity of the represented function, thus boundary conditions do not need to be provided.
```

docs/transformations.md
```diff
@@ -0,0 +1,105 @@
...
+- Fourier basis functions are indexed with the wave vector \f$ \vec{k} \f$, whose possible values form a set of coordinates which generates
+the Fourier space. Mesh of "real" space and mesh of Fourier space are in bijection one-to-the-other.
```
EmilyBourne, line 34:
```suggestion
- Fourier basis functions are indexed with the wave vector \f$ \vec{k} \f$, whose possible values form a set of coordinates which generate
the Fourier space. There is a bijection between the mesh of the "real" space and the mesh of the Fourier space.
```

PR #452, opened 2024-05-21
Splines minor classes doc

blegouix:


include/ddc/kernels/splines/knots_as_interpolation_points.hpp
```diff
@@ -24,6 +24,10 @@ namespace ddc {
  * In the case of strongly non-uniform splines this choice may result in a less
  * well conditioned problem, however most mathematical stability results are proven
  * with this choice of interpolation points.
+ *
+ * @tparam BSplines The type of the uniform or non-uniform spline basis whose knots are used as interpolation points.
+ * @tparam BcXmin The lower boundary condition.
+ * @tparam BcXmin The upper boundary condition.
```
tpadioleau, line 30: Actually same comment as for the spline evaluator. I think we should try to get uniform vocabulary

tpadioleau (approved):

include/ddc/kernels/splines/knots_as_interpolation_points.hpp
blegouix: #503

PR #471, opened 2024-05-29
Splines linear problem PDS band

blegouix:


include/ddc/kernels/splines/splines_linear_problem_pds_band.hpp
```diff
@@ -0,0 +1,172 @@
...
+        }
+        for (int k = 1; k < m_q.extent(0); ++k) {
+            if (i_tmp + k == j_tmp) {
+                m_q(k, i_tmp) = aij;
+                return;
+            }
+        }
+        assert(std::fabs(aij) < 1e-20);
+        return;
+    }
```
tpadioleau, line 115: Would something like this work ? Same question for `get_element`

```suggestion
    virtual void set_element(std::size_t i, std::size_t j, double const aij) override
    {
        assert(i < size());
        assert(j < size());

        // Indices are swapped for an element on subdiagonal
        if (i > j) {
            std::swap(i, j);
        }

        if (j - i < m_q.extent(0)) {
            m_q(j - i, i) = aij;
        } else {
            assert(std::fabs(aij) < 1e-20);
        }
    }
```

tests/splines/matrix.cpp
```diff
@@ -179,6 +179,28 @@ TEST(Matrix, Band)
...
+    // Build a positive-definite symmetric full-rank band matrix
+    for (std::size_t i(0); i < N; ++i) {
+        matrix->set_element(i, i, 2.0 * k);
+        for (std::size_t j(std::max(0, int(i) - int(k))); j < i; ++j) {
+            matrix->set_element(i, j, -1.0);
+        }
+        for (std::size_t j(i + 1); j < std::min(N, i + k + 1); ++j) {
+            matrix->set_element(i, j, -1.0);
+        }
+    }
```
tpadioleau, line 199: I am not sure it is actually PDS. I think the matrix we had in mind was supposed to be symmetric and diagonally dominant thus PDS. But it does not seem to be diagonally dominant. Can you check ?

tpadioleau (changes requested):

tests/splines/matrix.cpp
blegouix: This matrix comes from SLL, i will check though
https://gitlab.maisondelasimulation.fr/gysela-developpers/gyselalibxx/-/blob/main/vendor/sll/tests/matrix.cpp?ref_type=heads#L79

blegouix:
```
import numpy as np;

N=10;
k=3;

A = np.zeros((10,10));
for i in range(0, N):
  A[i,i] =  2. * k;
  for j in range(max(0,i-k),i):
    A[i, j] = -1.;
  for j in range(i+1, min(N,i+k+1)):
    A[i, j] = -1.;

print(A);
print(np.all(np.linalg.eigvals(A) > 0));
```
image

tpadioleau: Ok for this (N,k) couple but i would add 1 on the diagonal, this should make it diagonally dominant

include/ddc/kernels/splines/splines_linear_problem_pds_band.hpp
tpadioleau: Why do you need a loop ?

tpadioleau (approved):
Looks good to me

PR #473, opened 2024-05-29
Splines linear problem pds tridiag

blegouix:


tests/splines/matrix.cpp
```diff
@@ -201,6 +201,28 @@ TEST(Matrix, PDSBand)
...
 
+TEST(Matrix, PDSTridiag)
+{
+    std::size_t const N = 10;
+    std::size_t const k = 1;
+    std::unique_ptr<ddc::detail::SplinesLinearProblem<Kokkos::DefaultExecutionSpace>> matrix
+            = std::make_unique<
+                    ddc::detail::SplinesLinearProblemPDSTridiag<Kokkos::DefaultExecutionSpace>>(N);
+
+    // Build a positive-definite symmetric full-rank tridiagonal matrix
```
blegouix, line 212: positive-definiteness has been checked

tpadioleau (approved):
Looks ok to me. A minor concern about the layout of m_q. It has only one valid layout compared to the other classes.

blegouix:
Yes that's true, LayoutRight is mandatory here.

I need reapproval.

tpadioleau (approved):

yasahi-hpc:
I have a question about this PR. What is a typical storage format for this type of matrix. It seems that they have two separate matrices for diag and non-diag in [lapack](https://netlib.org/lapack/explore-html/d7/d77/dpttrs_8f_source.html).

tpadioleau:
Indeed we have made a different choice. I think most libraries are inspired by the original design of LAPACK which uses 3 different pointeurs to store the diagonals in the general tridiagonal solver.

yasahi-hpc:
Sure. I guess I need to employ the `lapack` interface

PR #474, opened 2024-05-29
Splines linear problem 2x2 blocks

blegouix:


include/ddc/kernels/splines/splines_linear_problem_2x2_blocks.hpp
```diff
@@ -0,0 +1,318 @@
...
+     * | lambda | delta - lambda*Q^-1*gamma |
+     */
+    void setup_solver() override
+    {
+        m_top_left_block->setup_solver();
+
+        // Compute Q^-1*gamma
+        auto top_right_block_device = create_mirror_view_and_copy(ExecSpace(), m_top_right_block);
+        m_top_left_block->solve(top_right_block_device);
+        deep_copy(m_top_right_block, top_right_block_device);
```
tpadioleau, line 148: Could you move these lines inside `compute_schur_complement` ?

include/ddc/kernels/splines/splines_linear_problem_2x2_blocks.hpp
```diff
@@ -0,0 +1,318 @@
...
+     *
+     * [SHOULD BE PRIVATE (GPU programming limitation)]
+     *
+     * @param u
+     * @param v
+     */
+    virtual void solve_bottom_left_block_section(MultiRHS const u, MultiRHS v) const
+    {
+        auto bottom_left_block_device
+                = create_mirror_view_and_copy(ExecSpace(), m_bottom_left_block);
```
tpadioleau, line 167: I think we should store these blocks on the device so that to avoid copy them at each call.

include/ddc/kernels/splines/splines_linear_problem_2x2_blocks.hpp
```diff
@@ -0,0 +1,318 @@
...
+        , m_top_right_block(
+                  "top_right_block",
+                  m_top_left_block->size(),
+                  mat_size - m_top_left_block->size())
+        , m_bottom_left_block(
+                  "bottom_left_block",
+                  mat_size - m_top_left_block->size(),
+                  m_top_left_block->size())
+        , m_bottom_right_block(
+                  new SplinesLinearProblemDense<ExecSpace>(mat_size - m_top_left_block->size()))
```
tpadioleau, line 65: I always find it a bit dangerous to use a member to initialize other members. If for some reason one reorders the class members, this will break.

include/ddc/kernels/splines/splines_linear_problem_2x2_blocks.hpp
```diff
@@ -0,0 +1,318 @@
...
+        } else if (i >= nq && j >= nq) {
+            return m_bottom_right_block->get_element(i - nq, j - nq);
+        } else if (j >= nq) {
+            return m_top_right_block(i, j - nq);
+        } else {
+            return m_bottom_left_block(i - nq, j);
+        }
+    }
+
+    virtual void set_element(std::size_t const i, std::size_t const j, double const aij) override
```
tpadioleau, line 90:
```suggestion
    void set_element(std::size_t const i, std::size_t const j, double const aij) override
```

include/ddc/kernels/splines/splines_linear_problem_2x2_blocks.hpp
```diff
@@ -0,0 +1,318 @@
...
+        , m_bottom_right_block(
+                  new SplinesLinearProblemDense<ExecSpace>(mat_size - m_top_left_block->size()))
+    {
+        assert(m_top_left_block->size() <= mat_size);
+
+        Kokkos::deep_copy(m_top_right_block, 0.);
+        Kokkos::deep_copy(m_bottom_left_block, 0.);
+    }
+
+    virtual double get_element(std::size_t const i, std::size_t const j) const override
```
tpadioleau, line 73:
```suggestion
    double get_element(std::size_t const i, std::size_t const j) const override
```

include/ddc/kernels/splines/splines_linear_problem_2x2_blocks.hpp
```diff
@@ -0,0 +1,318 @@
...
+                    const int j = teamMember.league_rank();
+
+
+                    Kokkos::parallel_for(
+                            Kokkos::TeamThreadRange(teamMember, v.extent(0)),
+                            [&](const int i) {
+                                for (int l = 0; l < u.extent(0); ++l) {
+                                    v(i, j) -= bottom_left_block_device(i, l) * u(l, j);
+                                }
+                            });
```
tpadioleau, line 182: I guess it could take the form a parallel_reduce with a 2D TeamThreadMDRange.

tpadioleau (changes requested):
This is looks good. I think we could go further and implement all `solve_*_block_section_*` functions in a single function, they all seem to do a "gemv" operation.

include/ddc/kernels/splines/splines_linear_problem_2x2_blocks.hpp
blegouix: Yes but it would not only compute the schur complement anymore ? It's a detail though

include/ddc/kernels/splines/splines_linear_problem_2x2_blocks.hpp
```diff
@@ -0,0 +1,318 @@
...
+
+    /**
+     * @brief Compute v <- v - lambda*u.
+     *
+     * [SHOULD BE PRIVATE (GPU programming limitation)]
+     *
+     * @param u
+     * @param v
+     */
+    virtual void solve_bottom_left_block_section(MultiRHS const u, MultiRHS v) const
```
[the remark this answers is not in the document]
blegouix: This class is going to inherit from this one:
https://github.com/CExA-project/ddc/blob/3e68db61dea6ac3abb74e3b230c59be8a9ffc7d1/include/ddc/kernels/splines/matrix_periodic_banded.hpp

So I need few virtuals but not all of them, Ill remove.

(`const` usage is definitely the most dismal thing of C++ for me ^^')

include/ddc/kernels/splines/splines_linear_problem_2x2_blocks.hpp
blegouix: ok

include/ddc/kernels/splines/splines_linear_problem_2x2_blocks.hpp
blegouix: Oh yes that's better

include/ddc/kernels/splines/splines_linear_problem_2x2_blocks.hpp
blegouix: I was wondering if we could not pass m_bottom_right_block to the constructor the same way we pass m_top_left_block. This is one additionnal line of code for the user but makes the class more "symmetric".

(in practice m_top_left_block will always be a SplinesLinearSolverDense)

include/ddc/kernels/splines/splines_linear_problem_2x2_blocks.hpp
tpadioleau: Well the documentation of the `compute_schur_complement` says
> // @brief Compute the Schur complement delta - lambda*Q^-1*gamma.

But the computation `Q^-1*gamma` seems to be handled outside of the function ?

include/ddc/kernels/splines/splines_linear_problem_2x2_blocks.hpp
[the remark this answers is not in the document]
blegouix: Why do we const here but not:
`void solve(MultiRHS b, bool const transpose) const override` ?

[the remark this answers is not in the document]
tpadioleau: `const` always applies to the object, here `v`, restricting the usage of some member functions (those not qualified by `const`). Then the behaviour of the set of `const` functions depends on the meaning the developer gave to the class. View containers like `std::span`, `std::mdspan`, `Kokkos::View` or raw pointers, do not own the data. If they represent non-const data, const qualifying them still allow you to modify the data. If you don't want to be able to modify the data, you need to tell they represent const data, for example `Kokkos::View<const double*>`.

include/ddc/kernels/splines/splines_linear_problem_2x2_blocks.hpp
blegouix: Yes because `Q^-1*gamma` is one of the blocks of the matrix whereas Schur complement is another block

include/ddc/kernels/splines/splines_linear_problem_2x2_blocks.hpp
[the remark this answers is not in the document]
blegouix: Ok so we should have `solve(MultiRHS const b, bool const transpose)` ?

include/ddc/kernels/splines/splines_linear_problem_2x2_blocks.hpp
blegouix: I do the change but I probably need to revert it in the MR on SplinesLinearProblemPeriodicBand

include/ddc/kernels/splines/splines_linear_problem_2x2_blocks.hpp
blegouix: Actually a parallel_reduce on 2D array cannot compute `sum_l bottom_left_block_device(i, l) * u(l, j)`, it would compute `sum_{i,l} bottom_left_block_device(i, l) * u(l, j);`.

I think this discussion is a bit complicated so I switch on slack.

include/ddc/kernels/splines/splines_linear_problem_2x2_blocks.hpp
tpadioleau: Is the size of the matrix worth it ?

include/ddc/kernels/splines/splines_linear_problem_2x2_blocks.hpp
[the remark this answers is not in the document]
tpadioleau: Yes we can, it has no influence on the API

include/ddc/kernels/splines/splines_linear_problem_2x2_blocks.hpp
tpadioleau: Ok I think i get it. So `compute_schur_complement` assumes that the top right block already contains `Q^-1 gamma`

include/ddc/kernels/splines/splines_linear_problem_2x2_blocks.hpp
blegouix: I guess not for big simulations where splines degree << mesh size but I do not really see downside of doing this. As you prefer.

include/ddc/kernels/splines/splines_linear_problem_2x2_blocks.hpp
blegouix: Yes exactly

include/ddc/kernels/splines/splines_linear_problem_2x2_blocks.hpp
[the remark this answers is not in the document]
blegouix: https://github.com/CExA-project/ddc/issues/475

include/ddc/kernels/splines/splines_linear_problem_2x2_blocks.hpp
```diff
@@ -0,0 +1,260 @@
...
+                                if (!transpose) {
+                                    y_tmp += LinOp(i, l) * x(l, j);
+                                } else {
+                                    y_tmp += LinOp(l, i) * x(l, j);
+                                }
+                            },
+                            LinOpTimesX);
+                    if (teamMember.team_rank() == 0) {
+                        y(i, j) -= LinOpTimesX;
+                    }
```
blegouix, line 212: I tried to avoid this by writing ie. `y_tmp -= LinOp(i, l) * x(l, j);` but it does not work.

include/ddc/kernels/splines/splines_linear_problem_2x2_blocks.hpp
blegouix: actually this is a bad idea because we compute inplace the Schur complement in delta.

blegouix: About your issue, the problem is top_left_block is moved to m_top_left_block, thus top_left_move->size() segfaults if called after the move.

include/ddc/kernels/splines/splines_linear_problem_2x2_blocks.hpp
```diff
@@ -0,0 +1,260 @@
...
+
+        // Push lambda on device in bottom-left block
+        m_bottom_left_block.modify_host();
+        m_bottom_left_block.sync_device();
+
+        // Compute delta - lambda*Q^-1*gamma in bottom-right block
+        compute_schur_complement();
+
+        // Setup the bottom-right solver
+        m_bottom_right_block->setup_solver();
```
blegouix, line 165:
```suggestion
        // Compute delta - lambda*Q^-1*gamma in bottom-right block & setup the bottom-right solver
        compute_schur_complement();
        m_bottom_right_block->setup_solver();
```
```suggestion
        // Compute delta - lambda*Q^-1*gamma in bottom-right block
        compute_schur_complement();

        // Setup the bottom-right solver
        m_bottom_right_block->setup_solver();
```

include/ddc/kernels/splines/splines_linear_problem_2x2_blocks.hpp
```diff
@@ -0,0 +1,258 @@
...
+
+                    double LinOpTimesX = 0.;
+                    Kokkos::parallel_reduce(
+                            Kokkos::TeamThreadRange(teamMember, x.extent(0)),
+                            [&](const int l, double& LinOpTimesX_tmp) {
+                                if (!transpose) {
+                                    LinOpTimesX_tmp += LinOp(i, l) * x(l, j);
+                                } else {
+                                    LinOpTimesX_tmp += LinOp(l, i) * x(l, j);
+                                }
```
tpadioleau, line 205: This condition is called at each iteration even though it is a loop invariant. I am thinking we could either:
1. Place it outside the outer loop to mitigate the cost
2. Define an intermediate LinOpT that would have a LayoutLeft and the dimensions reversed ?

include/ddc/kernels/splines/splines_linear_problem_2x2_blocks.hpp
```diff
@@ -0,0 +1,258 @@
...
+#include "splines_linear_problem_dense.hpp"
+
+namespace ddc::detail {
+
+/**
+ * @brief A 2x2-blocks linear problem dedicated to the computation of a spline approximation (taking in account boundary conditions),
+ * with all blocks except top-left one being stored in dense format.
+ *
+ * A = |   Q    | gamma |
+ *     | lambda | delta |
```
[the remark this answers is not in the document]
blegouix: I think this is better for the documentation, otherwise the LU decomposition would be very hard to read with terms like `bottom_right - bottom_left*top_left^-1*top_right` in a matrix.

include/ddc/kernels/splines/splines_linear_problem_2x2_blocks.hpp
```diff
@@ -0,0 +1,258 @@
...
+     */
+    void gemv_minus1_1(
+            MultiRHS const x,
+            MultiRHS const y,
+            Kokkos::View<double**, Kokkos::LayoutRight, typename ExecSpace::memory_space> const
+                    LinOp,
+            bool const transpose = false) const
+    {
+        assert(!transpose && LinOp.extent(0) == y.extent(0)
+               || transpose && LinOp.extent(1) == y.extent(0));
```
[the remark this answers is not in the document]
tpadioleau: Nice trick to remember the priority between && and || but it is not the end of the story. You need to take into account all unary and binary operators https://en.cppreference.com/w/cpp/language/operator_precedence. I clearly don't know this table by heart.

include/ddc/kernels/splines/splines_linear_problem_2x2_blocks.hpp
blegouix: Yeah, I tried:

```
                    Kokkos::parallel_reduce(
                            Kokkos::TeamThreadRange(teamMember, x.extent(0)),
                            [&](const int l, double& y_tmp) {
                                if (!transpose) {
                                    y_tmp -= LinOp(i, l) * x(l, j);
                                } else {
                                    y_tmp -= LinOp(l, i) * x(l, j);
                                }
                            },
                            y(i,j));
 ```
 
 With y being of course not initialized to 0.

include/ddc/kernels/splines/splines_linear_problem_2x2_blocks.hpp
blegouix: So we cannot rely on the compiler to optimize this even if transpose is a const ? I'll go to solution 1. because solution 2. may be annoying if we want to give a try to LayoutLeft matrix blocks.

tpadioleau: Mmmh maybe the GPU compiler could move the condition in between the two loops. I would not bet that it would move it out of the outer loop. It would mean splitting a GPU kernel in two ? Highly doubt it

blegouix: Yeah ok I agree

include/ddc/kernels/splines/splines_linear_problem_2x2_blocks.hpp
tpadioleau: If i remember correctly, `parallel_reduce` does not contribute to the variable result, `y(i, j)` in the above code, it simply overwrites it.

tpadioleau (approved):
Ok for me!

PR #476, opened 2024-06-03
Splines linear problem periodic band

blegouix:


blegouix:
I close this branch because the gain looks too low compared to the complexity of the class (more precisely, the gain of performance on GPU is very unsure due to the specific version of Gemv implemented here and even if it exists it is probably very small).

Unfortunately the current implementation does not work (probably a minor bug to find) but there is a working version of it here (using the previous Matrix API though): https://github.com/CExA-project/ddc/pull/329/files#diff-f26791a050f0e7ca77ce0629789bffb58551ec0190a43c369b95b00408e79ad5

PR #480, opened 2024-06-05
Splines kokkos kernels backend

blegouix:

PR #491, opened 2024-06-11
testing KK backend

yasahi-hpc:
This PR aims at testing the KK implementation for splines. 
Mainly used for benchmarking KK vs gingko versions.


blegouix:
@yasahi-hpc Sorry I completly messed up while trying to reorganize the branches. But now your changes are in #480. So if you want to make tests or benchmark you can derivate a new branch from it.

PR #496, opened 2024-06-18
Splines improve benchmark

blegouix:
Scripts used to produce the figures in my report.

- Capability to sweep on parameters determined at compile-time: Uniform/NonUniform, Degree, ExecutionSpace and in the futur solver Ginkgo/Lapack.


benchmarks/splines.cpp
```diff
@@ -175,49 +195,103 @@ static void characteristics_advection(benchmark::State& state)
...
+    benchs[0] = characteristics_advection_unitary<std::false_type, 3>;
+    benchs[1] = characteristics_advection_unitary<std::false_type, 4>;
+    benchs[2] = characteristics_advection_unitary<std::false_type, 5>;
+    benchs[3] = characteristics_advection_unitary<std::true_type, 3>;
+    benchs[4] = characteristics_advection_unitary<std::true_type, 4>;
+    benchs[5] = characteristics_advection_unitary<std::true_type, 5>;
+
+    // Run the desired bench
+    benchs[state.range(0) * 3 + state.range(1) - 3](state);
+}
```
blegouix, line 222: @tpadioleau any opinion about this ? It is hacky but helps a lot to make the benchmarks flexible. I did not find yet any way to write it as nested loops, maybe there is one still.

blegouix: I can imagine an improvement of TypeSeq with something like:

```
template<std::size_t N, typename List>
struct TypeByIndex;

template<std::typename Head, typename... Tail>
struct TypeByIndex<0, ddc::detail::TypeSeq<Head, Tail...>> {
    using type = Head;
};

template<std::size_t N, typename Head, typename... Tail>
struct TypeByIndex<N, ddc::detail::TypeSeq<Head, Tail...>> {
    using type = typename TypeAt<N-1, TypeList<Tail...>>::type;
};
```

And upon this a parameter pack expansions on `TypeSeq<Kokkos::DefaultHostExecutionSpace, Kokkos::DefaultExecutionSpace>`, `TypeSeq<std::false_type, std::true_type>` and  `TypeSeq<3, 4, 5>` should be feasible, but it may be a bit too complicated.

tpadioleau: > @tpadioleau any opinion about this ? It is hacky but helps a lot to make the benchmarks flexible. I did not find yet any way to write it as "nested loops", maybe there is one still.

Don't you think you need different tests for Ginkgo and Lapack ? They have different parameters ? `cols_per_chunk_ref` and `preconditionner_max_block_size_ref` do not make sense for Lapack right ?

tpadioleau: If you want to select a benchmark based on a runtime parameter could we use a string and store the different benchmarks in a `std::map<std::string, benchmark> my_benchmarks` ? you would just call `my_benchmarks.at("host")`

blegouix: > > @tpadioleau any opinion about this ? It is hacky but helps a lot to make the benchmarks flexible. I did not find yet any way to write it as "nested loops", maybe there is one still.
> 
> Don't you think you need different tests for Ginkgo and Lapack ? They have different parameters ? `cols_per_chunk_ref` and `preconditionner_max_block_size_ref` do not make sense for Lapack right ?

Atm the choice of Backend still has to be manually changed by editing the file. `cols_per_chunk_ref` and `preconditionner_max_block_size_ref` do not make sense for Lapack indeed, and this is weird to have them passed as optional arguments of SplineBuilder. But this is the current implementation, we could do something better with a kwArgs struct (similar to what we have in the FFT) but this requires more developments.

About the benchmark, the current way to use it is to comment/uncomment the benchmarks we want to run so we just have to not run the preconditioner of cols_per_chunk benchmark with Lapack.


> If you want to select a benchmark based on a runtime parameter could we use a string and store the different benchmarks in a `std::map<std::string, benchmark> my_benchmarks` ? you would just call `my_benchmarks.at("host")`

Making a more interfaced benchmark system would be great but requires development. And I think it would not impact the way benchmarks are performed (this would be a software overlay build upon it). Maybe gtest already provides something to do it.

blegouix: (my question was more about instanciating all the possible variations of the class we would like to benchmark, combining explicitly compile-time parameters like ExecutionSpace, Degree and Uniformity)

tpadioleau: Ok I think I get the problem, would this google benchmark feature https://github.com/google/benchmark/blob/main/docs/user_guide.md#templated-benchmarks-that-take-arguments help you ?

blegouix: I don't think so because I think it does not allow to "sweep" on those template parameters (i.e. degree) so it would be at least less conveniant than my solution (if I understand well the capabilities of this feature).

benchmarks/splines.cpp
```diff
@@ -47,7 +57,8 @@ void monitorMemoryAsync(std::mutex& mutex, bool& monitorFlag, size_t& maxUsedMem
     size_t freeMem = 0;
     size_t totalMem = 0;
     while (monitorFlag) {
-        std::this_thread::sleep_for(std::chrono::milliseconds(1)); // Adjust the interval as needed
+        std::this_thread::sleep_for(
+                std::chrono::microseconds(100)); // Adjust the interval as needed
```
blegouix, line 61:
```suggestion
                std::chrono::microseconds(10)); // Adjust the interval as needed
```

benchmarks/splines.cpp
```diff
@@ -14,27 +14,37 @@
 
 #include <benchmark/benchmark.h>
 
+static const ddc::SplineSolver Backend = ddc::SplineSolver::GINKGO;
```
blegouix, line 17:
```suggestion
static const ddc::SplineSolver Backend = ddc::SplineSolver::LAPACK;
```

benchmarks/splines.cpp
tpadioleau: I agree it does not seem to solve the problem.

However the current solution needs further improvement. If one updates the ranges of the benchmark, one also needs to think about updating `benchs` and the way to read inside.

tpadioleau: What do you think of a `std::map<std::array<long, 3>, std::function<void(benchmark::State&)>> benchs` ? You would read from it with `benchs.at({state.range(0), state.range(1), state.range(2)})(state)`.

blegouix: Sorry but I don't get the idea. Can you write a minimal chunk of code, like by sweeping only on one template argument (ie. Uniformity which take `std::false` or `std::true`) ?

tpadioleau:
```cpp
static void characteristics_advection(benchmark::State& state)
{
    long const host = 0;
    long const dev = 1;
    long const uniform = 0;
    long const non_uniform = 1;
    // Preallocate 12 unitary benchs for each combination of cpu/gpu execution space, uniform/non-uniform and spline degree we may want to benchmark (those are determined at compile-time, that's why we need to build explicitely 12 variants of the bench even if we call only one of them)
    std::map<std::array<long, 3>, std::function<void(benchmark::State&)>> benchs;
    benchs[std::array {host, uniform, 3}] = characteristics_advection_unitary<
            Kokkos::DefaultHostExecutionSpace,
            std::false_type,
            3>;
    benchs[std::array {host, uniform, 4}] = characteristics_advection_unitary<
            Kokkos::DefaultHostExecutionSpace,
            std::false_type,
            4>;
    benchs[std::array {host, uniform, 5}] = characteristics_advection_unitary<
            Kokkos::DefaultHostExecutionSpace,
            std::false_type,
            5>;
    benchs[std::array {host, non_uniform, 3}] = characteristics_advection_unitary<
            Kokkos::DefaultHostExecutionSpace,
            std::true_type,
            3>;
    benchs[std::array {host, non_uniform, 4}] = characteristics_advection_unitary<
            Kokkos::DefaultHostExecutionSpace,
            std::true_type,
            4>;
    benchs[std::array {host, non_uniform, 5}] = characteristics_advection_unitary<
            Kokkos::DefaultHostExecutionSpace,
            std::true_type,
            5>;
    benchs[std::array {dev, uniform, 3}]
            = characteristics_advection_unitary<Kokkos::DefaultExecutionSpace, std::false_type, 3>;
    benchs[std::array {dev, uniform, 4}]
            = characteristics_advection_unitary<Kokkos::DefaultExecutionSpace, std::false_type, 4>;
    benchs[std::array {dev, uniform, 5}]
            = characteristics_advection_unitary<Kokkos::DefaultExecutionSpace, std::false_type, 5>;
    benchs[std::array {dev, non_uniform, 3}] = characteristics_advection_unitary<Kokkos::DefaultExecutionSpace, std::true_type, 3>;
    benchs[std::array {dev, non_uniform, 4}]
            = characteristics_advection_unitary<Kokkos::DefaultExecutionSpace, std::true_type, 4>;
    benchs[std::array {dev, non_uniform, 5}]
            = characteristics_advection_unitary<Kokkos::DefaultExecutionSpace, std::true_type, 5>;

    // Run the desired bench
    benchs.at(std::array {state.range(0), state.range(1), state.range(2)})(state);
}
```

Something like this

benchmarks/splines.cpp
```diff
@@ -174,55 +199,155 @@ static void characteristics_advection(benchmark::State& state)
...
-// Tuning : 512 cols and 8 precond on CPU, 16384 cols and 1 precond on GPU
+static void characteristics_advection(benchmark::State& state)
+{
+    long const host = 0;
+    long const dev = 1;
+    long const uniform = 0;
+    long const non_uniform = 1;
+    // Preallocate 12 unitary benchs for each combination of cpu/gpu execution space, uniform/non-uniform and spline degree we may want to benchmark (those are determined at compile-time, that's why we need to build explicitely 12 variants of the bench even if we call only one of them)
+    std::map<std::array<long, 3>, std::function<void(benchmark::State&)>> benchs;
+    benchs[std::array {host, uniform, static_cast<long>(3)}] = characteristics_advection_unitary<
```
tpadioleau, line 221:
```suggestion
    benchs[std::array {host, uniform, 3L}] = characteristics_advection_unitary<
```

tpadioleau (approved):
Ok for me

benchmarks/splines.cpp
tpadioleau: Please apply to all of them

tpadioleau (approved):

PR #501, opened 2024-06-24
Doc minor fix again

blegouix:
Sorry for correcting the correction

### DIFF
--- a/include/ddc/kernels/splines/splines_linear_problem_sparse.hpp
+++ b/include/ddc/kernels/splines/splines_linear_problem_sparse.hpp
@@ -192,7 +192,7 @@ public:
      *
      * Removes the zeros from the CSR object and instantiate a Ginkgo solver. It also constructs a transposed version of the solver.
      *
-     * The stopping criterion is a reduction factor ||x||/||b||<1e-15 with 1000 maximum iterations.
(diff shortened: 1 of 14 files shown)


include/ddc/kernels/splines/splines_linear_problem_sparse.hpp
```diff
@@ -192,7 +192,7 @@ class SplinesLinearProblemSparse : public SplinesLinearProblem<ExecSpace>
      *
      * Removes the zeros from the CSR object and instantiate a Ginkgo solver. It also constructs a transposed version of the solver.
      *
-     * The stopping criterion is a reduction factor ||Ax-b||/||b||<1e-15 with 1000 maximum iterations.
+     * The stopping criterion is a reduction factor ||Ax-b||/||Ab-b||<1e-15 with 1000 maximum iterations.
```
tpadioleau, line 195: Are you sure of this ? Looks a bit weird to me. I am used to criterias like
- |Ax-b| < eps
- |Ax-b| < eps |b|

blegouix: You are right, I rechecked Ginkgo code and both options are available (rhs_norm and initial_resnorm, with initial x being b) but the default is rhs_norm.

PR #511, opened 2024-06-27
Rely on deep copies in SplineBuilder

### LINKED ISSUE #508: Rewrite SplineBuilder internal parallel_for_each with parallel_deep_copy
I think this loop (and the similar ones below) can be written using `Chunkspan[]`, `parallel_fill` and `parallel_deep_copy`:

https://github.com/CExA-project/ddc/blob/main/include/ddc/kernels/splines/spline_builder.hpp#L785

blegouix:
Closes #508


blegouix:
Sad but this branch is actually slower than the main (transpositions)... I wonder how `deepcopies` between LayoutRight and LayoutStride views on GPU could be slower (by a factor of ~3) than explicitely writing the `parallel_for` which do the same

blegouix:
I archive this in favor of #517

PR #515, opened 2024-07-01
Minor optimizations in splines linear problem maker

blegouix:
- Privilegie `PDSTridiag` or `PDSBand` block, even if band is width.

### DIFF
--- a/include/ddc/kernels/splines/splines_linear_problem_maker.hpp
+++ b/include/ddc/kernels/splines/splines_linear_problem_maker.hpp
@@ -126,11 +126,6 @@ public:
         assert(kl < n);
         assert(ku < n);
         int const bottom_size = std::max(kl, ku);
-        int const top_size = n - bottom_size;
-
-        if (bottom_size * (n + top_size) + (2 * kl + ku + 1) * top_size >= n * n) {
-            return std::make_unique<SplinesLinearProblemDense<ExecSpace>>(n);
-        }
 
         return make_new_block_matrix_with_band_main_block<ExecSpace>(n, kl, ku, pds, bottom_size);
     }
--- a/include/ddc/kernels/splines/splines_linear_problem_maker.hpp
+++ b/include/ddc/kernels/splines/splines_linear_problem_maker.hpp
@@ -126,6 +126,11 @@ public:
         assert(kl < n);
         assert(ku < n);
         int const bottom_size = std::max(kl, ku);
+        int const top_size = n - bottom_size;
+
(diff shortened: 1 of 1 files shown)


include/ddc/kernels/splines/splines_linear_problem_maker.hpp
```diff
@@ -54,12 +54,12 @@ class SplinesLinearProblemMaker
...
-        } else if (kl == ku && kl == 1 && pds) {
+        if (kl == ku && kl == 1 && pds) {
             return std::make_unique<SplinesLinearProblemPDSTridiag<ExecSpace>>(n);
         } else if (kl == ku && pds) {
             return std::make_unique<SplinesLinearProblemPDSBand<ExecSpace>>(n, kl);
+        } else if (2 * kl + ku + 1 >= n) {
+            return std::make_unique<SplinesLinearProblemDense<ExecSpace>>(n);
         } else {
             return std::make_unique<SplinesLinearProblemBand<ExecSpace>>(n, kl, ku);
         }
```
tpadioleau, line 65: Is this about restoring the logic from https://gitlab.maisondelasimulation.fr/gysela-developpers/gyselalibxx/-/blob/778a21b2111f66d9bf93374c31a2e1be64f2f0d6/vendor/sll/src/matrix.cpp#L80 ?

An other question though is why not using PDS band when not using a dense solver ? like
```suggestion
        if (kl == ku && kl == 1 && pds) {
            return std::make_unique<SplinesLinearProblemPDSTridiag<ExecSpace>>(n);
        } else if (2 * kl + ku + 1 >= n) {
            return std::make_unique<SplinesLinearProblemDense<ExecSpace>>(n);
        } else if (pds) {
            return std::make_unique<SplinesLinearProblemPDSBand<ExecSpace>>(n, kl);
        } else {
            return std::make_unique<SplinesLinearProblemBand<ExecSpace>>(n, kl, ku);
        }
```

tpadioleau (changes requested):

include/ddc/kernels/splines/splines_linear_problem_maker.hpp
blegouix: 1) Yes I suspect PDS solver to perform better than Dense even if the band if very wide. I did not checked it though. In practice for our use-cases we never have `2 * kl + ku + 1 >= n` anyway so it's really a detail.
2) The `2 * kl + ku + 1 >= n` comes from the additional allocation of size `kl` we have in `dgbtrf`. It does not apply to `dpbtrs`.

tpadioleau (approved):

PR #517, opened 2024-07-02
SplineBuilder minor optimization

blegouix:
- `spline_tr` was allocating too much rows (they were unused).
- Rely on ddc::parallel_fill and Kokkos::deep_copy to fill `splines` with `vals`

### DIFF
--- a/include/ddc/kernels/splines/spline_builder.hpp
+++ b/include/ddc/kernels/splines/spline_builder.hpp
@@ -329,7 +329,11 @@ private:
      */
     batched_spline_tr_domain_type batched_spline_tr_domain() const noexcept
     {
-        return batched_spline_tr_domain_type(spline_domain(), batch_domain());
+        return batched_spline_tr_domain_type(
+                batched_spline_domain().restrict(ddc::DiscreteDomain<bsplines_type>(
+                        ddc::DiscreteElement<bsplines_type>(0),
+                        ddc::DiscreteVector<bsplines_type>(
+                                ddc::discrete_space<bsplines_type>().nbasis()))));
     }
 
 public:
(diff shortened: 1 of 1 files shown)


include/ddc/kernels/splines/spline_builder.hpp
```diff
@@ -329,7 +329,11 @@ class SplineBuilder
      */
     batched_spline_tr_domain_type batched_spline_tr_domain() const noexcept
     {
-        return batched_spline_tr_domain_type(spline_domain(), batch_domain());
+        return batched_spline_tr_domain_type(
+                batched_spline_domain().restrict(ddc::DiscreteDomain<bsplines_type>(
+                        ddc::DiscreteElement<bsplines_type>(0),
+                        ddc::DiscreteVector<bsplines_type>(
+                                ddc::discrete_space<bsplines_type>().nbasis()))));
```
tpadioleau, line 336: Just to understand why it works with or without this change. This affects spline dimension which is the non-contiguous dimension in the linear system right ?

tpadioleau (commented):
Just one question, the change seems fine to me.

tpadioleau (changes requested):
Pass an instance of execution space to `parallel_fill` and `deep_copy`

include/ddc/kernels/splines/spline_builder.hpp
tpadioleau: I completely agree but I think it is not only a question about extents, looking at the source after this transpose loop we have
```cpp
    Kokkos::View<double**, Kokkos::LayoutRight, exec_space> bcoef_section(
            spline_tr.data_handle(),
            ddc::discrete_space<bsplines_type>().nbasis(),
            batch_domain().size());
```

If the dimension of interest would have been contiguous, this `bcoef_section` would not have been a valid view of the `spline_tr` array.

tpadioleau (approved):

PR #518, opened 2024-07-03
Optimize 3x3 blocks

blegouix:
Avoid deep_copies of `b_center`

Profiling on reference case with GREVILLE shows a 45% perf improvement (with KK on GPU) with this patch.

The downside of it is the SplinesLinearProblem3x3Blocks requires an allocation of `b` with `top_size` additional rows (which makes the SplineBuilder a bit more complex it already is)

### DIFF
--- a/include/ddc/kernels/splines/spline_builder.hpp
+++ b/include/ddc/kernels/splines/spline_builder.hpp
@@ -330,7 +330,7 @@ private:
      */
     batched_spline_tr_domain_type batched_spline_tr_domain() const noexcept
     {
-        int upper_block_size;
+        int upper_block_size = 0;
         if constexpr (Solver == ddc::SplineSolver::LAPACK && !bsplines_type::is_periodic()) {
             int lower_block_size;
             if constexpr (bsplines_type::is_uniform()) {
@@ -338,8 +338,6 @@ private:
             } else {
                 compute_block_sizes_non_uniform(lower_block_size, upper_block_size);
             }
-        } else if constexpr (Solver == ddc::SplineSolver::GINKGO) {
-            upper_block_size = 0;
         }
 
         return batched_spline_tr_domain_type(ddc::replace_dim_of<bsplines_type, bsplines_type>(
--- a/include/ddc/kernels/splines/spline_builder.hpp
+++ b/include/ddc/kernels/splines/spline_builder.hpp
@@ -330,11 +330,16 @@ private:
      */
     batched_spline_tr_domain_type batched_spline_tr_domain() const noexcept
     {
-        int lower_block_size, upper_block_size;
-        if constexpr (bsplines_type::is_uniform()) {
-            compute_block_sizes_uniform(lower_block_size, upper_block_size);
-        } else {
-            compute_block_sizes_non_uniform(lower_block_size, upper_block_size);
+        int upper_block_size;
+        if constexpr (Solver == ddc::SplineSolver::LAPACK) {
+            int lower_block_size;
+            if constexpr (bsplines_type::is_uniform()) {
+                compute_block_sizes_uniform(lower_block_size, upper_block_size);
+            } else {
+                compute_block_sizes_non_uniform(lower_block_size, upper_block_size);
+            }
+        } else if constexpr (Solver == ddc::SplineSolver::GINKGO) {
+            upper_block_size = 0;
         }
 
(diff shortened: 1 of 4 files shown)


tpadioleau:
> Avoid deep_copies of `b_center`
> 
> Profiling on reference case with GREVILLE shows a 45% perf improvement (with KK on GPU) with this patch.
> 
> The downside of it is the SplinesLinearProblem3x3Blocks requires an allocation of `b` with `top_size` additional rows (which makes the SplineBuilder a bit more complex it already is)

45% of which part ? the total execution time ?

blegouix:
yes

include/ddc/kernels/splines/spline_builder.hpp
```diff
@@ -324,17 +324,29 @@ class SplineBuilder
...
+        }
+
+        return batched_spline_tr_domain_type(ddc::replace_dim_of<bsplines_type, bsplines_type>(
+                batched_spline_domain(),
+                ddc::DiscreteDomain<bsplines_type>(
                         ddc::DiscreteElement<bsplines_type>(0),
                         ddc::DiscreteVector<bsplines_type>(
-                                ddc::discrete_space<bsplines_type>().nbasis()))));
+                                ddc::discrete_space<bsplines_type>().nbasis()
+                                + upper_block_size))));
```
tpadioleau, line 349: I am skeptical about this change. It seems difficult to maintain because we need to keep in mind the extra rows are specific to one particular solver. If we add a new LAPACK solver we may forget to update this.

Could we imagine adding a new virtual functions to solvers that would return the expected size for `B` ?

include/ddc/kernels/splines/splines_linear_problem_3x3_blocks.hpp
```diff
@@ -96,71 +96,92 @@ class SplinesLinearProblem3x3Blocks : public SplinesLinearProblem2x2Blocks<ExecS
...
+                        Kokkos::ALL);
+        MultiRHS const b_bottom_src = Kokkos::
+                subview(b,
+                        std::pair<
+                                std::size_t,
+                                std::size_t> {2 * m_top_size + nq, m_top_size + size()},
+                        Kokkos::ALL);
+
+        // Need a buffer to prevent overlapping
+        MultiRHS const buffer = Kokkos::create_mirror(ExecSpace(), b_bottom);
```
tpadioleau, line 173:
```suggestion
```

I guess this allocation can be removed ?

tpadioleau (changes requested):

include/ddc/kernels/splines/spline_builder.hpp
tpadioleau: Similar to [`required_span_size`](https://en.cppreference.com/w/cpp/container/mdspan/layout_right/mapping) on standard layouts

blegouix: Ok I will look at it

blegouix: thx for the proposal it's far better like that

include/ddc/kernels/splines/splines_linear_problem.hpp
```diff
@@ -78,6 +78,18 @@ class SplinesLinearProblem
...
+     * @brief Get the required number of rows of the multi-rhs view passed to solve().
+     *
+     * It is useful because SplinesLinearSolver3x3Blocks requires additional rows for internal operation.
+     *
+     * @return The required number of rows of the multi-rhs view.
+     */
+    virtual std::size_t required_number_of_rhs_rows() const
+    {
+        return m_size;
+    }
```
tpadioleau, line 92:
```suggestion
    /**
     * @brief Get the required number of rows of the multi-rhs view passed to solve().
     *
     * Implementations may require a number of rows larger than what `size` returns for optimization purposes.
     *
     * @return The required number of rows of the multi-rhs view. It is guaranteed to be greater or equal to `size`.
     */
    std::size_t required_number_of_rhs_rows() const
    {
        std::size_t const nrows = impl_required_number_of_rhs_rows();
        assert(nrows >= size());
        return nrows;
    }

    // in private
    virtual std::size_t impl_required_number_of_rhs_rows() const
    {
        return m_size;
    }
```

To avoid surprises

tpadioleau (changes requested):

tpadioleau (approved):
Looks ok to me. Please put in the report the timings before/after this optimization

PR #519, opened 2024-07-04
Bsplines relax float asserts

blegouix:
Necessary for ddc sync

### DIFF
--- a/include/ddc/kernels/splines/bsplines_non_uniform.hpp
+++ b/include/ddc/kernels/splines/bsplines_non_uniform.hpp
@@ -462,8 +462,8 @@ KOKKOS_INLINE_FUNCTION ddc::DiscreteElement<DDim> NonUniformBSplines<Tag, D>::
     std::array<double, degree()> left;
     std::array<double, degree()> right;
 
-    assert(x + Kokkos::abs(rmin()) * 1e-14 >= rmin());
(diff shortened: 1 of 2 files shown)


include/ddc/kernels/splines/bsplines_non_uniform.hpp
```diff
@@ -462,8 +462,8 @@ KOKKOS_INLINE_FUNCTION ddc::DiscreteElement<DDim> NonUniformBSplines<Tag, D>::
     std::array<double, degree()> left;
     std::array<double, degree()> right;
 
-    assert(x >= rmin());
-    assert(x <= rmax());
+    assert(x + Kokkos::abs(rmin()) * 1e-14 >= rmin());
```
tpadioleau, line 465: I would suggest to test a relative error like

```suggestion
    assert(Kokkos::abs(x - rmin()) < Kokkos::abs(rmin()) * 1E-14);
```

blegouix: I did the same with `length()` because if `rmin=0` no tolerancy is applied.

tpadioleau: Yes good idea. My test was wrong anyway it was checking near equality of `x` and `rmin()`

tpadioleau (approved):

PR #521, opened 2024-07-08
Characteristics advection non-periodic

blegouix:
Allow to run example in non-periodic setup, with GREVILLE and NullExtrapolation.

### DIFF
--- a/examples/characteristics_advection.cpp
+++ b/examples/characteristics_advection.cpp
@@ -12,6 +12,9 @@
 #include <ddc/kernels/splines.hpp>
 
 #include <Kokkos_Core.hpp>
+
+#define PERIODIC_DOMAIN // Comment this to run non-periodic simulation
+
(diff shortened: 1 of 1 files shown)


examples/characteristics_advection.cpp
```diff
@@ -12,26 +12,41 @@
...
 //! [X-discretization]
 /// A uniform discretization of X
 struct BSplinesX : ddc::UniformBSplines<X, s_degree_x>
 {
 };
-using GrevillePoints = ddc::GrevilleInterpolationPoints<
-        BSplinesX,
-        ddc::BoundCond::PERIODIC,
-        ddc::BoundCond::PERIODIC>;
+using GrevillePoints = ddc::
```
tpadioleau, line 48: I think we should use this type alias later in the file.

tpadioleau (approved):
Fine for me

PR #530, opened 2024-07-10
Coord<X> in benchmark

### LINKED ISSUE #528: Reduce memory footprint in the benchmark
In the benchmark

https://github.com/CExA-project/ddc/blob/b9807014ea78e103378f246dcef917a964b1ad38/benchmarks/splines.cpp#L142

we allocate 2D coordinates, only a single coordinate is necessary. I think we should use a 1D coordinate to reduce the memory footprint and have better memory accesses.

blegouix:
Closes #528

### DIFF
--- a/benchmarks/splines.cpp
+++ b/benchmarks/splines.cpp
@@ -139,7 +139,7 @@ static void characteristics_advection(benchmark::State& state)
     ddc::Chunk feet_coords_alloc(
             spline_builder.batched_interpolation_domain(),
             ddc::KokkosAllocator<
-                    ddc::Coordinate<X, Y>,
+                    ddc::Coordinate<X>,
                     Kokkos::DefaultExecutionSpace::memory_space>());
     ddc::ChunkSpan feet_coords = feet_coords_alloc.span_view();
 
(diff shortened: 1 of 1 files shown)


benchmarks/splines.cpp
```diff
@@ -148,10 +148,9 @@ static void characteristics_advection(benchmark::State& state)
         ddc::parallel_for_each(
                 feet_coords.domain(),
                 KOKKOS_LAMBDA(ddc::DiscreteElement<DDimX, DDimY> const e) {
-                    feet_coords(e) = ddc::Coordinate<X, Y>(
-                            ddc::coordinate(ddc::select<DDimX>(e))
-                                    - ddc::Coordinate<X>(0.0176429863),
-                            ddc::coordinate(ddc::select<DDimY>(e)));
+                    feet_coords(e) = ddc::coordinate(ddc::select<DDimX>(e))
+                                     - ddc::Coordinate<X>(0.0176429863),
+                    ddc::coordinate(ddc::select<DDimY>(e));
```
tpadioleau, line 153:
```suggestion
                    feet_coords(e) = ddc::coordinate(ddc::select<DDimX>(e)) - ddc::Coordinate<X>(0.0176429863);
```

tpadioleau (changes requested):
Ok after a small fix

tpadioleau (approved):
Ok for me

PR #533, opened 2024-07-10
Use alias GrevillePoints in characteristics_advection

### LINKED ISSUE #525: Use type alias `GrevillePoints` in the characteristics advection example
I think we should use this type alias later in the file.

_Originally posted by @tpadioleau in https://github.com/CExA-project/ddc/pull/521#discussion_r1668544896_

blegouix:
Close #525


tpadioleau (approved):
yep

PR #534, opened 2024-07-10
Left/Right -> Lower/Upper

### LINKED ISSUE #503: Use Lower/Upper boundary condition in place of left/right everywhere

blegouix:
Closes #503

### DIFF
--- a/include/ddc/kernels/splines/greville_interpolation_points.hpp
+++ b/include/ddc/kernels/splines/greville_interpolation_points.hpp
@@ -20,10 +20,10 @@ namespace ddc {
  * A class which provides helper functions to initialise the Greville points from a B-Spline definition.
  *
  * @tparam BSplines The bspline class relative to which the Greville points will be calculated.
- * @tparam BcXmin The lower boundary condition that will be used to build the splines.
- * @tparam BcXmax The upper boundary condition that will be used to build the splines.
+ * @tparam BcLower The lower boundary condition that will be used to build the splines.
+ * @tparam BcUpper The upper boundary condition that will be used to build the splines.
  */
-template <class BSplines, ddc::BoundCond BcXmin, ddc::BoundCond BcXmax>
+template <class BSplines, ddc::BoundCond BcLower, ddc::BoundCond BcUpper>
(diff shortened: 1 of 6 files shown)


include/ddc/kernels/splines/greville_interpolation_points.hpp
```diff
@@ -20,8 +20,8 @@ namespace ddc {
  * A class which provides helper functions to initialise the Greville points from a B-Spline definition.
  *
  * @tparam BSplines The bspline class relative to which the Greville points will be calculated.
- * @tparam BcXmin The (left) boundary condition that will be used to build the splines.
- * @tparam BcXmax The (right) boundary condition that will be used to build the splines.
+ * @tparam BcXmin The lower boundary condition that will be used to build the splines.
```
tpadioleau, line 23: Shall we also rename `BcXmin` ? `BcLower` ?

tpadioleau (commented):
Seems good

include/ddc/kernels/splines/greville_interpolation_points.hpp
blegouix: I agree. What about rmin(), rmax(), mixed_derivs_min1_max2, s_nbc_xmin ? Maybe in another MR just in case of problem with one of those commits ?

tpadioleau: Mmh, that would feel weird to me to go for `rupper`/`rlower`. Maybe we should do the reverse then, change the documentation to match the current notation `xmin`/`xmax` ?

blegouix: Or just use Lower/Upper in types (including BcLower, BcUpper) and min/max for variable ?

tpadioleau: So just `BcXmin` and `BcXmax` ? I am ok with it

tpadioleau (dismissed):
Ok for me

tpadioleau (approved):

PR #536, opened 2024-07-10
Remove is_spline_interpolation_mesh_uniform

### LINKED ISSUE #492: Deprecate is_spline_interpolation_mesh_uniform
The function `is_spline_interpolation_mesh_uniform` does not make sense. It is described as:
https://github.com/CExA-project/ddc/blob/b8ccaf41a5fa918146d968fb2e34096bc146c3a3/include/ddc/kernels/splines/spline_builder.hpp#L24-L35

However in reality it gives the uniform/non_uniform status of greville abcissae. The interpolation points have been decoupled from the greville abcissae for a while so this function should not be used. It should therefore be deprecated

blegouix:
Closes #492

### DIFF
--- a/include/ddc/kernels/splines/spline_builder.hpp
+++ b/include/ddc/kernels/splines/spline_builder.hpp
@@ -27,30 +27,6 @@ enum class SplineSolver {
 };
 
 /**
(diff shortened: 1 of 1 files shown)


tpadioleau (approved):
yes

PR #537, opened 2024-07-10
Add concept documentation

thierryantoun:
This PR aims to start concept documentation for DDC.


blegouix:
I think it may be useful for a new user coming from Fortran or Kokkos to understand how the memory layouts of the Chunks are determined with DDC.

In fact it is determined by the order of domains passed to a multidimensional DiscreteDomain constructor. The rightest is the most contiguous:

```
domain = DiscreteDomain<Tag1, Tag2>(dom1, dom2); // Cartesian product of dom1 and dom2
ddc::Chunk<double> chunk(domain); // Tag2 is the contiguous dimension
```

Afterward, it's interesting to note that:

```
chunk(DiscreteElement<Tag1>(0), DiscreteElement<Tag2>(1)) == chunk(DiscreteElement<Tag2>(1), DiscreteElement<Tag1>(0));
```

Which means dimensions are fully labellized and ordering has no impact anymore. I think this is quite evocative for a new user to understand the benefit of DDC, and constitutes a core concept.

(There are other ways to construct a DiscreteDomain but maybe it is not necessary to mention it)

docs/concepts.md
```diff
@@ -0,0 +1,85 @@
+# ddc concepts
+
+<!--
+Copyright (C) The ddc development team, see COPYRIGHT.md file
+
+SPDX-License-Identifier: MIT
+-->
+
+ddc introduces labels in the form of dimensions, and attributes on top of Kokkos views, which allows for a more intuitive, and less error-prone developer experience.
```
tpadioleau, line 9:
```suggestion
DDC introduces labels in the form of dimensions and attributes on top of Kokkos arrays, which allows for a more intuitive and less error-prone developer experience.
```

docs/concepts.md
```diff
@@ -0,0 +1,85 @@
...
+
+In fact, in Kokkos, the indices of views are weakly typed, meaning that each index is a simple integer. Let's consider a multidimensional view intended to represent a physical quantity, such as velocity in our example. The first index represents the velocity along the X-axis, and the second index represents the velocity along the Y-axis. In reality, there is nothing to distinguish between these two indices from the computer's perspective; they are both simply integers. This can lead to a situation where one mistakenly swaps the two indices. The code would compile successfully, but the resulting behavior would be incorrect, and the source of the error could be difficult to trace.
```
tpadioleau, line 11:
```suggestion
In fact, in Kokkos, the indices of views are weakly typed, meaning that each index is a simple integer. Let's consider a multidimensional view intended to represent a physical quantity, such as a velocity field for example. The first index represents the velocity along the X-axis, and the second index represents the velocity along the Y-axis. As both are simply integers this can lead to a situation where one mistakenly swaps the two indices. The code would compile successfully, but the resulting behavior would be incorrect, and the source of the error could be difficult to trace.
```

> The first index represents the velocity along the X-axis, and the second index represents the velocity along the Y-axis.

This is a bit confusing to me, it makes me think the first index represents somehow `u_x` and the second index represents `u_y`.

docs/concepts.md
```diff
@@ -0,0 +1,85 @@
...
+
+The advantage of using ddc is that it provides array-like containers that have labeled dimensions using strongly typed indices. For instance, by labeling dimensions by `X` and `Y`, the indices along those labeled dimensions become strongly typed preventing the user from making such mistakes.
```
tpadioleau, line 13:
```suggestion
The advantage of using DDC is that it provides array-like containers that have labeled dimensions using strongly typed indices. For instance, by labeling dimensions `X` and `Y`, the indices along those labeled dimensions become strongly typed preventing the user from making such mistakes.
```

docs/concepts.md
```diff
@@ -0,0 +1,85 @@
...
+
+> Note that the use of ddc is not restricted to solving equations. Indeed, one can easily imagine strong typing of variables corresponding to names or ages in a registry. The operations must then be adapted accordingly. Here, we largely base our approach on the uniform and non-uniform resolution of the heat equation in two dimensions, which is why we focus the presentation on solving equations using finite differences on a 2D grid.
```
tpadioleau, line 15:
```suggestion
\remark Note that the use of DDC is not restricted to solving equations. Indeed, one can easily imagine strong typing of variables corresponding to names or ages in a database. The operations must then be adapted accordingly. Here, we largely base our approach on the uniform and non-uniform resolution of the heat equation in two dimensions, which is why we focus the presentation on solving equations using finite differences on a 2D grid.
```

`>` is more for citation

docs/concepts.md
```diff
@@ -0,0 +1,85 @@
...
+
+## ddc::Chunk and ddc::ChunkSpan
+
+The `ddc::chunk` is a container that holds the data, while `ddc::chunkspan` behaves like `mdspan`, meaning they are pointers to the data contained within the chunk.
```
tpadioleau, line 19:
```suggestion
The `ddc::Chunk` is a container that holds the data, while `ddc::ChunkSpan` behaves like `std::mdspan`, meaning they are pointers to the data contained within the `ddc::Chunk`.
```

docs/concepts.md
```diff
@@ -0,0 +1,85 @@
...
+
+As mentioned in the introduction, ddc is a library that offers strongly typed indexing. Chunks contain data that can be located in space by coordinates. Thus, to access the data at a specific point in a 2D space, instead of entering two integers corresponding to the x and y positions, ddc requires entering the  coordinate `x` as a discrete element of the x position: `ddc::DiscreteElement<DDimX>`, and entering the `y` coordinate as a discrete element following the y dimension: `ddc::DiscreteElement<DDimY>`. This is done after predefining a strong typing for the discretized `X` dimension as `DDimX` and a strong typing for the discretized `Y` dimension as `DDimY` (see the heat equation example \subpage uniform_heat_equation "examples/uniform_heat_equation.cpp").
```
tpadioleau, line 21:
```suggestion
  As mentioned in the introduction, DDC is a library that offers strongly typed indexing. Chunks contain data that can be accessed by unique identifiers called discrete elements. Thus, to access the data at a specific point in a 2D space, instead of entering two integers corresponding to the 'x' and 'y' axes, DDC requires entering the coordinate `x` as a discrete element of the x position: `ddc::DiscreteElement<DDimX>`, and entering the `y` coordinate as a discrete element following the y dimension: `ddc::DiscreteElement<DDimY>`. This is done after predefining a strong typing for the discretized `X` dimension as `DDimX` and a strong typing for the discretized `Y` dimension as `DDimY` (see the heat equation example \subpage uniform_heat_equation "examples/uniform_heat_equation.cpp").
```

I am still not convinced we should talk about discretization in this page.

docs/concepts.md
```diff
@@ -0,0 +1,85 @@
...
+
+Note that swapping the `ddc::DiscreteElement<DDimX>` and `ddc::DiscreteElement<DDimY>` indices when calling the chunkspan does not affect the correctness of the code; the result remains the same.
```
tpadioleau, line 23:
```suggestion
Note that swapping the `ddc::DiscreteElement<DDimX>` and `ddc::DiscreteElement<DDimY>` indices when calling the `ddc::ChunkSpan` does not affect the correctness of the code; the result remains the same.
```

docs/concepts.md
```diff
@@ -0,0 +1,85 @@
...
+
+## ddc::DiscreteElement, ddc::DiscreteVector and ddc::Coordinate
+
+### ddc::DiscreteElement
+
+Let's continue with our previous example of a 2D grid along two discretized dimensions labeled as `DDimX` and `DDimY`. In the previous paragraph, we discussed how `ddc::DiscreteElement` could be used as indices to have access to a physical value at a precised point on the 2D grid.
```
tpadioleau, line 29:
```suggestion
Let's continue with our previous example of a 2D grid along two discretized dimensions labeled as `DDimX` and `DDimY`. In the previous section, we discussed how `ddc::DiscreteElement` could be used as indices to have access to a physical value at a precised point on the 2D grid.
```

docs/concepts.md
```diff
@@ -0,0 +1,85 @@
...
+
+More precisely, a `ddc::DiscreteElement` is a C++ variable that carries the strong type of the dimension in which it is defined.
```
tpadioleau, line 31:
```suggestion
More precisely, a `ddc::DiscreteElement` is a C++ type that carries the label of the dimension in which it is defined.
```

docs/concepts.md
```diff
@@ -0,0 +1,85 @@
...
+
+carries the strong typing of the ordinate dimension `DDimY` and corresponds to y<sub>0</sub> , the first point along the `Y` dimension.
+
+Moreover, `ddc::DiscreteElement` are very useful for another reason. If we take the example of a classic container in C++, let's say we want to access the element `(i,j)` of this container which corresponds to the grid point at the i th row and j th column, we would do it like this:
+
+```cpp
+container(i,j);
+```
+
+Now, if we take a slice of this container and still want to access the same element `(i,j)` from the grid, we will need to adjust the indices because the indexing of the new sliced container along each dimension starts at zero. However, with ddc, this is not the case. If we take a slice of a chunkspan, accessing a `ddc::DiscreteElement` is the same between the slice and the original chunkspan because of the uniqueness of each discrete element on the grid and because of the way we access data using ddc.
```
tpadioleau, line 46:
```suggestion
Now, if we take a slice of this container and still want to access the same element `(i,j)` from the grid, we will need to adjust the indices because the indexing of the new sliced container along each dimension starts at zero. However, with DDC, this is not the case. If we take a slice of a `ddc::ChunkSpan`, accessing a `ddc::DiscreteElement` is the same between the slice and the original `ddc::ChunkSpan` because of the uniqueness of each discrete element on the grid and because of the way we access data using DDC.
```

docs/concepts.md
```diff
@@ -0,0 +1,85 @@
...
+
+### ddc::Coordinate
+
+A `ddc::Coordinate` is one of the only ddc types which represents a continuous data type. This means that it can take any value that can be represented by a double. It represents the position of a coordinate in the vector space.
+
+Let's take the same discretization of the Y-axis {y<sub>0</sub>, y<sub>1</sub>, ..., y<sub>n</sub>}. We saw that ```cpp ddc::DiscreteElement<DDimY> y(0)``` points to the first element y<sub>0</sub> along the Y-axis. Now if we take ```cpp ddc::Coordinate<DDimY> (ddc::DiscreteElement<DDimY> y(0))``` it will correspond to the actual value of y<sub>0</sub> in the simulation. If the first point is at a position `y = -1`, it will be characterized by a `ddc::DiscreteElement` as the first point of the discretization, but physically its position is -1 and this is the value that `ddc::Coordinate` of this element will be.
+
+The `ddc::Coordinate` thus bridges the gap between the discretization accessible by `ddc::DiscreteElement` and the actual physical value of the coordinate.
```
tpadioleau, line 54:
```suggestion
```

docs/concepts.md
```diff
@@ -0,0 +1,85 @@
...
+
+### ddc::DiscreteVector
+
+A `ddc::DiscreteVector` corresponds to an integer that, like `ddc::DiscreteElement`, carries the strong typing of the dimension in which it is defined. For instance in the uniform heat equation example, defining the `ddc::DiscreteVector` `gwx` as follows:
```
tpadioleau, line 58:
```suggestion
A `ddc::DiscreteVector` corresponds to an integer that, like `ddc::DiscreteElement`, carries the label of the dimension in which it is defined. For instance in the uniform heat equation example, defining the `ddc::DiscreteVector` `gwx` as follows:
```

docs/concepts.md
```diff
@@ -0,0 +1,85 @@
...
+
+```cpp
+ddc::DiscreteVector<DDimX> gwx(1);
+```
+
+is equivalent to defining a number of points, here 1, along the `x` dimension.
+
+> Note that the difference between two `ddc::DiscreteElement` creates a `ddc::DiscreteVector`, and the sum of a `ddc::DiscreteVector` and a `ddc::DiscreteElement` results in a `ddc::DiscreteElement`. This illustrates how `ddc::DiscreteElement` could correspond to points in an affine space, while `ddc::DiscreteVector` could correspond to vectors in a vector space or to a distance between two points.
```
tpadioleau, line 66:
```suggestion
\remark Note that the difference between two `ddc::DiscreteElement` creates a `ddc::DiscreteVector`, and the sum of a `ddc::DiscreteVector` and a `ddc::DiscreteElement` results in a `ddc::DiscreteElement`. This illustrates how `ddc::DiscreteElement` could correspond to points in an affine space, while `ddc::DiscreteVector` could correspond to vectors in a vector space or to a distance between two points.
```

I think it shall be better illustrated with code than sentences.

docs/concepts.md
```diff
@@ -0,0 +1,85 @@
...
+
+In summary;
+
++ `ddc::DiscreteElement` corresponds to each unique point of the mesh, fixed throughout the duration of the simulation. They are similar to fixed points in an affine space.
```
tpadioleau, line 70: Could we formulate it without the notion of mesh/point ?

docs/concepts.md
```diff
@@ -0,0 +1,85 @@
...
++ `ddc::Coordinate` represents the actual value of the discretized points on the grid indexed with the `ddc::DiscreteElement`
```
tpadioleau, line 71:
```suggestion
```

docs/concepts.md
```diff
@@ -0,0 +1,85 @@
...
++ On the other hand, `ddc::DiscreteVector` corresponds to a number of points along a particular axis or to a distance between two points; in ddc, it corresponds to a distance between two DiscreteElements. These are integers that carry the strong typing of the dimension in which they are defined.
+
+## ddc::DiscreteDomain
+
+As mentioned earlier, ddc operates on a coordinate system. These coordinates are part of a domain. Users must start their code by constructing a `ddc::DiscreteDomain` that contains each `ddc::DiscreteElement`. To construct a domain, you need to build a 1D domain along each direction. This can be done using the function `ddc::init_discrete_space`.
```
tpadioleau, line 76:
```suggestion
As mentioned earlier, DDC operates on a coordinate system. These coordinates are part of a domain. Users must start their code by constructing a `ddc::DiscreteDomain` that contains each `ddc::DiscreteElement`. To construct a domain, you need to build a 1D domain along each direction. This is usually done when calling the function `ddc::init_discrete_space`.
```

docs/concepts.md
```diff
@@ -0,0 +1,85 @@
...
+
+Note that it is possible to construct a domain with both uniform (see \subpage uniform_heat_equation "examples/uniform_heat_equation.cpp") and non-uniform (see \subpage non_uniform_heat_equation "examples/non_uniform_heat_equation.cpp") distribution of points.
```
tpadioleau, line 78:
```suggestion
```

docs/concepts.md
```diff
@@ -0,0 +1,85 @@
...
+
+## Algorithms in ddc ?
+
++ `for_each`: this algorithm allows iterating over each DiscreteElement in the domain of study.
```
tpadioleau, line 82:
```suggestion
+ `for_each`: this algorithm allows iterating over each `ddc::DiscreteElement` in the domain of study.
```

tpadioleau (changes requested):
I still feel like this page should not contain anything related to discretization.

tpadioleau:
Please rebase on main

tpadioleau:
I can continue to work on this after my return.

docs/concepts.md
thierryantoun: Oh ok I think it can be misleading. Will try to make it clearer

docs/concepts.md
thierryantoun: Ok

docs/concepts.md
thierryantoun: I will try but I'm afraid It might be a bit too abstract

pzehner (changes requested):
I think it's a good idea to gather the core concepts of the library in one place. Unfortunately, I don't think it is clear the way it is for now.

I'll be harsh, I've many things to say about this PR. Please don't take it personally, or feel demotivated about it. I know you put a lot of effort in this work, I'm not negating this.

To start with, using one line per sentence for Markdown documents is better for reviewing and versioning, as diffs and reviews are more atomic.

I found the document confused, with too many convolutions, repetitions and articulation sentences. The English formulation is too close too French. For that, I did a lot of reformulations. It would have been good to go through them together. We can still do that whenever you want in the office. Improving your English writing skills is something you will acquire through experience.

Some concepts are misleading. We discussed in the office earlier about the meaning of `ddc::DiscreteElement` as in the document, it could be misunderstood as the physical coordinate of a point in the grid, not as an index.

The different sections should be more isolated and distinct. As it is, there are classes introduced in the wrong sections (e.g. `ddc::DiscreteElement` introduced in `ddc::Chunk`), and classes introduced after being used to define other objects (e.g. `ddc::DiscreteElement<DDimX>` is introduced after `DDimX`). It's not clear for me when this document should be read by the developer, either at the very beginning when exploring the lib or at the end in order to refresh knowledge when having a doubt about one concept. In the later case, the order of definition is less relevant, but for the sake of completeness, the description of one concept should stand in one (sub-)section, and not be scattered around.

The order of the different sections should be changed. Either adopt a top-bottom (dimension, domain, chunk, chunk span, discrete element or vector) or a bottom-up strategy. This is a bit cosmetic, but gives an idea of progression when reading it.

Given that this document exposes the very core logic of DDC in a condensed manner, I think that @tpadioleau should be more involved in its redaction, not only in its review. What I mean is, nobody has a better understanding of these concepts than him. He should be able to expose them clearly.

docs/concepts.md
```diff
@@ -0,0 +1,75 @@
...
+
+The advantage of using DDC is that it provides array-like containers that have labeled dimensions using strongly typed indices. For instance, by labeling dimensions `X` and `Y`, the indices along those labeled dimensions become strongly typed preventing the user from making such mistakes.
+
+\remark Note that the use of DDC is not restricted to solving equations. Indeed, one can easily imagine strong typing of variables corresponding to names or ages in a database. The operations must then be adapted accordingly. Here, we largely base our approach on the uniform and non-uniform resolution of the heat equation in two dimensions, which is why we focus the presentation on solving equations using finite differences on a 2D grid.
```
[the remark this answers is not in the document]
thierryantoun: By "here" I meant "In this document". Maybe better to make it explicit as you say.

I'm not sure if it's really relevant to mention the 'data' aspect of DDC here; in my opinion, it might confuse the reader. Perhaps we should develop a separate example and dedicate documentation to it ?

docs/concepts.md
```diff
@@ -0,0 +1,75 @@
...
+
+The `ddc::Chunk` is a container that holds the data, while `ddc::ChunkSpan` behaves like `std::mdspan`, meaning they are pointers to the data contained within the `ddc::Chunk`.
+
+  As mentioned in the introduction, DDC is a library that offers strongly typed indexing. Chunks contain data that can be accessed by unique identifiers called discrete elements. Thus, to access the data at a specific point in a 2D space, instead of entering two integers corresponding to the 'x' and 'y' axes, DDC requires entering the coordinate `x` as a discrete element of the x position: `ddc::DiscreteElement<DDimX>`, and entering the `y` coordinate as a discrete element following the y dimension: `ddc::DiscreteElement<DDimY>`. This is done after predefining a strong typing for the discretized `X` dimension as `DDimX` and a strong typing for the discretized `Y` dimension as `DDimY` (see the heat equation example \subpage uniform_heat_equation "examples/uniform_heat_equation.cpp").
+
+Note that swapping the `ddc::DiscreteElement<DDimX>` and `ddc::DiscreteElement<DDimY>` indices when calling the `ddc::ChunkSpan` does not affect the correctness of the code; the result remains the same.
```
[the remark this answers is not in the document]
thierryantoun: agree

docs/concepts.md
```diff
@@ -0,0 +1,75 @@
...
+
+In summary;
+
++ `ddc::DiscreteElement` corresponds to each unique point of the mesh, fixed throughout the duration of the simulation. They are similar to fixed points in an affine space.
++ On the other hand, `ddc::DiscreteVector` corresponds to a number of points along a particular axis or to a distance between two points; in ddc, it corresponds to a distance between two DiscreteElements. These are integers that carry the strong typing of the dimension in which they are defined.
+
+## ddc::DiscreteDomain
+
+As mentioned earlier, DDC operates on a coordinate system. These coordinates are part of a domain. Users must start their code by constructing a `ddc::DiscreteDomain` that contains each `ddc::DiscreteElement`. To construct a domain, you need to build a 1D domain along each direction. This is usually done when calling the function `ddc::init_discrete_space`.
```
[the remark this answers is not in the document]
thierryantoun: I think there are other ways to do it but I just used this method in the examples. If it's the only way to do it, I'll change the sentence you're right.

tpadioleau (approved):

PR #539, opened 2024-07-11
Rename mesh -> discrete_dimension in splines

blegouix:
`interpolation_mesh_type -> interpolation_discrete_dimension_type`
`InterpolationMesh -> InterpolationDDim`

Etc...

### DIFF
--- a/include/ddc/kernels/splines/bsplines_non_uniform.hpp
+++ b/include/ddc/kernels/splines/bsplines_non_uniform.hpp
@@ -89,7 +89,7 @@ public:
 
     public:
         /// @brief The type of the knots defining the B-splines.
-        using knot_mesh_type = NonUniformBsplinesKnots<DDim>;
+        using knot_discrete_dimension_type = NonUniformBsplinesKnots<DDim>;
 
         /// @brief The type of the discrete dimension representing the B-splines.
         using discrete_dimension_type = NonUniformBSplines;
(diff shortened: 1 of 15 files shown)


benchmarks/splines.cpp
```diff
@@ -148,7 +148,8 @@ static void characteristics_advection(benchmark::State& state)
         ddc::parallel_for_each(
                 feet_coords.domain(),
                 KOKKOS_LAMBDA(ddc::DiscreteElement<DDimX, DDimY> const e) {
-                    feet_coords(e) = ddc::coordinate(ddc::select<DDimX>(e)) - ddc::Coordinate<X>(0.0176429863);
+                    feet_coords(e) = ddc::coordinate(ddc::select<DDimX>(e))
+                                     - ddc::Coordinate<X>(0.0176429863);
```
blegouix, line 152: weird

tpadioleau (approved):
Looks ok to me

benchmarks/splines.cpp
tpadioleau: could be that the CI does not check this file ?

tpadioleau: https://github.com/CExA-project/ddc/blob/0da9a03f43e761848c57b7536f7217730a54a323/.github/workflows/tests.yml#L19 seems to not be checked. So maybe a different version of clang-format has been used leading to a different formatting

blegouix: #547

PR #541, opened 2024-07-12
Simplify fft + fix PeriodicSampling window

blegouix:
Rely on the formula:

```
k=2*pi*(N-1)^2/N/(xmax-xmin)
```

equivalent to:

```
k=2*pi*(N-1)/(b-a)
```

To identify the coordinate of the last point of the spectral mesh.

Also fix the misalignment to `numpy.fftfreq` of the convention to describe a periodic window.


tpadioleau:
Please add tests about the coordinates generated by `init_fourier_space`

include/ddc/periodic_sampling.hpp
```diff
@@ -127,8 +131,11 @@ class PeriodicSampling : detail::PeriodicSamplingBase
         {
             return m_origin
                    + continuous_element_type(
-                             static_cast<int>((icoord.uid() + m_n_period / 2 - 1) % m_n_period)
-                             - static_cast<int>(m_n_period / 2 - 1))
+                             static_cast<int>(
+                                     (icoord.uid()
+                                      + (m_n_period - static_cast<int>(m_odd_period)) / 2)
+                                     % m_n_period)
+                             - static_cast<int>((m_n_period - static_cast<int>(m_odd_period)) / 2))
```
blegouix, line 138: This code was not aligned with `numpy.fftfreq` convention to describe the periodic window. Now it is

blegouix: https://numpy.org/doc/stable/reference/generated/numpy.fft.fftfreq.html

tests/fft.cpp
```diff
@@ -29,6 +29,31 @@ struct DFDim : ddc::PeriodicSampling<Kx>
 {
 };
 
+template <typename X, std::size_t Nx>
+static void test_fourier_mesh()
```
tpadioleau, line 33:
```suggestion
template <typename X>
static void test_fourier_mesh(std::size_t Nx)
```

Maybe too much to pass `Nx` as a template parameter in this case. This is necessary when we want to generate discrete dimensions on the same continuous dimensions in the same run.

tests/fft.cpp
```diff
@@ -29,6 +29,31 @@ struct DFDim : ddc::PeriodicSampling<Kx>
...
+
+    DDom<DDim<X>> const x_mesh(ddc::init_discrete_space<DDim<X>>(DDim<X>::template init<DDim<X>>(
+            ddc::Coordinate<X>(a + (b - a) / Nx / 2),
+            ddc::Coordinate<X>(b - (b - a) / Nx / 2),
+            DVect<DDim<X>>(Nx))));
+    ddc::init_discrete_space<DFDim<ddc::Fourier<X>>>(
+            ddc::init_fourier_space<DFDim<ddc::Fourier<X>>>(ddc::DiscreteDomain<DDim<X>>(x_mesh)));
+
+    double const epsilon = 1e-14;
+    for (int i = 0; i < Nx; i++) {
```
tpadioleau, line 46: Isn't the role of `FourierMesh` to generate the `DiscreteDomain` related to Fourier ?

tests/fft.cpp
```diff
@@ -29,6 +29,31 @@ struct DFDim : ddc::PeriodicSampling<Kx>
...
+        double const ka = -Kokkos::numbers::pi * Nx / (b - a);
+        double const kb = Kokkos::numbers::pi * Nx / (b - a);
+        double const k = 2 * i * Kokkos::numbers::pi / (b - a);
+        EXPECT_NEAR(
+                ddc::coordinate(ddc::DiscreteElement<DFDim<ddc::Fourier<X>>>(i)),
+                k - (kb - ka) * Kokkos::floor((k - ka) / (kb - ka)),
```
tpadioleau, line 52: I do not get why we have to add this extra computation ?

tests/fft.cpp
```diff
@@ -199,6 +224,16 @@ struct RDimX;
...
 
+TEST(FourierMesh, Even)
+{
+    test_fourier_mesh<RDimX, 16>();
+}
+
+TEST(FourierMesh, Odd)
+{
+    test_fourier_mesh<RDimX, 17>();
+}
```
tpadioleau, line 235:
```suggestion
TEST(FourierMesh, Even)
{
    test_fourier_mesh<RDimX>(16);
}

TEST(FourierMesh, Odd)
{
    test_fourier_mesh<RDimX>(17);
}
```

include/ddc/periodic_sampling.hpp
tpadioleau: I guess it is necessary to get proper convergence for spectral methods ?

Anyway it is a good thing

include/ddc/periodic_sampling.hpp
```diff
@@ -127,8 +131,11 @@ class PeriodicSampling : detail::PeriodicSamplingBase
         {
             return m_origin
                    + continuous_element_type(
-                             static_cast<int>((icoord.uid() + m_n_period / 2 - 1) % m_n_period)
-                             - static_cast<int>(m_n_period / 2 - 1))
+                             static_cast<int>(
+                                     (icoord.uid()
+                                      + (m_n_period - static_cast<int>(m_odd_period)) / 2)
```
tpadioleau, line 136: About the computation `(m_n_period - static_cast<int>(m_odd_period)) / 2` is `m_odd_period` really necessary ? I tend to think the integer division already handles it fine ?

tests/fft.cpp
blegouix: if `i>(N- N%2) / 2`, `ddc::coordinate` projects in the periodic window.

Initially I had the following test in mind (test only the points in the periodic window):

https://github.com/CExA-project/ddc/pull/541/commits/43124194762b83e26fd3603b1f57e43fb098ab9a#diff-3204c222afc48502013a0d15cc15f3e45be803127fb2f6f9720397ad7157e3bdL218

But it cannot work because `ddc` does not support `ddc::DiscreteElement(negative_integer)` (I mean, there is a cast involved. Actually the test passes, but by chance ^^. Using `std::ptrdiff_t` for `uid` types would fix the problem, but I like this new version of the test which indirectly also tests `PeriodicSampling`.

include/ddc/periodic_sampling.hpp
blegouix: Ah yeah that's correct, thx

include/ddc/periodic_sampling.hpp
blegouix: That's a very good question, I am not sure of the answer. The computation is numerically different when we divide by /k^2 for sure, but I wonder if by the magic of Fourier transform it could become the same after performing the iFFT. In my understanding the choice of position of the periodic window is supposed to be arbitrary, so it would be good if it has no mathematical impact. But I have no proof for that :/

tests/fft.cpp
blegouix: Yes, but I would say we don't need it in 1D. `FourierMesh` is more to generate the appropriate mesh to perform the FFT after ward (taking in account C2C or not, etc.)

tpadioleau: Yes but then we are creating `DiscreteElement` from raw integers. It is like using `uid()`.

blegouix: [9fac769](https://github.com/CExA-project/ddc/pull/541/commits/9fac769182a53dd73e45950f8d045bf09da7c045) ?

tpadioleau (approved):
Seems good to me

PR #556, opened 2024-07-19
Comment about rounding errors

tpadioleau:


EmilyBourne (approved):

tpadioleau:
Ideally I would like to give error bounds but I don't know yet how to prove them.

Issue #542, opened 2024-07-12
DDC `create_mirror[_view][_and_copy]` does not follow synchronous behavior of Kokkos


tpadioleau:

tpadioleau:
closed by #558

PR #558, opened 2024-07-20
Improve `create_mirror[_view][_and_copy]` functions

tpadioleau:
- Fix asynchronous behavior using `parallel_deepcopy(dst, src)`.
- Allow to pass a memory space.
- Improve documentation.

### DIFF
--- a/include/ddc/create_mirror.hpp
+++ b/include/ddc/create_mirror.hpp
@@ -49,6 +49,12 @@ auto create_mirror_and_copy(
         Space const& space,
         ChunkSpan<ElementType, Support, Layout, MemorySpace> const& src)
 {
+    static_assert(
+            Kokkos::is_memory_space_v<Space> || Kokkos::is_execution_space_v<Space>,
+            "DDC: parameter \"Space\" must be either a Kokkos execution space or a memory space");
+    static_assert(
+            std::is_same_v<Layout, std::experimental::layout_right>,
+            "DDC: parameter \"Layout\" must be a `layout_right`");
     Chunk chunk = create_mirror(space, src);
     parallel_deepcopy(chunk, src);
     return chunk;
--- a/include/ddc/create_mirror.hpp
+++ b/include/ddc/create_mirror.hpp
@@ -13,66 +13,73 @@
 
 namespace ddc {
 
-/// Returns a new `Chunk` with the same layout as `src` allocated on the memory space `Space::memory_space`.
+/// @param[in] space A Kokkos memory space or execution space.
+/// @param[in] src A layout right ChunkSpan.
+/// @return a `Chunk` with the same support and layout as `src` allocated on the `Space::memory_space` memory space.
 template <class Space, class ElementType, class Support, class Layout, class MemorySpace>
 auto create_mirror(
         [[maybe_unused]] Space const& space,
         ChunkSpan<ElementType, Support, Layout, MemorySpace> const& src)
 {
     static_assert(
+            Kokkos::is_memory_space_v<Space> || Kokkos::is_execution_space_v<Space>,
+            "DDC: parameter \"Space\" must be either a Kokkos execution space or a memory space");
+    static_assert(
             std::is_same_v<Layout, std::experimental::layout_right>,
-            "Only layout right is supported");
+            "DDC: parameter \"Layout\" must be a `layout_right`");
     return Chunk(
             src.domain(),
             KokkosAllocator<std::remove_const_t<ElementType>, typename Space::memory_space>());
 }
 
-/// Returns a new host `Chunk` with the same layout as `src`.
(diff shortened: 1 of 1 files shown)


tpadioleau:
closes #542

include/ddc/create_mirror.hpp
```diff
@@ -13,118 +13,117 @@
...
+/// @param[in] src A layout right ChunkSpan.
+/// @return a `Chunk` with the same support and layout as `src` allocated on the `Space::memory_space` memory space and operates a deep copy between the two.
 template <class Space, class ElementType, class Support, class Layout, class MemorySpace>
 auto create_mirror_and_copy(
         Space const& space,
         ChunkSpan<ElementType, Support, Layout, MemorySpace> const& src)
 {
-    static_assert(
-            std::is_same_v<Layout, std::experimental::layout_right>,
-            "Only layout right is supported");
```
EmilyBourne, line 48: Why is this static assert removed?

include/ddc/create_mirror.hpp
```diff
@@ -13,118 +13,117 @@
...
     Chunk chunk = create_mirror(space, src);
-    parallel_deepcopy(space, chunk, src);
+    parallel_deepcopy(chunk, src);
```
EmilyBourne, line 53: What is the default space for `parallel_deepcopy`?

include/ddc/create_mirror.hpp
```diff
@@ -13,118 +13,117 @@
...
+/// @return If `MemorySpace` is accessible from `Space` then returns a copy of `src`, otherwise returns a `Chunk` with the same support and layout as `src` allocated on the `Space::memory_space` memory space.
 template <class Space, class ElementType, class Support, class Layout, class MemorySpace>
 auto create_mirror_view(
-        Space const& space,
+        [[maybe_unused]] Space const& space,
         ChunkSpan<ElementType, Support, Layout, MemorySpace> const& src)
 {
+    static_assert(
+            std::is_const_v<ElementType>,
+            "DDC: parameter \"ElementType\" must not be `const`");
```
EmilyBourne, line 76: Comment does not match condition
```suggestion
    static_assert(
            !std::is_const_v<ElementType>,
            "DDC: parameter \"ElementType\" must not be `const`");
```

include/ddc/create_mirror.hpp
```diff
@@ -13,118 +13,117 @@
...
-    parallel_deepcopy(chunk, src);
-    return chunk;
+    return create_mirror_and_copy(Kokkos::HostSpace(), src);
 }
 
-/// If `src` is accessible from `space` then returns a copy of `src`,
-/// otherwise returns a new `Chunk` with the same layout as `src` allocated on the memory space `Space::memory_space`.
+/// @param[in] space A Kokkos memory space or execution space.
+/// @param[in] src A non-const, layout right ChunkSpan.
+/// @return If `MemorySpace` is accessible from `Space` then returns a copy of `src`, otherwise returns a `Chunk` with the same support and layout as `src` allocated on the `Space::memory_space` memory space.
```
EmilyBourne, line 68: Not sure if you care about line length in DDC?
```suggestion
/// @return If `MemorySpace` is accessible from `Space` then returns a copy of `src`, otherwise returns a `Chunk`
///         with the same support and layout as `src` allocated on the `Space::memory_space` memory space.
```

EmilyBourne (changes requested):

include/ddc/create_mirror.hpp
tpadioleau: As this function calls `create_mirror` which already does the check I removed it. I don't have yet a clear rule when to repeat the `static_assert` or not. Maybe I should reintroduce the check here because the function is not simple enough to remove it ?

include/ddc/create_mirror.hpp
tpadioleau: There is no default execution space. Kokkos ensures that the copy will begin after any kernel and will be finished when returning from this function.

include/ddc/create_mirror.hpp
tpadioleau: It makes me realize we don't have enabled `ReflowComments` in clang-format. I guess we should do it.

include/ddc/create_mirror.hpp
tpadioleau: Thanks

include/ddc/create_mirror.hpp
EmilyBourne: Up to you if it exists elsewhere then this should be fine

include/ddc/create_mirror.hpp
EmilyBourne: Probably in a separate PR though I guess?

tpadioleau: Yep

PR #560, opened 2024-07-22
Splines quadrature coefficients getter

blegouix:
Related to https://gitlab.maisondelasimulation.fr/gysela-developpers/gyselalibxx/-/issues/263

### DIFF
--- a/include/ddc/kernels/splines/spline_builder.hpp
+++ b/include/ddc/kernels/splines/spline_builder.hpp
@@ -397,6 +397,13 @@ public:
                     derivs_xmax
             = std::nullopt) const;
 
+    ddc::Chunk<
+            double,
+            ddc::DiscreteDomain<interpolation_discrete_dimension_type>,
+            ddc::KokkosAllocator<double, memory_space>>
+    quadrature_coefficients(
+            ddc::DiscreteDomain<interpolation_discrete_dimension_type> const& domain) const;
+
 private:
     void compute_block_sizes_uniform(int& lower_block_size, int& upper_block_size) const;
 
@@ -892,5 +899,70 @@ operator()(
                     }
                 });
     }
+}
+
+template <
+        class ExecSpace,
+        class MemorySpace,
+        class BSplines,
+        class InterpolationDDim,
+        ddc::BoundCond BcLower,
+        ddc::BoundCond BcUpper,
+        SplineSolver Solver,
+        class... IDimX>
+ddc::Chunk<
+        double,
+        ddc::DiscreteDomain<InterpolationDDim>,
+        ddc::KokkosAllocator<double, MemorySpace>>
+SplineBuilder<
+        ExecSpace,
+        MemorySpace,
+        BSplines,
+        InterpolationDDim,
+        BcLower,
+        BcUpper,
+        Solver,
+        IDimX...>::quadrature_coefficients(ddc::DiscreteDomain<InterpolationDDim> const& domain)
+        const
+{
+    ddc::Chunk<double, ddc::DiscreteDomain<bsplines_type>> integral_bsplines(spline_domain());
+    ddc::discrete_space<bsplines_type>().integrals(integral_bsplines.span_view());
+
+    // Solve matrix equation
+    ddc::ChunkSpan integral_bsplines_without_periodic_point
+            = integral_bsplines.span_view()[ddc::DiscreteDomain<bsplines_type>(
+                    ddc::DiscreteElement<bsplines_type>(0),
+                    ddc::DiscreteVector<bsplines_type>(matrix->size()))];
+    Kokkos::View<double**, Kokkos::LayoutRight, Kokkos::DefaultHostExecutionSpace>
+            integral_bsplines_mirror_with_additional_allocation(
+                    "integral_bsplines_mirror_with_additional_allocation",
+                    matrix->required_number_of_rhs_rows(),
+                    1);
+    Kokkos::View<double*, Kokkos::LayoutRight, Kokkos::DefaultHostExecutionSpace>
+            integral_bsplines_mirror = Kokkos::
+                    subview(integral_bsplines_mirror_with_additional_allocation,
+                            std::pair<std::size_t, std::size_t> {
+                                    0,
+                                    integral_bsplines_without_periodic_point.size()},
+                            0);
+    Kokkos::deep_copy(
+            integral_bsplines_mirror,
+            integral_bsplines_without_periodic_point.allocation_kokkos_view());
+    matrix->solve(integral_bsplines_mirror_with_additional_allocation, true);
+    Kokkos::deep_copy(
+            integral_bsplines_without_periodic_point.allocation_kokkos_view(),
+            integral_bsplines_mirror);
+
+    ddc::Chunk<
+            double,
+            ddc::DiscreteDomain<InterpolationDDim>,
+            ddc::KokkosAllocator<double, MemorySpace>>
+            coefficients(domain);
+
+    Kokkos::deep_copy(
+            coefficients.allocation_kokkos_view(),
+            integral_bsplines_without_periodic_point.allocation_kokkos_view());
+
+    return coefficients;
 }
 } // namespace ddc
--- a/include/ddc/kernels/splines/spline_builder.hpp
+++ b/include/ddc/kernels/splines/spline_builder.hpp
@@ -959,9 +959,16 @@ SplineBuilder<
             ddc::KokkosAllocator<double, MemorySpace>>
             coefficients(domain);
 
+    // Coefficients of quadrature in integral_bsplines (values which would always be multiplied
+    // by f'(x)=0 are removed
+    ddc::DiscreteDomain<bsplines_type> slice =
+            spline_domain()
+                      .remove(ddc::DiscreteVector<bsplines_type> {s_nbc_xmin},
+                              ddc::DiscreteVector<bsplines_type> {s_nbc_xmax});
+
     Kokkos::deep_copy(
             coefficients.allocation_kokkos_view(),
-            integral_bsplines_without_periodic_point.allocation_kokkos_view());
+            integral_bsplines_without_periodic_point[slice].allocation_kokkos_view());
 
     return coefficients;
 }


include/ddc/kernels/splines/spline_builder.hpp
```diff
@@ -893,4 +928,115 @@ operator()(
...
+        BSplines,
+        InterpolationDDim,
+        BcLower,
+        BcUpper,
+        Solver,
+        IDimX...>::quadrature_coefficients() const
+{
+    // Compute integrals of bsplines
+    ddc::Chunk<double, ddc::DiscreteDomain<bsplines_type>> integral_bsplines(spline_domain());
+    ddc::discrete_space<bsplines_type>().integrals(integral_bsplines.span_view());
```
EmilyBourne, line 968: Why is this calculated on CPU even if the ExecSpace/MemorySpace of this class is GPU?

include/ddc/kernels/splines/spline_builder.hpp
```diff
@@ -893,4 +928,115 @@ operator()(
...
+
+    // Remove last cell in the periodic case
+    ddc::ChunkSpan integral_bsplines_without_periodic_point
+            = integral_bsplines.span_view()[ddc::DiscreteDomain<bsplines_type>(
+                    ddc::DiscreteElement<bsplines_type>(0),
+                    ddc::DiscreteVector<bsplines_type>(matrix->size()))];
+
+    // Allocate mirror with additional rows (cf. SplinesLinearProblem3x3Blocks documentation)
+    Kokkos::View<double**, Kokkos::LayoutRight, Kokkos::DefaultHostExecutionSpace>
```
EmilyBourne, line 977: Why `Kokkos::DefaultHostExecutionSpace`?

include/ddc/kernels/splines/spline_builder.hpp
```diff
@@ -893,4 +928,115 @@ operator()(
...
+                            std::pair<std::size_t, std::size_t> {
+                                    0,
+                                    integral_bsplines_without_periodic_point.size()},
+                            0);
+
+    // Solve matrix equation A^t*X=integral_bsplines
+    Kokkos::deep_copy(
+            integral_bsplines_mirror,
+            integral_bsplines_without_periodic_point.allocation_kokkos_view());
+    matrix->solve(integral_bsplines_mirror_with_additional_allocation, true);
```
EmilyBourne, line 996: If the matrix is stored on GPU won't it be a problem that integral_bsplines_mirror_with_additional_allocation is on CPU?

tests/splines/non_periodic_spline_builder.cpp
```diff
@@ -82,9 +82,8 @@ TEST(NonPeriodicSplineBuilderTest, Identity)
 #elif defined(BSPLINES_TYPE_NON_UNIFORM)
         DVectX constexpr npoints(ncells + 1);
         std::vector<CoordX> breaks(npoints);
-        double dx = (xN - x0) / ncells;
         for (int i(0); i < npoints; ++i) {
-            breaks[i] = CoordX(x0 + i * dx);
+            breaks[i] = CoordX(x0 + i * (xN - x0) / ncells);
```
EmilyBourne, line 86: Why this change? I find it more readable before. This change also changes the order of the operations. Is this intentional?

EmilyBourne (changes requested):

tests/splines/non_periodic_spline_builder.cpp
blegouix: Yeah I wanted to use `dx`  here: https://github.com/CExA-project/ddc/pull/560/files#diff-566ae334c766db6d714d50e38d0c0057788a2a11a014c079f4f6a4307bfdd2f0R202

include/ddc/kernels/splines/spline_builder.hpp
blegouix: Everything is computed on CPU (because we work on a single thread (single rhs) then copied on GPU here:

https://github.com/CExA-project/ddc/pull/560/files#diff-a120a1119000f5d36cb8a669f09e5b36ac36fdea28981ae6ae5f034f60e3801dR1016

(I said the contrary in the Gysela MR but this was a mistake).

include/ddc/kernels/splines/spline_builder.hpp
blegouix: You are correct, I was not attentive when doing that, thx

tests/splines/non_periodic_spline_builder.cpp
EmilyBourne: Does that cause a conflict? Doesn't the `{}` at lines 79 and 90 avoid any conflict with that line?

include/ddc/kernels/splines/spline_builder.hpp
blegouix: Now it performs the computation on the ExecSpace but take an optional template argument OutMemorySpace, such that the computation may be performed and CPU but the coefficients are returned on GPU.

tests/splines/non_periodic_spline_builder.cpp
EmilyBourne: Don't agree that this is resolved. I saw your change [d99d4af](https://github.com/CExA-project/ddc/pull/560/commits/d99d4af8f43eba924a3d14ef74b0f846946ac40f) but don't understand why the later `dx` was renamed. I would still like an answer to my last question

include/ddc/kernels/splines/spline_builder.hpp
EmilyBourne: This looks better :+1:

EmilyBourne (dismissed):

tests/splines/non_periodic_spline_builder.cpp
blegouix: I did not see your message. Indeed the `{}`, avoid conflict, I did not noticed it

EmilyBourne:
ok for me when doc is fixed

blegouix:
don't understand what is the problem with doc. Maybe default template parameter conficts with doxygen somehow ?

include/ddc/kernels/splines/spline_builder.hpp
```diff
@@ -890,4 +929,125 @@ operator()(
...
+template <
+        class ExecSpace,
+        class MemorySpace,
+        class BSplines,
+        class InterpolationDDim,
+        ddc::BoundCond BcLower,
+        ddc::BoundCond BcUpper,
+        SplineSolver Solver,
+        class... IDimX>
+template <class OutMemorySpace>
```
EmilyBourne, line 942:
```suggestion
template <class OutMemorySpace>
template <
        class ExecSpace,
        class MemorySpace,
        class BSplines,
        class InterpolationDDim,
        ddc::BoundCond BcLower,
        ddc::BoundCond BcUpper,
        SplineSolver Solver,
        class... IDimX>
```
seems to fix doxygen issue

blegouix: Thx, I did not know that

include/ddc/kernels/splines/spline_builder.hpp
```diff
@@ -890,4 +929,125 @@ operator()(
...
+template <
+        class ExecSpace,
+        class MemorySpace,
+        class BSplines,
+        class InterpolationDDim,
+        ddc::BoundCond BcLower,
+        ddc::BoundCond BcUpper,
+        SplineSolver Solver,
+        class... IDimX>
```
tpadioleau, line 942:
```suggestion
template <
        class ExecSpace,
        class MemorySpace,
        class BSplines,
        class InterpolationDDim,
        ddc::BoundCond BcLower,
        ddc::BoundCond BcUpper,
        SplineSolver Solver,
        class... IDimX>
template <class OutMemorySpace>
```

include/ddc/kernels/splines/spline_builder.hpp
blegouix: I guess your fix passes doxygen requirement but does not compile ? I should revert it ?

include/ddc/kernels/splines/spline_builder.hpp
```diff
@@ -890,4 +929,125 @@ operator()(
...
+        BcUpper,
+        Solver,
+        IDimX...>::quadrature_coefficients() const
+{
+    // Compute integrals of bsplines
+    ddc::Chunk<
+            double,
+            ddc::DiscreteDomain<bsplines_type>,
+            ddc::KokkosAllocator<double, MemorySpace>>
+            integral_bsplines(spline_domain());
```
tpadioleau, line 973:
```suggestion
    ddc::Chunk integral_bsplines(spline_domain(), ddc::KokkosAllocator<double, MemorySpace>());
```

include/ddc/kernels/splines/spline_builder.hpp
```diff
@@ -890,4 +929,125 @@ operator()(
...
+    ddc::discrete_space<bsplines_type>().integrals(integral_bsplines.span_view());
+
+    // Remove last cell in the periodic case
+    ddc::ChunkSpan integral_bsplines_without_periodic_point
+            = integral_bsplines.span_view()[ddc::DiscreteDomain<bsplines_type>(
+                    ddc::DiscreteElement<bsplines_type>(0),
+                    ddc::DiscreteVector<bsplines_type>(matrix->size()))];
```
tpadioleau, line 980:
```suggestion
    ddc::ChunkSpan integral_bsplines_without_periodic_point
            = integral_bsplines[ddc::DiscreteDomain<bsplines_type>(
                    ddc::DiscreteElement<bsplines_type>(0),
                    ddc::DiscreteVector<bsplines_type>(matrix->size()))];
```

include/ddc/kernels/splines/spline_builder.hpp
EmilyBourne: Ah ^^. I was hoping that was the merge.

tpadioleau (changes requested):

include/ddc/kernels/splines/spline_builder.hpp
EmilyBourne: Any idea how to get doxygen to recognise it when it's this way round?

tpadioleau: Could it be related to the default value of the template parameter ?

EmilyBourne: It doesn't seem to be. Adding the default value doesn't make a difference. Maybe the definitions below the class should be in a `/// \cond` region?

EmilyBourne (approved):

include/ddc/kernels/splines/spline_builder.hpp
```diff
@@ -890,4 +931,118 @@ operator()(
...
+{
+    // Compute integrals of bsplines
+    ddc::Chunk integral_bsplines(spline_domain(), ddc::KokkosAllocator<double, MemorySpace>());
+    ddc::discrete_space<bsplines_type>().integrals(integral_bsplines.span_view());
+
+    // Remove last cell in the periodic case
+    ddc::ChunkSpan integral_bsplines_without_periodic_point
+            = integral_bsplines[ddc::DiscreteDomain<bsplines_type>(
+                    ddc::DiscreteElement<bsplines_type>(0),
+                    ddc::DiscreteVector<bsplines_type>(matrix->size()))];
```
tpadioleau, line 978: Can you explain a bit ? It is not clear to me reading the code that it does actually remove periodic point. `spline_domain()` is basically `full_domain()` of size `ncells() + degree()`. On the contrary `matrix->size()` equals `nbasis()` that is `ncells() + !is_periodic() * degree()`. I don't see the link with your comment "Remove last cell in the periodic case"

include/ddc/kernels/splines/spline_builder.hpp
```diff
@@ -890,4 +931,118 @@ operator()(
...
+            ddc::DiscreteDomain<deriv_type>(
+                    ddc::DiscreteElement<deriv_type>(1),
+                    ddc::DiscreteVector<deriv_type>(s_nbc_xmax)),
+            ddc::KokkosAllocator<double, OutMemorySpace>());
+    Kokkos::deep_copy(
+            coefficients_derivs_xmin.allocation_kokkos_view(),
+            integral_bsplines_without_periodic_point[ddc::DiscreteDomain<bsplines_type>(
+                                                             ddc::DiscreteElement<bsplines_type>(0),
+                                                             ddc::DiscreteVector<bsplines_type>(
+                                                                     s_nbc_xmin))]
```
tpadioleau, line 1026: Could be also a `take_first` call here ?

tpadioleau (commented):
Again we must minimize as much as possible the creation of `DiscreteElement` from raw integers, it is equivalent of calling `uid()`

include/ddc/kernels/splines/spline_builder.hpp
blegouix: Actually it is a mistake to speak about "periodic cell". The explanation is the documentation of `full_domain()` function.

include/ddc/kernels/splines/spline_builder.hpp
tpadioleau: There was a mismatch of templates between declaration and definition.

tests/splines/non_periodic_spline_builder.cpp
```diff
@@ -179,6 +179,61 @@ TEST(NonPeriodicSplineBuilderTest, Identity)
...
+            Kokkos::DefaultHostExecutionSpace(),
+            quadrature_coefficients_derivs_xmax.domain(),
+            0.0,
+            ddc::reducer::sum<double>(),
+            [&](ddc::DiscreteElement<ddc::Deriv<typename IDimX::continuous_dimension_type>> const
+                        ix) {
+                ddc::Coordinate<DimX> const dx
+                        = ddc::distance_at_left(interpolation_domain.back() - 1);
+                return quadrature_coefficients_derivs_xmax(ix)
+                       * (*deriv_r)(ix)*ddc::detail::ipow(dx, ix.uid());
```
tpadioleau, line 230: @EmilyBourne the term `ddc::detail::ipow(dx, ix.uid())` will always appear on the user side. Shall it be directly included in `quadrature_coefficients_derivs_xmax` ? Are there situations where you only need the current `quadrature_coefficients_derivs_xmax` ?

blegouix: Yeah good question, I would prefer to put `* (*deriv_r)(ix)*ddc::detail::ipow(dx, ix.uid())` inside `quadrature_coefficients_derivs_xmax(ix)`

EmilyBourne: I agree that this should appear in `quadrature_coefficients_derivs_xmax`. This factor is included to improve the condition number of the matrix. Once the matrix equation is solved there is no reason to keep this trick

blegouix: [38c2c0c](https://github.com/CExA-project/ddc/pull/560/commits/38c2c0cae61f1a77ecaf3fc453dbea451edf06d6)

PR #561, opened 2024-07-26
Fix create_mirror_view

blegouix:


EmilyBourne (changes requested):
This missing return was caught by Gysela's tests but wasn't caught by DDC's tests. This probably means that there is a test missing for this function. Could you add one to prevent this error reoccuring

blegouix:
> This missing return was caught by Gysela's tests but wasn't caught by DDC's tests. This probably means that there is a test missing for this function. Could you add one to prevent this error reoccuring

I'm sorry but it becomes very hard to add new things to my TODO list ^^. I don't see any test for `create_mirror`s, so I prefer to open an issue.

EmilyBourne (approved):

tpadioleau (approved):
Yes my bad

PR #594, opened 2024-08-02
Add template variable to retrieve the number of types in a TypeSeq

tpadioleau:


EmilyBourne (dismissed):

include/ddc/detail/type_seq.hpp
```diff
@@ -161,6 +161,12 @@ struct ToTypeSeq;
 
 } // namespace detail
 
+template <class TypeSeq>
+constexpr std::size_t type_seq_size_v = std::numeric_limits<std::size_t>::max();
```
EmilyBourne, line 165: Can we add a short docstring?

tpadioleau:
Do you need to rely on the value `std::numeric_limits<std::size_t>::max()` for non `TypeSeq` types ? It might be a detail of implementation in the future.

include/ddc/detail/type_seq.hpp
tpadioleau: This line is kind of a detail of implementation to me. I may change the way it is defined later to only keep the relevant part in the public namespace.

EmilyBourne: I agree. I just thought that the docstring had to be on the most general case for it to appear correctly in the documentation

EmilyBourne:
> Do you need to rely on the value `std::numeric_limits<std::size_t>::max()` for non `TypeSeq` types ? It might be a detail of implementation in the future.

No not at all. This can be whatever you prefer

include/ddc/detail/type_seq.hpp
tpadioleau: Can be, I need to move `type_seq.hpp` out of the detail directory also to make it appear on the website.

EmilyBourne: Ok. In that case it is not critical, especially for this case where the meaning is obvious, but I don't see any harm in adding docs now so they are ready if/when you expose type seq

tpadioleau: Sorry I thought it was on the `TypeSeq` specialization but I did not save the file.

EmilyBourne (approved):

PR #338, opened 2024-03-12
Reuse compliant license

jbigot:
Addition of a [reuse](https://reuse.software/) compliant-licensing info

PR #601, opened 2024-08-14
Fix the evaluation of bsplines infinitesimally close to the boundary

### LINKED ISSUE #599: Rounding error introduced by #519
PR #519 introduced tolerances on the evaluation of bsplines in order to avoid issues due to rounding errors near the domain boundaries. However the PR didn't change the evaluation algorithm. As a result if the rounding error occurs near the upper bound the value of the returned index is too large. This can lead to chunks being evaluated outside their domain.

## Minimum reproducer

```cpp
#include <iostream>

#include <ddc/ddc.hpp>
#include <ddc/kernels/splines.hpp>

struct Vx
{
    static bool constexpr PERIODIC = false;
};

struct GridVx : ddc::UniformPointSampling<Vx>
{
};

struct BSplines : ddc::UniformBSplines<Vx, 3>
{
};

int main(int argc, char** argv)
{
    Kokkos::ScopeGuard kokkos_scope(argc, argv);
    ddc::ScopeGuard ddc_scope(argc, argv);

    ddc::Coordinate<Vx> min(-6.0);
    ddc::Coordinate<Vx> max(6.0);
    ddc::DiscreteVector<GridVx> ncells(34);
    ddc::init_discrete_space<BSplines>(min, max, ncells);

    using InterpPointInitMethod = ddc::GrevilleInterpolationPoints<BSplines, ddc::BoundCond::HERMITE, ddc::BoundCond::HERMITE>;
    ddc::init_discrete_space<GridVx>(InterpPointInitMethod::get_sampling<GridVx>());
    ddc::DiscreteDomain<GridVx> idx_range(InterpPointInitMethod::get_domain<GridVx>());

    // Shows that ddc::coordinate(idx_range.back()) > max
    std::cout << ddc::coordinate(idx_range.back()) << " " << max << " "
              << (ddc::coordinate(idx_range.back()) - max) << std::endl;

    using SplineVxBuilder = ddc::SplineBuilder<
            Kokkos::DefaultExecutionSpace,
            Kokkos::DefaultExecutionSpace::memory_space,
            BSplines,
            GridVx,
            ddc::BoundCond::HERMITE,
            ddc::BoundCond::HERMITE,
            ddc::SplineSolver::LAPACK,
            GridVx>;

    SplineVxBuilder builder(idx_range);

    ddc::DiscreteDomain<BSplines> spline_idx_range = builder.spline_domain();

    std::array<double, 4> basis_vals_ptr;
    ddc::DSpan1D basis_vals(basis_vals_ptr.data(), 4);

    ddc::DiscreteElement<BSplines> idx
            = ddc::discrete_space<BSplines>().eval_basis(basis_vals, ddc::coordinate(idx_range.back()));

    std::cout << "Iteration range : " << idx << " " << idx+3 << std::endl;
    std::cout << "Last valid index : " << spline_idx_range.back() << std::endl;

    return EXIT_SUCCESS;
}
```

Output:
```
(6) (6) (3.55271e-15)
Iteration range : (34) (37)
Last valid index : (36)
```
The "iteration range" refers to the range of bsplines which should be referenced by the 4 bspline values in the array.
Iteration over this range occurs here:
https://github.com/CExA-project/ddc/blob/93b822f713fa326432aec9376c347a7f5b63df59/include/ddc/kernels/splines/spline_evaluator.hpp#L450-L452
(not critical as the too large coordinate leads to extrapolation conditions being evaluated instead)

and here:
https://github.com/CExA-project/ddc/blob/93b822f713fa326432aec9376c347a7f5b63df59/include/ddc/kernels/splines/constant_extrapolation_rule.hpp#L71-L73
(can cause unexpected failures in simulations)

cc: @Geoflow

EmilyBourne:
Fixes #599

### DIFF
--- a/tests/splines/bsplines.cpp
+++ b/tests/splines/bsplines.cpp
@@ -142,8 +142,7 @@ TYPED_TEST(BSplinesFixture, Rounding_NonUniform)
     }
     ddc::init_discrete_space<BSplinesX>(breaks);
 
-    ddc::DiscreteVector<BSplinesX> nbasis(ddc::discrete_space<BSplinesX>().nbasis());
-    ddc::DiscreteDomain<BSplinesX> bspl_basis_domain
+    ddc::DiscreteDomain<BSplinesX> bspl_full_domain
             = ddc::discrete_space<BSplinesX>().full_domain();
 
     std::array<double, degree + 1> values_ptr;
@@ -153,12 +152,12 @@ TYPED_TEST(BSplinesFixture, Rounding_NonUniform)
     CoordX test_point_min(xmin - std::numeric_limits<double>::epsilon());
     ddc::DiscreteElement<BSplinesX> front_idx
             = ddc::discrete_space<BSplinesX>().eval_basis(values, test_point_min);
-    EXPECT_EQ(front_idx, bspl_basis_domain.front());
+    EXPECT_EQ(front_idx, bspl_full_domain.front());
 
     CoordX test_point_max(xmax + std::numeric_limits<double>::epsilon());
     ddc::DiscreteElement<BSplinesX> back_idx
             = ddc::discrete_space<BSplinesX>().eval_basis(values, test_point_max);
-    EXPECT_EQ(back_idx, bspl_basis_domain.back() - BSplinesX::degree());
+    EXPECT_EQ(back_idx, bspl_full_domain.back() - BSplinesX::degree());
 }
 
 TYPED_TEST(BSplinesFixture, Rounding_Uniform)
@@ -172,8 +171,7 @@ TYPED_TEST(BSplinesFixture, Rounding_Uniform)
     static constexpr std::size_t ncells = TestFixture::ncells;
     ddc::init_discrete_space<BSplinesX>(xmin, xmax, ncells);
 
-    ddc::DiscreteVector<BSplinesX> nbasis(ddc::discrete_space<BSplinesX>().nbasis());
-    ddc::DiscreteDomain<BSplinesX> bspl_basis_domain
+    ddc::DiscreteDomain<BSplinesX> bspl_full_domain
             = ddc::discrete_space<BSplinesX>().full_domain();
 
     std::array<double, degree + 1> values_ptr;
@@ -183,10 +181,10 @@ TYPED_TEST(BSplinesFixture, Rounding_Uniform)
     CoordX test_point_min(xmin - std::numeric_limits<double>::epsilon());
     ddc::DiscreteElement<BSplinesX> front_idx
             = ddc::discrete_space<BSplinesX>().eval_basis(values, test_point_min);
-    EXPECT_EQ(front_idx, bspl_basis_domain.front());
+    EXPECT_EQ(front_idx, bspl_full_domain.front());
 
     CoordX test_point_max(xmax + std::numeric_limits<double>::epsilon());
     ddc::DiscreteElement<BSplinesX> back_idx
             = ddc::discrete_space<BSplinesX>().eval_basis(values, test_point_max);
-    EXPECT_EQ(back_idx, bspl_basis_domain.back() - BSplinesX::degree());
+    EXPECT_EQ(back_idx, bspl_full_domain.back() - BSplinesX::degree());
 }
--- a/tests/splines/bsplines.cpp
+++ b/tests/splines/bsplines.cpp
@@ -126,7 +126,7 @@ TEST(KnotDiscreteDimension, Type)
                  ddc::NonUniformBsplinesKnots<DDim2>>));
 }
 
-TEST(BSplinesFixture, Rounding_NonUniform)
+TYPED_TEST(BSplinesFixture, Rounding_NonUniform)
 {
     std::size_t constexpr degree = TestFixture::spline_degree;
     using DimX = typename TestFixture::DimX;
@@ -142,7 +142,9 @@ TEST(BSplinesFixture, Rounding_NonUniform)
     }
     ddc::init_discrete_space<BSplinesX>(breaks);
 
-    ddc::DiscreteDomain<BSplinesX> bspl_basis_domain = ddc::discrete_space<BSplinesX>().nbasis();
+    ddc::DiscreteVector<BSplinesX> nbasis(ddc::discrete_space<BSplinesX>().nbasis());
+    ddc::DiscreteDomain<BSplinesX> bspl_basis_domain
+            = ddc::discrete_space<BSplinesX>().full_domain().take_first(nbasis);
 
     std::array<double, degree + 1> values_ptr;
     std::experimental::mdspan<double, std::experimental::extents<std::size_t, degree + 1>> const
@@ -161,7 +163,7 @@ TEST(BSplinesFixture, Rounding_NonUniform)
     EXPECT_LE(back_idx, bspl_basis_domain.back());
 }
 
-TEST(BSplinesFixture, Rounding_Uniform)
+TYPED_TEST(BSplinesFixture, Rounding_Uniform)
 {
     std::size_t constexpr degree = TestFixture::spline_degree;
     using DimX = typename TestFixture::DimX;
@@ -172,7 +174,9 @@ TEST(BSplinesFixture, Rounding_Uniform)
     static constexpr std::size_t ncells = TestFixture::ncells;
     ddc::init_discrete_space<BSplinesX>(xmin, xmax, ncells);
 
-    ddc::DiscreteDomain<BSplinesX> bspl_basis_domain = ddc::discrete_space<BSplinesX>().nbasis();
+    ddc::DiscreteVector<BSplinesX> nbasis(ddc::discrete_space<BSplinesX>().nbasis());
+    ddc::DiscreteDomain<BSplinesX> bspl_basis_domain
+            = ddc::discrete_space<BSplinesX>().full_domain().take_first(nbasis);
 
     std::array<double, degree + 1> values_ptr;
     std::experimental::mdspan<double, std::experimental::extents<std::size_t, degree + 1>> const
--- a/tests/splines/bsplines.cpp
+++ b/tests/splines/bsplines.cpp
@@ -153,14 +153,12 @@ TYPED_TEST(BSplinesFixture, Rounding_NonUniform)
     CoordX test_point_min(xmin - std::numeric_limits<double>::epsilon());
     ddc::DiscreteElement<BSplinesX> front_idx
             = ddc::discrete_space<BSplinesX>().eval_basis(values, test_point_min);
-    EXPECT_EQUAL(front_idx, bspl_basis_domain.front());
-    EXPECT_EQUAL(front_idx, bspl_basis_domain.back());
+    EXPECT_EQ(front_idx, bspl_basis_domain.front());
 
     CoordX test_point_max(xmax + std::numeric_limits<double>::epsilon());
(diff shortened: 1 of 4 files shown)


tests/splines/bsplines.cpp
```diff
@@ -124,3 +125,72 @@ TEST(KnotDiscreteDimension, Type)
...
+
+    std::array<double, degree + 1> values_ptr;
+    std::experimental::mdspan<double, std::experimental::extents<std::size_t, degree + 1>> const
+            values(values_ptr.data());
+
+    CoordX test_point_min(xmin - DBL_EPSILON);
+    ddc::DiscreteElement<BSplinesX> front_idx
+            = ddc::discrete_space<BSplinesX>().eval_basis(values, test_point_min);
+    EXPECT_GE(front_idx, bspl_basis_domain.front());
+    EXPECT_LE(front_idx, bspl_basis_domain.back());
```
tpadioleau, line 157: Can we get equality ?

tpadioleau (changes requested):
Thanks for the fix. I have minor comments.

tpadioleau:
Waiting for the main branch to publish the docker images.

tests/splines/bsplines.cpp
```diff
@@ -124,3 +125,72 @@ TEST(KnotDiscreteDimension, Type)
...
+
+    std::array<double, degree + 1> values_ptr;
+    std::experimental::mdspan<double, std::experimental::extents<std::size_t, degree + 1>> const
+            values(values_ptr.data());
+
+    CoordX test_point_min(xmin - std::numeric_limits<double>::epsilon());
+    ddc::DiscreteElement<BSplinesX> front_idx
+            = ddc::discrete_space<BSplinesX>().eval_basis(values, test_point_min);
+    EXPECT_EQUAL(front_idx, bspl_basis_domain.front());
+    EXPECT_EQUAL(front_idx, bspl_basis_domain.back());
```
tpadioleau, line 157:
```suggestion
    EXPECT_EQ(front_idx, bspl_basis_domain.front());
```

tests/splines/bsplines.cpp
```diff
@@ -124,3 +125,72 @@ TEST(KnotDiscreteDimension, Type)
...
+
+    std::array<double, degree + 1> values_ptr;
+    std::experimental::mdspan<double, std::experimental::extents<std::size_t, degree + 1>> const
+            values(values_ptr.data());
+
+    CoordX test_point_min(xmin - std::numeric_limits<double>::epsilon());
+    ddc::DiscreteElement<BSplinesX> front_idx
+            = ddc::discrete_space<BSplinesX>().eval_basis(values, test_point_min);
+    EXPECT_GE(front_idx, bspl_basis_domain.front());
+    EXPECT_LE(front_idx, bspl_basis_domain.back());
```
tpadioleau, line 189:
```suggestion
    EXPECT_EQ(front_idx, bspl_basis_domain.front());
```

tpadioleau:
@EmilyBourne We must use `ncells` instead of `nbasis` right ?

EmilyBourne:
> @EmilyBourne We must use `ncells` instead of `nbasis` right ?

No. `nbasis` is correct. The returned index should be the index of one of the basis elements. There are nbasis possibilities. ncells is much smaller in the non-periodic case. The problem is that rounding errors can go both ways. The second to last basis reaches 0 at the boundary so it is also affected by rounding errors. So the correct index is either the last or the second to last one. This is probably why I was just checking that we were returning a basis element (previously we returned Idx(nbasis) which is out of bounds).

EmilyBourne:
@tpadioleau Do you prefer that I revert to checking that the index is in the acceptable domain or would you rather I check if it is one of the two plausible values?

tpadioleau:
> @tpadioleau Do you prefer that I revert to checking that the index is in the acceptable domain or would you rather I check if it is one of the two plausible values?

Well in my opinion the more precise the test the better

EmilyBourne:
> > > @EmilyBourne We must use `ncells` instead of `nbasis` right ?
> > 
> > 
> > No. `nbasis` is correct. The returned index should be the index of one of the basis elements. There are nbasis possibilities. ncells is much smaller in the non-periodic case. The problem is that rounding errors can go both ways. The second to last basis reaches 0 at the boundary so it is also affected by rounding errors. So the correct index is either the last or the second to last one. This is probably why I was just checking that we were returning a basis element (previously we returned Idx(nbasis) which is out of bounds).
> 
> I am surprised, shouldn't there be a difference of at most one between `back_idx` and `bspl_basis_domain.back()` ?

Not sure what your question is. "second to last" means `bspl_basis_domain.back()-1`

tpadioleau:
> > > > @EmilyBourne We must use `ncells` instead of `nbasis` right ?
> > > 
> > > 
> > > No. `nbasis` is correct. The returned index should be the index of one of the basis elements. There are nbasis possibilities. ncells is much smaller in the non-periodic case. The problem is that rounding errors can go both ways. The second to last basis reaches 0 at the boundary so it is also affected by rounding errors. So the correct index is either the last or the second to last one. This is probably why I was just checking that we were returning a basis element (previously we returned Idx(nbasis) which is out of bounds).
> > 
> > 
> > I am surprised, shouldn't there be a difference of at most one between `back_idx` and `bspl_basis_domain.back()` ?
> 
> Not sure what your question is. "second to last" means `bspl_basis_domain.back()-1`

The CI shows there is a difference that corresponds to the degree of the splines.

```
337/762 Test #338: BSplinesFixture.Rounding_NonUniform<std::tuple<std::integral_constant<unsigned long, 5ul>, std::integral_constant<unsigned long, 100ul>, std::integral_constant<bool, false> >> ...........***Failed    0.02 sec
Note: Google Test filter = BSplinesFixture/34.Rounding_NonUniform
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from BSplinesFixture/34, where TypeParam = std::tuple<std::integral_constant<unsigned long, 5ul>, std::integral_constant<unsigned long, 100ul>, std::integral_constant<bool, false> >
[ RUN      ] BSplinesFixture/34.Rounding_NonUniform
/src/tests/splines/bsplines.cpp:161: Failure
Expected equality of these values:
  back_idx
    Which is: (99)
  bspl_basis_domain.back()
    Which is: (104)
```

EmilyBourne:
> > > > > @EmilyBourne We must use `ncells` instead of `nbasis` right ?
> > > > 
> > > > 
> > > > No. `nbasis` is correct. The returned index should be the index of one of the basis elements. There are nbasis possibilities. ncells is much smaller in the non-periodic case. The problem is that rounding errors can go both ways. The second to last basis reaches 0 at the boundary so it is also affected by rounding errors. So the correct index is either the last or the second to last one. This is probably why I was just checking that we were returning a basis element (previously we returned Idx(nbasis) which is out of bounds).
> > > 
> > > 
> > > I am surprised, shouldn't there be a difference of at most one between `back_idx` and `bspl_basis_domain.back()` ?
> > 
> > 
> > Not sure what your question is. "second to last" means `bspl_basis_domain.back()-1`
> 
> The CI shows there is a difference that corresponds to the degree of the splines.
> 
> ```
> 337/762 Test #338: BSplinesFixture.Rounding_NonUniform<std::tuple<std::integral_constant<unsigned long, 5ul>, std::integral_constant<unsigned long, 100ul>, std::integral_constant<bool, false> >> ...........***Failed    0.02 sec
> Note: Google Test filter = BSplinesFixture/34.Rounding_NonUniform
> [==========] Running 1 test from 1 test suite.
> [----------] Global test environment set-up.
> [----------] 1 test from BSplinesFixture/34, where TypeParam = std::tuple<std::integral_constant<unsigned long, 5ul>, std::integral_constant<unsigned long, 100ul>, std::integral_constant<bool, false> >
> [ RUN      ] BSplinesFixture/34.Rounding_NonUniform
> /src/tests/splines/bsplines.cpp:161: Failure
> Expected equality of these values:
>   back_idx
>     Which is: (99)
>   bspl_basis_domain.back()
>     Which is: (104)
> ```

I have had a closer look. What I said earlier was incorrect. There is no rounding error because the point that I am testing here is out of bounds. The value that is returned should be a value `idx` such that the d+1 bspline elements `[idx, idx+degree]` can all be accessed. For a periodic domain this means that `idx == bspl_basis_domain.back()` as there are degree ghost elements. For a non-periodic domain `idx == bspl_basis_domain.back() - degree`. I have updated the test. It is now passing again

tests/splines/bsplines.cpp
```diff
@@ -124,3 +125,68 @@ TEST(KnotDiscreteDimension, Type)
...
+    static constexpr CoordX xmax = CoordX(0.2);
+    static constexpr std::size_t ncells = TestFixture::ncells;
+    std::vector<CoordX> breaks(ncells + 1);
+    double dx = (xmax - xmin) / ncells;
+    for (std::size_t i(0); i < ncells + 1; ++i) {
+        breaks[i] = CoordX(xmin + i * dx);
+    }
+    ddc::init_discrete_space<BSplinesX>(breaks);
+
+    ddc::DiscreteVector<BSplinesX> nbasis(ddc::discrete_space<BSplinesX>().nbasis());
```
tpadioleau, line 145:
```suggestion
```

tests/splines/bsplines.cpp
```diff
@@ -124,3 +125,68 @@ TEST(KnotDiscreteDimension, Type)
...
+    std::size_t constexpr degree = TestFixture::spline_degree;
+    using DimX = typename TestFixture::DimX;
+    using BSplinesX = typename TestFixture::UBSplinesX;
+    using CoordX = ddc::Coordinate<DimX>;
+    static constexpr CoordX xmin = CoordX(0.0);
+    static constexpr CoordX xmax = CoordX(0.2);
+    static constexpr std::size_t ncells = TestFixture::ncells;
+    ddc::init_discrete_space<BSplinesX>(xmin, xmax, ncells);
+
+    ddc::DiscreteVector<BSplinesX> nbasis(ddc::discrete_space<BSplinesX>().nbasis());
```
tpadioleau, line 175:
```suggestion
```

tpadioleau (changes requested):

tpadioleau:
> I have had a closer look. What I said earlier was incorrect. There is no rounding error because the point that I am testing here is out of bounds. The value that is returned should be a value `idx` such that the d+1 bspline elements `[idx, idx+degree]` can all be accessed. For a periodic domain this means that `idx == bspl_basis_domain.back()` as there are degree ghost elements.

But we also do `bspl_basis_domain.back() - BSplinesX::degree()` even for the periodic case ?

EmilyBourne:
> > I have had a closer look. What I said earlier was incorrect. There is no rounding error because the point that I am testing here is out of bounds. The value that is returned should be a value `idx` such that the d+1 bspline elements `[idx, idx+degree]` can all be accessed. For a periodic domain this means that `idx == bspl_basis_domain.back()` as there are degree ghost elements.
> 
> But we also do `bspl_basis_domain.back() - BSplinesX::degree()` even for the periodic case ?

Maybe `bspl_basis_domain` is badly named now. It is no longer restricted to the nbasis elements. It is the full domain

tpadioleau:
> For a periodic domain this means that idx == bspl_basis_domain.back() as there are degree ghost elements.

Did you mean `idx == bspl_basis_domain.back() - BSplinesX::degree()` ?

tpadioleau (approved):
Looks good to me, thanks

PR #607, opened 2024-08-29
Do not systematically pull images

tpadioleau:


.github/workflows/tests.yml
```diff
@@ -52,7 +52,6 @@ jobs:
     - name: Build
       run: |
         echo ${{ secrets.GITHUB_TOKEN }} | docker login ghcr.io -u ${{ github.actor }} --password-stdin || true
-        docker pull ghcr.io/cexa-project/ddc/${{matrix.image}}_${{matrix.backend}} || true
```
tpadioleau, line 55: Temporary workaround

PR #613, opened 2024-09-02
Install the minimum of dependencies and build ginkgo in release

tpadioleau:


docker/latest/Dockerfile
```diff
@@ -30,56 +32,79 @@ RUN chmod +x /bin/bash_run \
...
       && echo "/opt/rocm/lib64" >> /etc/ld.so.conf.d/rocm.conf \
       && ldconfig \
-;; esac \
+ ;; esac \
  && git clone -b v1.8.0 https://github.com/ginkgo-project/ginkgo.git \
  && cd ginkgo \
  && case "${BACKEND}" in \
     "cpu") \
-      cmake -S . -B build -DCMAKE_BUILD_TYPE=RelWithDebInfo -DGINKGO_BUILD_OMP=ON -DGINKGO_BUILD_TESTS=OFF -DGINKGO_BUILD_EXAMPLES=OFF -DGINKGO_BUILD_BENCHMARKS=OFF \
+      cmake -S . -B build -DCMAKE_BUILD_TYPE=Release -DGINKGO_BUILD_OMP=ON -DGINKGO_BUILD_TESTS=OFF -DGINKGO_BUILD_EXAMPLES=OFF -DGINKGO_BUILD_BENCHMARKS=OFF \
```
yasahi-hpc, line 99: I would like to install ginkgo during GitHub actions not inside a Dockerfile. If there is a reason, could you please add comments?

yasahi-hpc (changes requested):
As a general remark, I do not think it is a good idea to introduce conditional branches inside a Dockerfile for multiple backends. I would rather recommend to prepare multiple Dockerfiles for each backend. This way each Dockerfile is significantly easier to read and maintain.

docker/latest/Dockerfile
```diff
@@ -30,56 +32,79 @@ RUN chmod +x /bin/bash_run \
...
     "cpu") \
       apt-get install -y --no-install-recommends \
       clang \
       clang-tidy \
       clang-format \
  ;; "cuda") \
+      # libnvjitlink-12-6 seems to be a missing dependency of cusparse
       apt-get install -y --no-install-recommends \
-      nvidia-cuda-toolkit \
+      cuda-minimal-build-12-6 \
```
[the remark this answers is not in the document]
tpadioleau: Most cuda/hip dependencies are ginkgo dependencies. If it does not increase too much the size of the image we could dedicate a layer for that ?

I agree that mixing all the dependencies in a single place is difficult to maintain.

docker/latest/Dockerfile
tpadioleau: It takes a long time to compile, more than 30 min. So we compile it once and for all in the images and use it the CI.

tpadioleau:
> As a general remark, I do not think it is a good idea to introduce conditional branches inside a Dockerfile for multiple backends. I would rather recommend to prepare multiple Dockerfiles for each backend. This way each Dockerfile is significantly easier to read and maintain.

As the images tend to diverge it could be worth dedicate separate files. I also notice that modifying a single backend triggers the build of all backends which is a bit annoying. I am not against but it is not in the scope of the current PR.

docker/latest/Dockerfile
yasahi-hpc: Once we install ginkgo during actions, we can use it afterwards as well as other libraries. I guess other libraries are also compiled once and used in CI. Why you need special treatments for ginkgo?

tpadioleau: I don't think there is a special treatment for Ginkgo. Most of our dependencies are installed with distribution packages in the image. Actually I think the special treatment is for Kokkos/Kokkos-Kernels/benchmark/googletest. In my opinion these latter should belong to the image. Did I understand correctly ?

yasahi-hpc: Actually I just thought the treatment for `Kokkos/Kokkos-Kernels/benchmark/googletest` is usual. If you would like to install these libraries in the image as well as `ginkgo`, that is fine. My question is why you install `ginkgo` in a different way than others.

tpadioleau: I think it is kind of legacy. Before using Ginkgo, we were building Kokkos/benchmark/googletest when building DDC. It was fast enough. However when we introduced Ginkgo it was not possible to use this strategy as Ginkgo takes much more time than others. We then moved Ginkgo directly in the image but not the others. It should be ok to also move the remaining dependencies in the image.

yasahi-hpc: Sure, then it is OK to keep as is.
I prefer not to install too much inside an image, but if the installation takes too long, we have no choice.

tpadioleau:
@yasahi-hpc thank you for the review

PR #620, opened 2024-09-04
Make integrals a host function

tpadioleau:


tpadioleau:
tests pass on V100 gpu

PR #631, opened 2024-09-07
Remove switches and catch bad values

tpadioleau:
- turn switches into `if/else` blocks
- remove duplicate switches in `compute_block_sizes_*` functions

### DIFF
--- a/include/ddc/kernels/fft.hpp
+++ b/include/ddc/kernels/fft.hpp
@@ -6,6 +6,7 @@
 
 #include <array>
 #include <cstddef>
+#include <stdexcept>
 #include <type_traits>
 #include <utility>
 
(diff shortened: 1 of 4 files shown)


EmilyBourne:
Why do you prefer if/else over switches?

include/ddc/kernels/fft.hpp
```diff
@@ -564,36 +563,35 @@ void impl(
 
     if (kwargs.normalization != ddc::FFT_Normalization::OFF) {
         real_type_t<Tout> norm_coef = 1;
-        switch (kwargs.normalization) {
-        case ddc::FFT_Normalization::OFF:
-            break;
```
EmilyBourne, line 569: This case seems to have been removed. Is that intentional?

tpadioleau:
> Why do you prefer if/else over switches?

I find it old construct from the language, less obvious to implement correctly. We used to have a bug forgetting the break statement

EmilyBourne (approved):

PR #632, opened 2024-09-08
Use C++ random

tpadioleau:


tpadioleau:
@EmilyBourne Can you review ? I thought using `std::uniform_real_distribution dis(0, 1)` would work but it failed randomly. So I have just tried to make a C++ equivalent.

EmilyBourne (dismissed):

EmilyBourne:
Any idea why `uniform_real_distribution` was failing?

tpadioleau:
> Any idea why `uniform_real_distribution` was failing?

Seems to me it was only a question of threshold, you can have a look at it here https://github.com/CExA-project/ddc/actions/runs/10960987200/job/30438477342.

I can revert to using `uniform_real_distribution`, I was just not sure it was the original intent.

EmilyBourne:
> > Any idea why `uniform_real_distribution` was failing?
> 
> Seems to me it was only a question of threshold, you can have a look at it here https://github.com/CExA-project/ddc/actions/runs/10960987200/job/30438477342.
> 
> I can revert to using `uniform_real_distribution`, I was just not sure it was the original intent.

The original intent is simply to create a polynomial whose coefficients are between 0.0 and 1.0. Limiting the coefficients to this range stops the max norm from becoming excessively large.

The tolerance for the threshold should be an analytical bound so it is a bit odd if we are going over it. That said in the example that you linked to we are not over the threshold by much. The tolerance is:
```
std:: max(error_bounds.error_bound_on_deriv_12( dx<I1>(ncells), dx<I2>(ncells), s_degree, s_degree), 1e-12 * max_norm_diff12))
                                                    ^                                                             ^
                                            analytical bound                                            machine error handling
```
Maybe it would be sufficient to increase the `1e-12` to `1e-11`

EmilyBourne (approved):

PR #643, opened 2024-09-10
Add readability-avoid-const-params-in-decls check

tpadioleau:


tpadioleau:
superseded by #650 due to naming error

PR #660, opened 2024-10-04
Avoid redundant cmake logic

tpadioleau:
Factorize configuration of vendor dependencies into a single file. This avoids to repeat the logic for `EMBEDDED` and `AUTO` configuration options. Each macro is called `ddc_configure_*`.


science-enthusiast:
I looked at the commits. I don't know the details but the refactoring looks reasonable.

tpadioleau:
> I looked at the commits. I don't know the details but the refactoring looks reasonable.

thanks a lot

PR #662, opened 2024-10-13
Add google-explicit-constructor check

tpadioleau:


include/ddc/detail/tagged_vector.hpp
```diff
@@ -232,6 +232,7 @@ template <class ElementType, class Tag>
 class TaggedVectorConversionOperators<TaggedVector<ElementType, Tag>>
 {
 public:
+    // NOLINTBEGIN(google-explicit-constructor)
```
yasahi-hpc, line 235: Could you please comment why you need this?
This seems to be implicit conversion?

include/ddc/discrete_vector.hpp
```diff
@@ -217,6 +217,7 @@ template <class Tag>
 class DiscreteVectorConversionOperators<DiscreteVector<Tag>>
 {
 public:
+    // NOLINTBEGIN(google-explicit-constructor)
```
yasahi-hpc, line 220: Same question as before

include/ddc/kernels/splines/bsplines_non_uniform.hpp
```diff
@@ -117,7 +117,7 @@ class NonUniformBSplines : detail::NonUniformBSplinesBase
          *
          * @param breaks The std::initializer_list of the coordinates of break points.
          */
-        explicit Impl(std::initializer_list<ddc::Coordinate<CDim>> breaks)
+        Impl(std::initializer_list<ddc::Coordinate<CDim>> breaks)
```
yasahi-hpc, line 120: I thought we need an explicit here. Could you please comment why not needed?

include/ddc/non_uniform_point_sampling.hpp
```diff
@@ -64,7 +64,7 @@ class NonUniformPointSampling : detail::NonUniformPointSamplingBase
         Impl() = default;
 
         /// @brief Construct a `NonUniformPointSampling` using a brace-list, i.e. `NonUniformPointSampling mesh({0., 1.})`
-        explicit Impl(std::initializer_list<continuous_element_type> points)
+        Impl(std::initializer_list<continuous_element_type> points)
```
yasahi-hpc, line 67: Same as above

yasahi-hpc (approved):
Mostly good to me, but I have several questions. 
Also, it is appreciated that if you could have a quick summary of the aim of this PR.

include/ddc/detail/tagged_vector.hpp
tpadioleau: The check also work cast operators, see https://clang.llvm.org/extra/clang-tidy/checks/google/explicit-constructor.html

include/ddc/kernels/splines/bsplines_non_uniform.hpp
tpadioleau: This is also how the check works https://clang.llvm.org/extra/clang-tidy/checks/google/explicit-constructor.html. They advise implicit constructors with initializer lists. I think it is be able to write
```cpp
struct A
{
    A(std::initializer_list<int> l);
};

void foo(A const& a);

// using foo
foo({1, 2, 3});
```

tpadioleau:
> Mostly good to me, but I have several questions. 
> Also, it is appreciated that if you could have a quick summary of the aim of this PR.

Sorry i thought the title was enough. I did a lot of these checks recently.

include/ddc/detail/tagged_vector.hpp
yasahi-hpc: Sure

yasahi-hpc:
> > Mostly good to me, but I have several questions.
> > Also, it is appreciated that if you could have a quick summary of the aim of this PR.
> 
> Sorry i thought the title was enough. I did a lot of these checks recently.

Thank you for explanation. Now it is clear.

PR #663, opened 2024-10-13
Simplify fft full normalization

tpadioleau:
The "full" normalization can be simplified by defining a 1d normalization factor that would also avoid duplication
- [x] fix `double` implicit usage through `Kokkos::numbers::pi`
- [x] introduce two functions to help reading the normalization factors and the link between the forward and backward transformations

### DIFF
--- a/include/ddc/kernels/fft.hpp
+++ b/include/ddc/kernels/fft.hpp
@@ -179,14 +179,13 @@ void rescale(
 template <class DDim>
 Real forward_full_norm_coef(DiscreteDomain<DDim> const& ddom) noexcept
 {
-    return rlength(ddom) / Kokkos::sqrt(2 * Kokkos::numbers::pi)
-           / (ddc::get<DDim>(ddom.extents()) - 1);
+    return rlength(ddom) / Kokkos::sqrt(2 * Kokkos::numbers::pi) / (ddom.extents() - 1).value();
 }
 
 template <class DDim>
 Real backward_full_norm_coef(DiscreteDomain<DDim> const& ddom) noexcept
 {
-    return 1 / (forward_full_norm_coef(ddom) * ddc::get<DDim>(ddom.extents()));
+    return 1 / (forward_full_norm_coef(ddom) * ddom.extents().value());
 }
 
 /// @brief Core internal function to perform the FFT.
--- a/include/ddc/kernels/fft.hpp
+++ b/include/ddc/kernels/fft.hpp
@@ -176,19 +176,19 @@ void rescale(
             });
 }
 
-template <class DDimIn>
-Real forward_full_norm_coef(DiscreteDomain<DDimIn> const& ddom) noexcept
+template <class DDim>
+Real forward_full_norm_coef(DiscreteDomain<DDim> const& ddom) noexcept
 {
     return (coordinate(ddom.back()) - coordinate(ddom.front()))
-           / (ddc::get<DDimIn>(ddom.extents()) - 1) / Kokkos::sqrt(2 * Kokkos::numbers::pi);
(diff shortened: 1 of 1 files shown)


include/ddc/kernels/fft.hpp
```diff
@@ -176,6 +176,18 @@ void rescale(
             });
 }
 
+template <class DDim>
+Real forward_full_norm_coef(DiscreteDomain<DDim> const& ddom) noexcept
```
yasahi-hpc, line 180: I may not fully understand the return type. How do you ensure that `Real` is identical to `real_type_t<Tout>`?

yasahi-hpc (dismissed):
LGTM
Could you please comment about the consistency between `Real` and `real_type_t<Tout>`

include/ddc/kernels/fft.hpp
tpadioleau: So `Real` is the type that is used by discrete dimensions when they need to use floating points, for example the function coordinate uses that type. Should be a template parameter at some point, for now it is a global compile time parameter.

`real_type` is used to get the floating point type when passing an array to fft functions.

About the consistency, the user can mix them. I don't know how bad/good it is.

yasahi-hpc: Is there a way to derive a base floating point type from `DiscreteDomain<DDim>` and use it instead of `Real`. 
I expect that this function returns the coefficient in the same precision as `DiscreteDomain<DDim>`.
This would also make future changes easier

tpadioleau: I am not sure to see the advantage ? The precision of this function is tied to the precision of rlength which is Real.

The caller will cast if needed ? For now it is implicit, i can make it explicit to better identify it ?

yasahi-hpc: > I am not sure to see the advantage ? The precision of this function is tied to the precision of rlength which is Real.

Well, there is no advantage for the moment. Just to inform users that the precision of this function is tied to the precision of `rlength`. It is not very clear how to identify the precision of `rlength` is `Real`.

> The caller will cast if needed ? For now it is implicit, i can make it explicit to better identify it ?

I think it is better to cast explicitly if there is a change that `real_type_t<Tout>` is not identical to `Real`

yasahi-hpc, line 180:
```suggestion
template <class DDim>
auto forward_full_norm_coef(DiscreteDomain<DDim> const& ddom) noexcept -> DiscreteDomain<DDim>::discrete_element_type::value_type
```

I may just prefer `auto`. Maybe with a guidance?

tpadioleau: In DDC we try to avoid `auto` unless there is a `if constexpr` logic or an unreadable type.

That being said, i don't see how it helps to understand the return type of the function. I find it even less clear there will be an implicit conversion.

yasahi-hpc: Sure. Then, that is fine
My point is to avoid a global type and allow users to choose a precision

tpadioleau: They should already be able to pick the types they want, even mixing the types. I don't really know how to address the latter case. There are different solutions and i don't how to decide which one is the best. But this discussion goes out of scope of the current PR.

tpadioleau: What do you think of this new version ?

tpadioleau: done in 4c95c14e7f2db103b08875b6f35fa91506e84c6f

yasahi-hpc (approved):

tpadioleau:
thanks @yasahi-hpc

PR #673, opened 2024-11-01
Add missing DDCConfigVersion

tpadioleau:
I am thinking of using SameMinorVersion compatibility for major version 0

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -290,4 +290,14 @@ configure_package_config_file(
     INSTALL_DESTINATION lib/cmake/DDC
 )
 
-install(FILES ${CMAKE_CURRENT_BINARY_DIR}/DDCConfig.cmake DESTINATION lib/cmake/DDC)
+write_basic_package_version_file(
+    ${CMAKE_CURRENT_BINARY_DIR}/DDCConfigVersion.cmake
+    COMPATIBILITY SameMinorVersion
+)
+
+install(
+    FILES
+        ${CMAKE_CURRENT_BINARY_DIR}/DDCConfig.cmake
+        ${CMAKE_CURRENT_BINARY_DIR}/DDCConfigVersion.cmake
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -260,4 +260,14 @@ configure_package_config_file(
     INSTALL_DESTINATION lib/cmake/DDC
 )
 
-install(FILES ${CMAKE_CURRENT_BINARY_DIR}/DDCConfig.cmake DESTINATION lib/cmake/DDC)
+write_basic_package_version_file(
+    ${CMAKE_CURRENT_BINARY_DIR}/DDCConfigVersion.cmake
+    COMPATIBILITY SameMinorVersion
```
cedricchevalier19, line 265: Add a comment that you follow semantic versioning and that you are currently at version 0.

cedricchevalier19 (approved):
If version 0.x is really what you want, it looks good to me.

CMakeLists.txt
tpadioleau: Where should I add that comment ? Near the version declaration ? Isn't it implicit ?

cedricchevalier19: I thought to make clear why you use `SameMinorVersion` which is a bit surprising in a general context.

tpadioleau: Maybe I misunderstood the role of version 0, I thought the API was unstable between minor versions and thus I thought it would make sense to request compatibility between same minor versions ?

cedricchevalier19: Yes, you are true. I was just thinking about adding a comment that `SameMinorVersion` is used because you are in the `0.x` versions.

Most projects use `SameMajorVersion` because they are above `0`, so `SameMinorVersion` looks very restrictive when you do have context.

tpadioleau: I see I agree, I will add the comment, we will remove it as soon as we go to version 1

PR #675, opened 2024-11-01
Include headers relative to -I

tpadioleau:


include/ddc/detail/kokkos.hpp
```diff
@@ -11,7 +11,7 @@
 
 #include <Kokkos_Core.hpp>
 
-#include "macros.hpp"
+#include "ddc/detail/macros.hpp"
```
cedricchevalier19, line 14: Not sure this change is sound. Using `"macros.hpp"` will make the compiler look for a macros.hpp file in the current directory and not to browse the whole path list of includes.

cedricchevalier19: "source": https://blogs.stonesteps.ca/1/p/64

include/ddc/kernels/fft.hpp
```diff
@@ -8,11 +8,11 @@
 #include <type_traits>
 #include <utility>
 
-#include <ddc/ddc.hpp>
-
 #include <KokkosFFT.hpp>
 #include <Kokkos_Core.hpp>
 
+#include "ddc/ddc.hpp"
```
cedricchevalier19, line 14:
```suggestion
#include <ddc/ddc.hpp>
```

Here `ddc.hpp` is not in a `ddc` subdirectory of the current `kernels` directory.

include/ddc/kernels/splines.hpp
```diff
@@ -4,29 +4,30 @@
...
+#include "ddc/kernels/splines/splines_linear_problem.hpp"
+#include "ddc/kernels/splines/splines_linear_problem_2x2_blocks.hpp"
+#include "ddc/kernels/splines/splines_linear_problem_3x3_blocks.hpp"
+#include "ddc/kernels/splines/splines_linear_problem_band.hpp"
+#include "ddc/kernels/splines/splines_linear_problem_dense.hpp"
+#include "ddc/kernels/splines/splines_linear_problem_maker.hpp"
+#include "ddc/kernels/splines/splines_linear_problem_pds_band.hpp"
+#include "ddc/kernels/splines/splines_linear_problem_pds_tridiag.hpp"
+#include "ddc/kernels/splines/splines_linear_problem_sparse.hpp"
+#include "ddc/kernels/splines/view.hpp"
```
cedricchevalier19, line 33: If you go this path, I think `<>` is more appropriate.

include/ddc/kernels/splines/bsplines_non_uniform.hpp
```diff
@@ -9,9 +9,8 @@
 #include <memory>
 #include <vector>
 
-#include <ddc/ddc.hpp>
-
-#include "view.hpp"
+#include "ddc/ddc.hpp"
```
cedricchevalier19, line 12: I think this change is not necessary.

include/ddc/kernels/splines/bsplines_uniform.hpp
```diff
@@ -8,10 +8,9 @@
 #include <cassert>
 #include <memory>
 
-#include <ddc/ddc.hpp>
-
-#include "math_tools.hpp"
-#include "view.hpp"
+#include "ddc/ddc.hpp"
```
cedricchevalier19, line 11: Same here

cedricchevalier19 (commented):
Most of the changes are making things slightly worse. We probably should gather more documentation on how to deal with `#include` before committing to a choice.

tpadioleau:
Back to draft the time we get consensus

tpadioleau:
@cedricchevalier19 What do you think if:
1. we use double quotes for the internals of the core part 
2. for the components FFT, PDI, Splines use `<ddc/ddc.hpp>` and double quotes for their respective internals

Regarding 1., an issue is about the `experimental` directory that depends on the parent directory. An easy solution is to remove it as it is not used at all for now but it does not solve the issue on a longer term. We could also introduce a new directory `core` so that we would have:
```
ddc
|-core
|-ddc.hpp
|-details
|-experimental
|-kernels
  |-fft.hpp
  |-splines
  |-splines.hpp
```

Let me know if you have better ideas.

tpadioleau:
Other idea:
```
ddc/
├── CMakeLists.txt
├── include
│   └── ddc
│       ├── ddc.hpp
│       └── impl
│           ├── detail
│           └── some_header.hpp
└── tests
```

with components:
```
ddc_fft/
├── CMakeLists.txt
├── include
│   └── ddc
│       ├── fft.hpp
│       └── impl
│           ├── detail
│           └── some_header.hpp
└── tests
```

tpadioleau:
> I think this approach is easy to enforce if you are ok dealing with deep file hierarchies.

Are you aware of difficulties with deep file hierarchies ?

In a such a config there would still be one file that I don't really know how to handle which is `config.hpp` generated by cmake in the build directory. At the build stage it should be `#include <config.hpp>` because it would be found by specifying the path to the build directory. But once installed we could `#include "config.hpp"` because it gets copied with the rest of headers.

PR #680, opened 2024-11-08
Use cmake GNUInstallDirs variables

tpadioleau:


CMakeLists.txt
```diff
@@ -29,6 +29,10 @@ set(DDC_DEPENDENCY_POLICIES "AUTO" "EMBEDDED" "INSTALLED" "SUBPROJECT")
 
 include(CMakePackageConfigHelpers)
 include(CTest)
+include(GNUInstallDirs)
+
+set(DDC_INSTALL_CMAKEDIR ${CMAKE_INSTALL_LIBDIR}/cmake/${PROJECT_NAME})
+set(DDC_INSTALL_INCLUDEDIR ${CMAKE_INSTALL_INCLUDEDIR})
```
jbigot, line 35: I would set these as cache variables so that the user can change them if required

jbigot: In fact, only the 1st one. Not sure the second one is useful

jbigot (commented):
Looks great :)

CMakeLists.txt
tpadioleau: I don't know, so far I don't think I have ever seen a project exposing this kind of option. I only introduced the variable to avoid introducing mistakes in the path.

jbigot: I know at least PDI exposes this. 
Package that don't expose these are often a nightmare to package in Debian when they exist in multiple versions, because they are installed in a subdirectory and then linked to the main directory. This often breaks cmake it resolves file location because the files are not where it expects to find them.
Exposing this as an option makes it possible to specify that the cmake files are in a different location relative to the rest of the library.

PR #683, opened 2024-11-12
Fix Chunkspan::operator[] corner case

blegouix:
Calling `ChunkSpan::operator[]` on an element with same dimensions as the domain on which the `ChunkSpan` is defined should return a `ChunkSpan` on an empty domain `DiscreteDomain<>`. However it currently triggers the compilation error:

```
/home/cart3sianbear/SimiLie/vendor/ddc/include/ddc/chunk_span.hpp:210:31: fatal error: call to implicitly-deleted default constructor of 'ChunkCommon<double, DiscreteDomain<>, layout_right>'
  210 |     KOKKOS_FUNCTION constexpr ChunkSpan(
      |                               ^
/home/cart3sianbear/SimiLie/vendor/ddc/include/ddc/chunk_span.hpp:284:36: note: in instantiation of member function 'ddc::ChunkSpan<double, ddc::DiscreteDomain<>>::ChunkSpan' requested here
  284 |                             return ChunkSpan<
      |                                    ^
/home/cart3sianbear/SimiLie/vendor/ddc/tests/chunk_span.cpp:42:24: note: in instantiation of function template specialization 'ddc::ChunkSpan<double, ddc::DiscreteDomain<(anonymous namespace)::DDimX>>::operator[]<(anonymous namespace)::DDimX>' requested here
   42 |     auto subspan = span[DElemX(0)];
      |                        ^
/home/cart3sianbear/SimiLie/vendor/ddc/include/ddc/chunk_common.hpp:218:41: note: explicitly defaulted function was implicitly deleted here
  218 |     KOKKOS_DEFAULTED_FUNCTION constexpr ChunkCommon() = default;
      |                                         ^
/home/cart3sianbear/SimiLie/vendor/ddc/include/ddc/chunk_common.hpp:96:26: note: default constructor of 'ChunkCommon<double, ddc::DiscreteDomain<>, Kokkos::layout_right>' is implicitly deleted because field 'm_internal_mdspan' has no default constructor
   96 |     internal_mdspan_type m_internal_mdspan;
```

Not sure if the proposed patch is the best solutions though.


tpadioleau:
Thanks for the report and the fix, I can reproduce this compilation bug. I don't really understand what is happening looking at the error `fatal error: call to implicitly-deleted default constructor of`. I would like to understand it before merging. If you already have an explanation can you share ?

tpadioleau:
My understanding is that it is not allowed to default construct a mdspan of dynamic rank 0, see https://en.cppreference.com/w/cpp/container/mdspan/mdspan.

blegouix:
Okay! I have no more information about it

PR #686, opened 2024-11-12
Add functions to check if the provided interpolation points are valid

### LINKED ISSUE #659: User-defined spline interpolation points
When using splines DDC provides 2 options for defining interpolation points:
- Greville points
- Knots as interpolation points

Sometimes it is useful to be able to use user-defined points (as long as the provided points are compatible with the definition). For example if the points come from real world data or we want to test  a new way of defining points.

EmilyBourne:
Closes #659

### DIFF
--- a/include/ddc/kernels/splines/bsplines_uniform.hpp
+++ b/include/ddc/kernels/splines/bsplines_uniform.hpp
@@ -373,7 +373,7 @@ public:
         }
 
         KOKKOS_INLINE_FUNCTION discrete_element_type
-        eval_basis(DSpan1D values, ddc::Coordinate<CDim> const& x, std::size_t degree) const;
+        eval_basis(DSpan1D values, ddc::Coordinate<CDim> const& x, [[maybe_unused]] std::size_t degree) const;
 
         KOKKOS_INLINE_FUNCTION void get_icell_and_offset(
                 int& icell,
@@ -401,7 +401,7 @@ KOKKOS_INLINE_FUNCTION ddc::DiscreteElement<DDim> UniformBSplines<CDim, D>::
         Impl<DDim, MemorySpace>::eval_basis(
                 DSpan1D values,
                 ddc::Coordinate<CDim> const& x,
-                [[maybe_unused]] std::size_t const deg) const
+                [[maybe_unused]] std::size_t const degree) const
 {
     assert(values.size() == deg + 1);
 
--- a/include/ddc/kernels/splines/bsplines_uniform.hpp
+++ b/include/ddc/kernels/splines/bsplines_uniform.hpp
@@ -372,10 +372,8 @@ public:
             return 1.0 / ddc::step<knot_discrete_dimension_type>();
         }
 
-        KOKKOS_INLINE_FUNCTION discrete_element_type eval_basis(
-                DSpan1D values,
-                ddc::Coordinate<CDim> const& x,
-                [[maybe_unused]] std::size_t degree) const;
+        KOKKOS_INLINE_FUNCTION discrete_element_type
+        eval_basis(DSpan1D values, ddc::Coordinate<CDim> const& x, std::size_t degree) const;
 
         KOKKOS_INLINE_FUNCTION void get_icell_and_offset(
                 int& icell,
@@ -405,7 +403,7 @@ KOKKOS_INLINE_FUNCTION ddc::DiscreteElement<DDim> UniformBSplines<CDim, D>::
                 ddc::Coordinate<CDim> const& x,
                 [[maybe_unused]] std::size_t const degree) const
 {
-    assert(values.size() == deg + 1);
+    assert(values.size() == degree + 1);
 
     double offset;
     int jmin;
--- a/include/ddc/kernels/splines/bsplines_uniform.hpp
+++ b/include/ddc/kernels/splines/bsplines_uniform.hpp
@@ -372,8 +372,10 @@ public:
             return 1.0 / ddc::step<knot_discrete_dimension_type>();
         }
 
-        KOKKOS_INLINE_FUNCTION discrete_element_type
-        eval_basis(DSpan1D values, ddc::Coordinate<CDim> const& x, [[maybe_unused]] std::size_t degree) const;
+        KOKKOS_INLINE_FUNCTION discrete_element_type eval_basis(
+                DSpan1D values,
+                ddc::Coordinate<CDim> const& x,
+                [[maybe_unused]] std::size_t degree) const;
 
         KOKKOS_INLINE_FUNCTION void get_icell_and_offset(
                 int& icell,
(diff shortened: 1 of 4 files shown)


include/ddc/kernels/splines/spline_builder.hpp
tpadioleau, line 1: Sorry but I don't get the logic of the check. Are we checking the number of points in each cell ? If so shouldn't `current_cell_end_idx` increment in the loop as soon as `point > ddc::coordinate(current_cell_end_idx)` is true ?

include/ddc/kernels/splines/spline_builder.hpp
```diff
@@ -1064,4 +1069,80 @@ SplineBuilder<
...
+                "Incorrect number of points supplied to NonUniformInterpolationPoints. "
+                "(Received : "
+                + std::to_string(n_interp_points)
+                + ", expected : " + std::to_string(expected_npoints));
+    }
+    int n_points_in_cell = 0;
+    auto current_cell_end_idx = ddc::discrete_space<BSplines>().break_point_domain().front() + 1;
+    ddc::for_each(interpolation_domain(), [&](auto idx) {
+        ddc::Coordinate<continuous_dimension_type> point = ddc::coordinate(idx);
+        if (point == ddc::coordinate(current_cell_end_idx)) {
```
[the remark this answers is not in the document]
EmilyBourne: No I don't think so. The less than and greater than cases are both covered. Equal to only needs treating separately because it is in 2 cells. If we are at knot+eps then there is no ambiguity so no special treatment is necessary

include/ddc/kernels/splines/spline_builder.hpp
EmilyBourne: Hmm, I agree. Either I missed something while copy/pasting or the error output isn't correct in Gysela either. I don't see where `current_cell_idx` is updated

include/ddc/kernels/splines/spline_builder.hpp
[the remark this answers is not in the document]
tpadioleau: I see, so if I understand correctly, in this case you assign the point to one cell ?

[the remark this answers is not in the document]
EmilyBourne: Yes exactly. I can add a comment to explain this

include/ddc/kernels/splines/spline_builder.hpp
EmilyBourne: Fixed [26840d5](https://github.com/CExA-project/ddc/pull/686/commits/26840d5b05ee1f0c0bbe432803e1e0099d38fee5)

include/ddc/kernels/splines/spline_builder.hpp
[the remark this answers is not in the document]
EmilyBourne: Comments added : [3e3d35c](https://github.com/CExA-project/ddc/pull/686/commits/3e3d35c0edb3d2e5adee167dd5583e09c348d6c1)

EmilyBourne:
@tpadioleau Any idea what is wrong with the docs? There is nothing in the CI. If I try locally I get an error on a file that I didn't touch so maybe I have a different doxygen version

tpadioleau:
> @tpadioleau Any idea what is wrong with the docs? There is nothing in the CI. If I try locally I get an error on a file that I didn't touch so maybe I have a different doxygen version

I will have a look. Sorry i thought i had already made the CI print the doxygen.log file. I had this issue not so long ago.

EmilyBourne:
> > @tpadioleau Any idea what is wrong with the docs? There is nothing in the CI. If I try locally I get an error on a file that I didn't touch so maybe I have a different doxygen version
> 
> I will have a look. Sorry i thought i had already made the CI print the doxygen.log file. I had this issue not so long ago.

I think the `set -xe` that ensures an error is raised is causing failure before the output is printed

include/ddc/kernels/splines/bsplines_uniform.hpp
```diff
@@ -372,8 +372,10 @@ class UniformBSplines : detail::UniformBSplinesBase
             return 1.0 / ddc::step<knot_discrete_dimension_type>();
         }
 
-        KOKKOS_INLINE_FUNCTION discrete_element_type
-        eval_basis(DSpan1D values, ddc::Coordinate<CDim> const& x, std::size_t degree) const;
+        KOKKOS_INLINE_FUNCTION discrete_element_type eval_basis(
+                DSpan1D values,
+                ddc::Coordinate<CDim> const& x,
+                [[maybe_unused]] std::size_t degree) const;
```
tpadioleau, line 378: I suspect it comes from the `[[maybe_unused]]` here but first I would like to see if the CI can display the error.

tpadioleau:
It worked, see https://github.com/CExA-project/ddc/actions/runs/12067535482/job/33650784704. In the end it is the same issue as the `tests` workflow.

tpadioleau:
Shall we call the function in the constructor ?

tpadioleau:
I am trying to implement tests that would trigger one of the exceptions using the SplineBuilder constructor.

I first tried with a too short `interpolation_domain` but it failed in `compute_offset` before `check_valid_grid`. There is an out-of-bound access inside `compute_offset`. This kind of break the purpose of `check_valid_grid` that checks the size of the `interpolation_domain`.

I don't have yet an idea to workaround this.

EmilyBourne:
> I am trying to implement tests that would trigger one of the exceptions using the SplineBuilder constructor.
> 
> I first tried with a too short `interpolation_domain` but it failed in `compute_offset` before `check_valid_grid`. There is an out-of-bound access inside `compute_offset`. This kind of break the purpose of `check_valid_grid` that checks the size of the `interpolation_domain`.
> 
> I don't have yet an idea to workaround this.

`m_offset` is not const. So can't we just initialise this value in the body of the constructor after the call to `check_valid_grid`?

tpadioleau:
> `m_offset` is not const. So can't we just initialise this value in the body of the constructor after the call to `check_valid_grid`?

We can, the linter will complain but I can disable it.

tpadioleau:
@EmilyBourne can you review the test ?

tests/splines/spline_builder.cpp
```diff
@@ -0,0 +1,139 @@
...
+
+TEST(SplineBuilder, CorrectInterpolationGrid)
+{
+    CoordX const x0(0.);
+    CoordX const xN(1.);
+    std::size_t const ncells = 5;
+
+    ddc::init_discrete_space<BSplinesX>(x0, xN, ncells);
+
+    std::vector<double> const range {0.1, 0.11, 0.25, 0.23, 0.24};
```
EmilyBourne, line 125: These are not ordered. Is that intentional? Should we also check this? (The points should be ordered)

tpadioleau: good catch, and we should detect that! What do you think in the construction of the NonUniformPointSampling ? (you do not have unordered tricks in Gysela ?)

EmilyBourne: I think the construction of the NonUniformPointSampling seems like a logical place for this and that will be more general than putting it in `check_n_points_in_cell`

We don't use unordered points in Gysela

tpadioleau: Fixed in 04523404841cbe4ac863839a5be1797d41bd7a55, the sorting check is already in `main`

Issue #438, opened 2024-05-03
Create a SubDomain


EmilyBourne:
I would like to be able to create some kind of SubDomain. For example this could be achieved by adding a stride to the Domain. This would be useful in several places.

It can be used to iterate over a sliced domain. This can be useful for creating a Lagrange interpolation or calculating a quadrature (e.g. Simpson's quadrature).
E.g. currently a Simpson's quadrature is written as:
```cpp
    for (auto it = domain.begin() + 1; it < domain.end() - 1; it += 2) {
        ddc::DiscreteElement<IDim> idx = *it;
        coefficients(idx) = 2. / 3. * (distance_at_left(idx) + distance_at_right(idx));
        idx += ddc::DiscreteVector<IDim>(1);
        coefficients(idx) = 1. / 3. * (distance_at_left(idx) + distance_at_right(idx));
    }
```
with a SubDomain it would be:
```cpp
    for (ddc::DiscreteElement<IDim> idx : subdomain) {
        ddc::DiscreteElement<IDim> idx2 = idx + ddc::DiscreteVector<IDim>(1);
        coefficients(idx) = 2. / 3. * (distance_at_left(idx) + distance_at_right(idx));
        coefficients(idx2) = 1. / 3. * (distance_at_left(idx2) + distance_at_right(idx2));
    }
```

It can also be used to create a Chunk on a SubDomain. Eg when working with patches it is common to need values of a function over all the domain, but the derivatives only at certain points (usually the boundary). Currently this is only possible by explicitly choosing the number of points at the compilation, or creating a separate mesh for the derivatives (which then have different `DiscreteElements`)

tpadioleau:
I am thinking of defining a `ddc::Chunk`/`ddc::ChunkSpan` not only over a `ddc::DiscreteDomain` but over a cartesian product of 1D ranges of `ddc::DiscreteElement`. In this case `ddc::DiscreteDomain` would be the particular case of a cartesian product of 1D continuous ranges. This way we could also introduce a strided range of `ddc::DiscreteElement`.

Accessing `ddc::Chunk`/`ddc::ChunkSpan` data using `ddc::DiscreteElement` could be slow depending on the type of range. For other reasons I would also like to introduce an access using `ddc::DiscreteVector` that would always be "fast" compared to the latter.

Do you think it would help you ?

EmilyBourne:
[...]

PR #698, opened 2024-12-05
Generalized domains

tpadioleau:
PoC for #438 

Steps:
- [x] remove internal_mdspan
- [x] implement `Chunk` and `ChunkSpan` with `SupportType` being a customization point
- [x] implement a `StridedDiscreteDomain`
- [x] implement a `StorageDiscreteDomain` 
- [x] update algorithms

Not clear if we should define an equivalent of the slice operator taking a `DiscreteDomain`

cc @science-enthusiast

### DIFF
--- a/include/ddc/chunk_common.hpp
+++ b/include/ddc/chunk_common.hpp
@@ -173,8 +173,7 @@ public:
     template <class QueryDDim>
     KOKKOS_FUNCTION constexpr size_type stride() const
     {
-        return m_allocation_mdspan.stride(
-                type_seq_rank_v<QueryDDim, detail::ToTypeSeq<SupportType>>);
+        return m_allocation_mdspan.stride(type_seq_rank_v<QueryDDim, to_type_seq_t<SupportType>>);
     }
 
     /** Provide access to the domain on which this chunk is defined
--- a/include/ddc/chunk_common.hpp
+++ b/include/ddc/chunk_common.hpp
@@ -32,21 +32,21 @@ KOKKOS_FUNCTION auto get_domain(ChunkType const& chunk) noexcept
 template <class ElementType, class SupportType, class LayoutStridedPolicy>
 class ChunkCommon;
 
-template <class ElementType, class... DDims, class LayoutStridedPolicy>
-class ChunkCommon<ElementType, DiscreteDomain<DDims...>, LayoutStridedPolicy>
+template <class ElementType, class SupportType, class LayoutStridedPolicy>
+class ChunkCommon
 {
 public:
-    using discrete_domain_type = DiscreteDomain<DDims...>;
+    using discrete_domain_type = SupportType;
 
     /// The dereferenceable part of the co-domain but with a different domain, starting at 0
     using allocation_mdspan_type = Kokkos::mdspan<
             ElementType,
-            Kokkos::dextents<std::size_t, sizeof...(DDims)>,
+            Kokkos::dextents<std::size_t, SupportType::rank()>,
             LayoutStridedPolicy>;
 
     using const_allocation_mdspan_type = Kokkos::mdspan<
             const ElementType,
-            Kokkos::dextents<std::size_t, sizeof...(DDims)>,
+            Kokkos::dextents<std::size_t, SupportType::rank()>,
             LayoutStridedPolicy>;
 
     using discrete_element_type = typename discrete_domain_type::discrete_element_type;
@@ -86,7 +86,7 @@ protected:
     allocation_mdspan_type m_allocation_mdspan;
 
     /// The mesh on which this chunk is defined
-    discrete_domain_type m_domain;
+    SupportType m_domain;
 
 public:
     static KOKKOS_FUNCTION constexpr int rank() noexcept
@@ -123,9 +123,9 @@ private:
     template <class Mapping = mapping_type>
     static KOKKOS_FUNCTION constexpr std::
             enable_if_t<std::is_constructible_v<Mapping, extents_type>, allocation_mdspan_type>
-            make_allocation_mdspan(ElementType* ptr, discrete_domain_type const& domain)
+            make_allocation_mdspan(ElementType* ptr, SupportType const& domain)
     {
-        return allocation_mdspan_type(ptr, ::ddc::extents<DDims>(domain).value()...);
+        return allocation_mdspan_type(ptr, detail::array(domain.extents()));
     }
 
 public:
@@ -134,7 +134,7 @@ public:
         return m_allocation_mdspan.accessor();
     }
 
-    KOKKOS_FUNCTION constexpr DiscreteVector<DDims...> extents() const noexcept
+    KOKKOS_FUNCTION constexpr SupportType::discrete_vector_type extents() const noexcept
     {
         return m_domain.extents();
     }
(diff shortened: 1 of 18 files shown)


include/ddc/discrete_domain.hpp
```diff
@@ -174,6 +174,12 @@ class DiscreteDomain
         return DiscreteDomain(front() + n1, extents() - n1 - n2);
     }
 
+    KOKKOS_FUNCTION constexpr DiscreteElement<DDims...> operator()(
```
science-enthusiast, line 177: This looks like an essential operator. Till now, what did the user do to achieve same functionality?

include/ddc/chunk_common.hpp
```diff
@@ -129,26 +122,16 @@ class ChunkCommon<ElementType, DiscreteDomain<DDims...>, LayoutStridedPolicy>
 private:
     template <class Mapping = mapping_type>
     static KOKKOS_FUNCTION constexpr std::
```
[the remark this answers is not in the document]
tpadioleau: What do you suggest ? Disabling automatic formatting for these lines ?

include/ddc/discrete_domain.hpp
tpadioleau: This operation used to be done directly in `Chunk` and `ChunkSpan`, see https://github.com/CExA-project/ddc/blob/9f53744b90f0b216fa27217561a4d9b21344b6d1/include/ddc/chunk_span.hpp#L338

include/ddc/chunk_common.hpp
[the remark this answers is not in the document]
science-enthusiast: I see your point. If it is automatically formatted this way then nothing can be done.

include/ddc/chunk_span.hpp
```diff
@@ -50,22 +49,14 @@ class ChunkSpan<ElementType, DiscreteDomain<DDims...>, LayoutStridedPolicy, Memo
...
 
     /// type of a view of this full chunk
-    using view_type = ChunkSpan<
-            ElementType const,
-            DiscreteDomain<DDims...>,
-            LayoutStridedPolicy,
-            MemorySpace>;
+    using view_type = ChunkSpan<ElementType const, SupportType, LayoutStridedPolicy, MemorySpace>;
 
     using discrete_domain_type = typename base_type::discrete_domain_type;
```
science-enthusiast, line 61: In this version of chunk_span.hpp, the only place where discrete_domain_type is used, is to define discrete_element_type below. Moreover, discrete_element_type is not used anymore. Thus both discrete_domain_type and discrete_element_type can be removed.

include/ddc/discrete_domain.hpp
```diff
@@ -187,6 +193,33 @@ class DiscreteDomain
                 DiscreteVector<DDims...>((get_or<DDims>(oextents, get<DDims>(myextents)))...));
     }
 
+    template <class... DElems>
+    bool is_inside(DElems const&... delems) const noexcept
+    {
```
science-enthusiast, line 198: is_inside(...) is now available in both DiscreteDomain and StridedDiscreteDomain. Is the function a feature request or is it useful for implementing generalized domains?

tpadioleau: It is the precondition to call `distance_from_front`. There should be an assert `is_inside` in `distance_from_front`

science-enthusiast: Thanks for the clarification.

include/ddc/storage_discrete_domain.hpp
```diff
@@ -0,0 +1,622 @@
...
+
+template <class... DDims, class... ODDims>
+struct RebindDomain<StorageDiscreteDomain<DDims...>, detail::TypeSeq<ODDims...>>
+{
+    using type = StorageDiscreteDomain<ODDims...>;
+};
+
+template <class InputIt1, class InputIt2>
+KOKKOS_FUNCTION bool equal(InputIt1 first1, InputIt1 last1, InputIt2 first2)
+{
```
science-enthusiast, line 63: Could a `is_pointer_v` based concept help with better usage of this free function?

include/ddc/storage_discrete_domain.hpp
```diff
@@ -0,0 +1,622 @@
...
+        it = first;
+        typename std::iterator_traits<ForwardIt>::difference_type const step = count / 2;
+        it += step;
+
+        if (comp(*it, value)) {
+            first = ++it;
+            count -= step + 1;
+        } else {
+            count = step;
+        }
```
science-enthusiast, line 91: Could `count` be updated as `count = last - first` just after this `if-then-else` construct? `first` could be updated as it is  being updated right now and `last` could be updated inside the `else` block as `last = it`. By doing this `count` is updated in the same manner it was initialized.

include/ddc/storage_discrete_domain.hpp
```diff
@@ -0,0 +1,622 @@
...
+}
+
+template <
+        class ForwardIt,
+        class T = typename std::iterator_traits<ForwardIt>::value_type,
+        class Compare>
+bool binary_search(ForwardIt first, ForwardIt last, const T& value, Compare comp)
+{
+    first = ::ddc::detail::lower_bound(first, last, value, comp);
+    return (!(first == last) && !(comp(value, *first)));
```
science-enthusiast, line 104: If we are sufficiently careful with how `lower_bound` and `comp` are implemented, could just `!(first == last)` be enough?

include/ddc/storage_discrete_domain.hpp
tpadioleau: Discussed in person, I don't think so because it would require to assume an implementation detail on Kokkos iterators.

include/ddc/storage_discrete_domain.hpp
```diff
@@ -0,0 +1,622 @@
...
+}
+
+template <
+        class ForwardIt,
+        class T = typename std::iterator_traits<ForwardIt>::value_type,
+        class Compare>
+KOKKOS_FUNCTION ForwardIt lower_bound(ForwardIt first, ForwardIt last, const T& value, Compare comp)
+{
+    ForwardIt it;
+    typename std::iterator_traits<ForwardIt>::difference_type count = last - first;
```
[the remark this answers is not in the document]
tpadioleau: Yes I will update

include/ddc/storage_discrete_domain.hpp
tpadioleau: I don't, to be honest I copy-pasted the version found on cppref

include/ddc/storage_discrete_domain.hpp
tpadioleau: Same as `lower_bound`, this is a copy paste of the version found on cppref

science-enthusiast: I had a closer look at [std::binary_search](https://en.cppreference.com/w/cpp/algorithm/binary_search) and [std::lower_bound](https://en.cppreference.com/w/cpp/algorithm/lower_bound) and understand better now. I was wrong in my previous comment. Both `!(first == last)` and `!(comp(value, *first))` are needed. `lower_bound` searches for the first element in the partitioned range `[first, last)` which is not ordered before `value`. `lower_bound` can return `last` only if ` bool(comp(*iter, value))` is `true` for all `iter`s. Both `lower_bound` and `binary_search` assume that `[first, last)` is partitioned. Also `comp` behaves like `<` and not `<=`. If for any element `elem` of `[first, last)`, if `bool(comp(elem, value))` does not imply `!bool(comp(value, elem))` then the behavior is undefined. Hence the first element for which ` bool(comp(*iter, value))` is `false` should be either equal or greater than `value`.

tpadioleau:
@science-enthusiast do you think it can be merged as is ?

science-enthusiast:
> @science-enthusiast do you think it can be merged as is ?

Yes, I think it can be merged. So that people can start using it. I have a few doubts that I will ask you. For me, this has been a good way to go deeper into the DDC code base. In the future, I will be able to make deeper comments.

science-enthusiast (approved):
The DDC code base is new to me and I continue to understand it better by reviewing this PR. It is better to merge these changes and let people start using it. Issues can be addressed in the future.

tpadioleau:
> The DDC code base is new to me and I continue to understand it better by reviewing this PR. It is better to merge these changes and let people start using it. Issues can be addressed in the future.

Yes thank you for your time.

PR #708, opened 2024-12-15
KOKKOS_FUNCTION-annotated for_each and transform_reduce

### LINKED ISSUE #695: ddc::for_each should be marked KOKKOS_FUNCTION

blegouix:
Closes #695

### DIFF
--- a/include/ddc/for_each.hpp
+++ b/include/ddc/for_each.hpp
@@ -16,8 +16,9 @@ namespace ddc {
 
 namespace detail {
 
+#pragma hd_warning_disable
 template <class RetType, class Element, std::size_t N, class Functor, class... Is>
-void for_each_serial(
+KOKKOS_FUNCTION void for_each_serial(
         std::array<Element, N> const& begin,
         std::array<Element, N> const& end,
         Functor const& f,
@@ -39,8 +40,9 @@ void for_each_serial(
  * @param[in] domain the domain over which to iterate
  * @param[in] f      a functor taking an index as parameter
  */
+#pragma hd_warning_disable
 template <class... DDims, class Functor>
-void for_each(DiscreteDomain<DDims...> const& domain, Functor&& f) noexcept
+KOKKOS_FUNCTION void for_each(DiscreteDomain<DDims...> const& domain, Functor&& f) noexcept
 {
     DiscreteElement<DDims...> const ddc_begin = domain.front();
     DiscreteElement<DDims...> const ddc_end = domain.front() + domain.extents();
--- a/include/ddc/transform_reduce.hpp
+++ b/include/ddc/transform_reduce.hpp
@@ -23,13 +23,14 @@ namespace detail {
  *            range. The return type must be acceptable as input to reduce
  * @param[in] dcoords discrete elements from dimensions already in a loop
  */
+#pragma hd_warning_disable
 template <
         class... DDims,
         class T,
         class BinaryReductionOp,
         class UnaryTransformOp,
         class... DCoords>
-T transform_reduce_serial(
+KOKKOS_FUNCTION T transform_reduce_serial(
         DiscreteDomain<DDims...> const& domain,
         [[maybe_unused]] T const neutral,
         BinaryReductionOp const& reduce,
@@ -62,8 +63,9 @@ T transform_reduce_serial(
  * @param[in] transform a unary FunctionObject that will be applied to each element of the input
  *            range. The return type must be acceptable as input to reduce
  */
+#pragma hd_warning_disable
 template <class... DDims, class T, class BinaryReductionOp, class UnaryTransformOp>
-T transform_reduce(
+KOKKOS_FUNCTION T transform_reduce(
         DiscreteDomain<DDims...> const& domain,
         T neutral,
         BinaryReductionOp&& reduce,
--- a/tests/for_each.cpp
+++ b/tests/for_each.cpp
@@ -10,6 +10,8 @@
 
 #include <gtest/gtest.h>
 
+#include <Kokkos_StdAlgorithms.hpp>
+
 namespace DDC_HIP_5_7_ANONYMOUS_NAMESPACE_WORKAROUND(FOR_EACH_CPP) {
 
 using DElem0D = ddc::DiscreteElement<>;
@@ -80,4 +82,29 @@ TEST(ForEachSerialHost, TwoDimensions)
     ddc::ChunkSpan<int, DDomXY> const view(storage.data(), dom);
     ddc::for_each(dom, [=](DElemXY const ixy) { view(ixy) += 1; });
     EXPECT_EQ(std::count(storage.begin(), storage.end(), 1), dom.size());
+}
+
+void TestForEachSerialDevice(ddc::ChunkSpan<int, DDomXY> view)
+{
+    ddc::parallel_for_each(
+            Kokkos::DefaultExecutionSpace(),
+            ddc::DiscreteDomain<>(),
+            KOKKOS_LAMBDA(ddc::DiscreteElement<> i) {
+                ddc::for_each(view.domain(), [=](DElemXY const ixy) { view(ixy) = 1; });
+            });
+}
+
+TEST(ForEachSerialDevice, TwoDimensions)
+{
+    DDomXY const dom(lbound_x_y, nelems_x_y);
+    Kokkos::View<int*, Kokkos::LayoutRight, Kokkos::DefaultExecutionSpace> storage("", dom.size());
+    ddc::ChunkSpan<int, DDomXY> const view(storage.data(), dom);
+    TestForEachSerialDevice(view);
+    EXPECT_EQ(
+            Kokkos::Experimental::
+                    count(Kokkos::DefaultExecutionSpace(),
+                          Kokkos::Experimental::begin(storage),
+                          Kokkos::Experimental::end(storage),
+                          1),
+            dom.size());
 }
--- a/tests/transform_reduce.cpp
+++ b/tests/transform_reduce.cpp
@@ -74,3 +74,31 @@ TEST(TransformReduce, TwoDimensions)
             ddc::transform_reduce(dom, 0, ddc::reducer::sum<int>(), chunk),
             dom.size() * (dom.size() - 1) / 2);
 }
+
+int TestTransformReduceDevice(ddc::ChunkSpan<int, DDomXY> chunk)
+{
+    ddc::parallel_for_each(
+            Kokkos::DefaultExecutionSpace(),
+            ddc::DiscreteDomain<>(),
+            KOKKOS_LAMBDA(ddc::DiscreteElement<> i) {
+                ddc::for_each(chunk.domain(), [=](DElemXY const ixy) { chunk(ixy) = 1; });
+            });
+    Kokkos::View<int, Kokkos::LayoutRight, Kokkos::DefaultExecutionSpace> count("");
+    ddc::parallel_for_each(
+            Kokkos::DefaultExecutionSpace(),
+            ddc::DiscreteDomain<>(),
+            KOKKOS_LAMBDA(ddc::DiscreteElement<> i) {
+                count() = ddc::transform_reduce(chunk.domain(), 0, ddc::reducer::sum<int>(), chunk);
+            });
+    Kokkos::View<int, Kokkos::LayoutRight, Kokkos::DefaultHostExecutionSpace> count_host
+            = Kokkos::create_mirror_view_and_copy(Kokkos::DefaultHostExecutionSpace(), count);
+    return count_host();
+}
+
+TEST(TransformReduceDevice, TwoDimensions)
+{
+    DDomXY const dom(lbound_x_y, nelems_x_y);
+    Kokkos::View<int*, Kokkos::LayoutRight, Kokkos::DefaultExecutionSpace> storage("", dom.size());
+    ddc::ChunkSpan<int, DDomXY> const chunk(storage.data(), dom);
+    EXPECT_EQ(TestTransformReduceDevice(chunk), dom.size() * dom.size());
+}


blegouix:
- The tests fails on GPU (this reveals a bug with this usage of `ddc::for_each` ?).
- `hd_warning_disable` is not supported by hipcc. Should we accept the warnings ?

tests/for_each.cpp
```diff
@@ -81,3 +83,36 @@ TEST(ForEachSerialHost, TwoDimensions)
...
+void TestForEachSerialDevice(ddc::ChunkSpan<
+                             int,
+                             DDomXY,
+                             Kokkos::layout_right,
+                             typename Kokkos::DefaultExecutionSpace::memory_space> view)
+{
+    ddc::parallel_for_each(
+            Kokkos::DefaultExecutionSpace(),
+            ddc::DiscreteDomain<>(),
+            KOKKOS_LAMBDA([[maybe_unused]] ddc::DiscreteElement<> i) {
```
blegouix, line 96:
```suggestion
            KOKKOS_LAMBDA([[maybe_unused]] ddc::DiscreteElement<>) {
```

tests/transform_reduce.cpp
```diff
@@ -74,3 +74,39 @@ TEST(TransformReduce, TwoDimensions)
...
+int TestTransformReduceDevice(ddc::ChunkSpan<
+                              int,
+                              DDomXY,
+                              Kokkos::layout_right,
+                              typename Kokkos::DefaultExecutionSpace::memory_space> chunk)
+{
+    ddc::parallel_for_each(
+            Kokkos::DefaultExecutionSpace(),
+            ddc::DiscreteDomain<>(),
+            KOKKOS_LAMBDA([[maybe_unused]] ddc::DiscreteElement<> i) {
```
blegouix, line 87:
```suggestion
            KOKKOS_LAMBDA([[maybe_unused]] ddc::DiscreteElement<>) {
```

tests/transform_reduce.cpp
```diff
@@ -74,3 +74,39 @@ TEST(TransformReduce, TwoDimensions)
...
+            Kokkos::DefaultExecutionSpace(),
+            ddc::DiscreteDomain<>(),
+            KOKKOS_LAMBDA([[maybe_unused]] ddc::DiscreteElement<>) {
+                ddc::for_each(chunk.domain(), [=](DElemXY const ixy) { chunk(ixy) = 1; });
+            });
+    Kokkos::View<int, Kokkos::LayoutRight, Kokkos::DefaultExecutionSpace> count("");
+    ddc::parallel_for_each(
+            Kokkos::DefaultExecutionSpace(),
+            ddc::DiscreteDomain<>(),
+            KOKKOS_LAMBDA([[maybe_unused]] ddc::DiscreteElement<> i) {
```
blegouix, line 94:
```suggestion
            KOKKOS_LAMBDA([[maybe_unused]] ddc::DiscreteElement<>) {
```

include/ddc/for_each.hpp
```diff
@@ -16,8 +16,11 @@ namespace ddc {
 
 namespace detail {
 
+#if defined KOKKOS_ENABLE_CUDA || defined KOKKOS_ENABLE_HIP
```
blegouix, line 19:
```suggestion
#if defined KOKKOS_ENABLE_CUDA
```

include/ddc/for_each.hpp
```diff
@@ -39,8 +42,11 @@ void for_each_serial(
  * @param[in] domain the domain over which to iterate
  * @param[in] f      a functor taking an index as parameter
  */
+#if defined KOKKOS_ENABLE_CUDA || defined KOKKOS_ENABLE_HIP
```
blegouix, line 45:
```suggestion
#if defined KOKKOS_ENABLE_CUDA
```

include/ddc/transform_reduce.hpp
```diff
@@ -23,13 +23,16 @@ namespace detail {
  *            range. The return type must be acceptable as input to reduce
  * @param[in] dcoords discrete elements from dimensions already in a loop
  */
+#if defined KOKKOS_ENABLE_CUDA || defined KOKKOS_ENABLE_HIP
```
blegouix, line 26:
```suggestion
#if defined KOKKOS_ENABLE_CUDA
```

include/ddc/transform_reduce.hpp
```diff
@@ -62,8 +65,11 @@ T transform_reduce_serial(
  * @param[in] transform a unary FunctionObject that will be applied to each element of the input
  *            range. The return type must be acceptable as input to reduce
  */
+#if defined KOKKOS_ENABLE_CUDA || defined KOKKOS_ENABLE_HIP
```
blegouix, line 68:
```suggestion
#if defined KOKKOS_ENABLE_CUDA
```

tests/for_each.cpp
```diff
@@ -81,3 +83,36 @@ TEST(ForEachSerialHost, TwoDimensions)
...
+void TestForEachSerialDevice(ddc::ChunkSpan<
+                             int,
+                             DDomXY,
+                             Kokkos::layout_right,
+                             typename Kokkos::DefaultExecutionSpace::memory_space> view)
+{
+    ddc::parallel_for_each(
+            Kokkos::DefaultExecutionSpace(),
+            ddc::DiscreteDomain<>(),
+            KOKKOS_LAMBDA([[maybe_unused]] ddc::DiscreteElement<>) {
```
blegouix, line 96:
```suggestion
            KOKKOS_LAMBDA([[maybe_unused]] ddc::DiscreteElement<> unused_elem) {
```

tests/transform_reduce.cpp
```diff
@@ -74,3 +74,39 @@ TEST(TransformReduce, TwoDimensions)
...
+int TestTransformReduceDevice(ddc::ChunkSpan<
+                              int,
+                              DDomXY,
+                              Kokkos::layout_right,
+                              typename Kokkos::DefaultExecutionSpace::memory_space> chunk)
+{
+    ddc::parallel_for_each(
+            Kokkos::DefaultExecutionSpace(),
+            ddc::DiscreteDomain<>(),
+            KOKKOS_LAMBDA([[maybe_unused]] ddc::DiscreteElement<>) {
```
blegouix, line 87:
```suggestion
            KOKKOS_LAMBDA([[maybe_unused]] ddc::DiscreteElement<> unused_elem) {
```

tests/transform_reduce.cpp
```diff
@@ -74,3 +74,39 @@ TEST(TransformReduce, TwoDimensions)
...
+                ddc::for_each(chunk.domain(), [=](DElemXY const ixy) { chunk(ixy) = 1; });
+            });
+    Kokkos::View<int, Kokkos::LayoutRight, Kokkos::DefaultExecutionSpace> count("");
+    ddc::parallel_for_each(
+            Kokkos::DefaultExecutionSpace(),
+            ddc::DiscreteDomain<>(),
+            KOKKOS_LAMBDA([[maybe_unused]] ddc::DiscreteElement<>) {
```
blegouix, line 94:
```suggestion
            KOKKOS_LAMBDA([[maybe_unused]] ddc::DiscreteElement<> unused_elem) {
```

tpadioleau:
I don't know if i will have time to have a closer look this week. You could look for the differences with the implementation in the first PR, i think the tests were passing

blegouix:
I think I have the same implementation as the closed branch. But you don't have 2D test for nested for_each, and I think the bug with nvcc appears only for 2D+ cases.

tpadioleau:
> and I think the bug with nvcc appears only for 2D+ cases.

Did you test ?

blegouix:
The test passes with this:
image

tests/for_each.cpp
```diff
@@ -81,3 +83,75 @@ TEST(ForEachSerialHost, TwoDimensions)
...
+                                        Kokkos::layout_right,
+                                        typename Kokkos::DefaultExecutionSpace::memory_space> view)
+{
+    ddc::parallel_for_each(
+            Kokkos::DefaultExecutionSpace(),
+            DDom0D(),
+            KOKKOS_LAMBDA([[maybe_unused]] DElem0D unused_elem) {
+                Kokkos::printf("%i\n", view.template extent<DDimX>());
+                ddc::annotated_for_each(view.domain(), [=](DElemX const ix) {
+                    Kokkos::printf("lul");
```
blegouix, line 99:
```suggestion
```

tests/for_each.cpp
```diff
@@ -81,3 +83,75 @@ TEST(ForEachSerialHost, TwoDimensions)
...
+                                        int,
+                                        DDomX,
+                                        Kokkos::layout_right,
+                                        typename Kokkos::DefaultExecutionSpace::memory_space> view)
+{
+    ddc::parallel_for_each(
+            Kokkos::DefaultExecutionSpace(),
+            DDom0D(),
+            KOKKOS_LAMBDA([[maybe_unused]] DElem0D unused_elem) {
+                Kokkos::printf("%i\n", view.template extent<DDimX>());
```
blegouix, line 97:
```suggestion
```

include/ddc/for_each.hpp
```diff
@@ -33,6 +33,23 @@ void for_each_serial(
...
+        std::array<Element, N> const& begin,
+        std::array<Element, N> const& end,
+        Functor const& f,
+        Is const&... is) noexcept
+{
+    static constexpr std::size_t I = sizeof...(Is);
+    if constexpr (I == N) {
+        f(RetType(is...));
+    } else {
+        for (Element ii = static_cast<int>(begin[I]); ii < end[I]; ++ii) {
```
blegouix, line 47: This `static_cast<int>` is a nonsense workaround since `Element` aliases `std::size_t`. Without it the loop does not iterate with `nvcc` in the new tests.

tpadioleau: What does it print if you remove the cast ?

blegouix: It prints `0` as expected 🤷

image

blegouix: (if I print something inside the loop it appears only once)

tpadioleau: If it can print something inside the loop it means it can iterate once ?

What is the value of end[I] ?

Have you tried to run in debug ?

blegouix: Yes sorry, by "it does not iterate" I meant it goes just once through the loop body.

`end[I]` is printed as `10`

I am exclusively working in Debug mode.

blegouix:
```
Kokkos::printf("%i %i\n", end[I], end[I]);
Kokkos::printf("%i %i\n", static_cast<int>(end[I]), static_cast<int>(end[I]));
```

prints:

```
10 0
10 10
```

Like if accessing `begin[I]` and `end[I]` was buggy without the cast

tpadioleau: Here it is likely an incompatible format specifier passed to printf, it needs to match the type of integer you pass. You need to change %i if you want to print a size_t

blegouix: Ah right, changing it to %llu correctly prints 0 for begin[I] and 10 for end[I] in all situations so my last comment was irrelevant

tpadioleau: If you print from inside the loop with this correction, is it better ?

blegouix: sorry what do you want me to print inside the loop ?

tpadioleau: You could print `ii` ?

blegouix: `Kokkos::printf("%llu\n", ii);` prints `0`

blegouix:
I have found another bug. This is not visible in the tests (should I add a test which reproduces it ?). I have a use-case where the `annotated_transform_reduce` returns `0` with CUDA (it should not). Here is a fix:

```
result = reduce(
                    result,
                    annotated_transform_reduce_serial(
                            domain,
                            neutral,
                            reduce,
                            transform,
                            dcoords...,
                            ii));
```

Must be replaced with:

```
T tmp;
            Kokkos::atomic_store(
                    &tmp,
                    annotated_transform_reduce_serial(
                            domain,
                            neutral,
                            reduce,
                            transform,
                            dcoords...,
                            ii));
            Kokkos::atomic_store(&result, reduce(result, tmp));
```

(I don't understand what is the problem as `result` is a local per-thread variable though)

tpadioleau:
All of that is weird, can you provide more information about your environment ?
- the hardware, cpu and gpu
- the version of your compiler
- the version of the libraries, cmake, cuda, kokkos...
- the complete cmake command line
- anything you think relevant

Also do all other DDC tests pass ?
Do kokkos tests pass ?

It feels like the compiler optimized away some function calls

blegouix:
- Ubuntu 24
- Intel(R) Xeon(R) Gold 6226R + V100S
- nvcc 12.6 + g++13.3
- cmake 3.28, cuda 12.6, the Kokkos is the DDC submodule up-to-date with the main DDC branch.
- cmake -DCMAKE_CXX_COMPILER=$KOKKOS_BIN/nvcc_wrapper -DCMAKE_BUILD_TYPE=Debug -DKokkos_ENABLE_CUDA=ON -DKokkos_ARCH_VOLTA70=ON -DCMAKE_CXX_FLAGS="-Wfatal-errors" ..

I will try quite soon on a 1070Ti with mostly the same software configuration. I will try to dig in `compute-sanitizer` also. Maybe if you can see if my first problem reproduces on Ruche or Persée by going in this branch and compiling with or without the `static_cast<int>` it may be helpful ? I agree all this is weird and i'm feeling `nvcc` is capricious, but it is possible that I am missing something.

tpadioleau:
From the table https://docs.nvidia.com/cuda/cuda-installation-guide-linux/#system-requirements you are one minor version above the maximum tested host compiler version, i.e. 13.2.

Can you try with a different compiler ?

include/ddc/for_each.hpp
blegouix: I do reproduce with `g++-12`, but not with `-G`, so it seems to be an optimization issue indeed.

blegouix: I don't see anything with compute-sanitizer (tried every `--tool`)

tpadioleau:
I should have time to look at it in february, thanks for the work so far

PaulGannay:
@blegouix I will be finishing the work for this PR.
I haven't been able to reproduce the bug you are reporting, even on a GTX1070, so it will be merged mostly as is.

PR #732, opened 2025-01-02
Compute coordinates only on dimensions of interest

tpadioleau:
Remove the evaluation of coordinates in batch dimensions. The reason is that these dimensions don't need to be samplings, the related coordinates are actually ignored in the evaluator.

The batch dimensions are replaced by dummy dimensions that don't need to be initialised.

### DIFF
--- a/tests/splines/batched_2d_spline_builder.cpp
+++ b/tests/splines/batched_2d_spline_builder.cpp
@@ -62,6 +62,10 @@ struct DimZ
 };
 #endif
 
+struct DDimBatch
+{
+};
+
 constexpr std::size_t s_degree = DEGREE;
 
 #if defined(BC_PERIODIC)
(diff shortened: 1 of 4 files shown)


tpadioleau:
@EmilyBourne I would like your opinion on this PR. What do you think ?

EmilyBourne (approved):
This looks fine to me. I'm not sure that `DElem<DDim<I1, I1, I2>, DDim<I2, I1, I2>>` is the most readable but as this is the tests not the doc I don't think it matters.

tpadioleau:
> This looks fine to me. I'm not sure that `DElem<DDim<I1, I1, I2>, DDim<I2, I1, I2>>` is the most readable but as this is the tests not the doc I don't think it matters.

I agree, it is on the todo list to remove this alias, it will be easier after this PR.

tpadioleau:
thank you

PR #777, opened 2025-02-15
Set cmake warnings as errors and rework compilation warnings as errors

tpadioleau:


tpadioleau:
Closing, we use cmake 3.22 and HIP triggers an error while detecting the number of devices.

PR #780, opened 2025-02-18
Remove deprecated usage of public member variables of DualView

tpadioleau:
see https://github.com/kokkos/kokkos/commit/b027fc25977f0e879c43febe1d884585fe8c1db0

### DIFF
--- a/include/ddc/kernels/splines/splines_linear_problem_band.hpp
+++ b/include/ddc/kernels/splines/splines_linear_problem_band.hpp
@@ -154,7 +154,7 @@ public:
         }
 
         // Convert 1-based index to 0-based index
-        for (int i = 0; i < size(); ++i) {
+        for (std::size_t i = 0; i < size(); ++i) {
             m_ipiv.view_host()(i) -= 1;
         }
 
--- a/include/ddc/kernels/splines/splines_linear_problem_band.hpp
+++ b/include/ddc/kernels/splines/splines_linear_problem_band.hpp
@@ -81,7 +81,7 @@ public:
         assert(m_kl <= mat_size);
         assert(m_ku <= mat_size);
 
-        Kokkos::deep_copy(m_q.h_view, 0.);
+        Kokkos::deep_copy(m_q.view_host(), 0.);
     }
 
 private:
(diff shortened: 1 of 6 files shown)


include/ddc/kernels/splines/splines_linear_problem_band.hpp
```diff
@@ -144,18 +144,18 @@ class SplinesLinearProblemBand : public SplinesLinearProblem<ExecSpace>
...
+                m_q.view_host().stride(
+                        0), // m_q.view_host().stride(0) if LAPACK_ROW_MAJOR, m_q.view_host().stride(1) if LAPACK_COL_MAJOR
+                m_ipiv.view_host().data());
         if (info != 0) {
             throw std::runtime_error(
                     "LAPACKE_dgbtrf failed with error code " + std::to_string(info));
         }
 
         // Convert 1-based index to 0-based index
         for (int i = 0; i < size(); ++i) {
```
yasahi-hpc, line 157:
```suggestion
        for (std::size_t i = 0; i < size(); ++i) {
```

It is nothing to do with this PR, but should this be `std::size_t`?

include/ddc/kernels/splines/splines_linear_problem_dense.hpp
```diff
@@ -83,17 +83,17 @@ class SplinesLinearProblemDense : public SplinesLinearProblem<ExecSpace>
...
                 size(),
-                m_ipiv.h_view.data());
+                m_ipiv.view_host().data());
         if (info != 0) {
             throw std::runtime_error(
                     "LAPACKE_dgetrf failed with error code " + std::to_string(info));
         }
 
         // Convert 1-based index to 0-based index
         for (int i = 0; i < size(); ++i) {
```
yasahi-hpc, line 95: Same as the comment before.

yasahi-hpc (dismissed):
LGTM. 
I have minor comments about the index type. 
Another point is the usage of `assert` and `throw` (not the scope of this PR). In my opinion, we should always `throw` here since we cannot operate appropriately if conditions are not satisfied.

include/ddc/kernels/splines/splines_linear_problem_band.hpp
tpadioleau: Yes it would be better

include/ddc/kernels/splines/splines_linear_problem_dense.hpp
tpadioleau, line 95:
```suggestion
        for (std::size_t i = 0; i < size(); ++i) {
```

PR #785, opened 2025-02-21
Introduce ddc::type_seq_cat_t

tpadioleau:


include/ddc/detail/type_seq.hpp
```diff
@@ -104,6 +104,23 @@ struct TypeSeqMerge<TypeSeq<TagsA...>, TypeSeq<HeadTagsB, TailTagsB...>, TypeSeq
...
+{
+    using type = TypeSeq<TagsR...>;
+};
+
+template <class... TagsA, class... TagsB>
+struct TypeSeqCat<TypeSeq<TagsA...>, TypeSeq<TagsB...>, TypeSeq<>>
+    : TypeSeqCat<TypeSeq<>, TypeSeq<>, TypeSeq<TagsA..., TagsB...>>
+{
+};
+
```
EmilyBourne, line 123: Why do you prefer inheritance here over simply saving the result?
```suggestion
/// R contains all elements in A then all elements in B.
/// Example: A = [a, b, c], B = [z, c, y], R = [a, b, c, z, c, y]
template <class TagSeqA, class TagSeqB, class TagSeqR>
struct TypeSeqCat;

template <class... TagsA, class... TagsB>
struct TypeSeqCat<TypeSeq<TagsA...>, TypeSeq<TagsB...>, TypeSeq<>>
{
    using type = TypeSeq<TagsA..., TagsB...>;
};

```

tpadioleau: Well i thought i would continue the pattern we already had for the other operations

EmilyBourne: Don't the other operations require recursive calculations though? I thought the extra class was included as a stopping condition.

E.g. `TypeSeqElement` which doesn't need recursion doesn't work like this:
https://github.com/CExA-project/ddc/blob/6d2a4938b84ceb08c674041e26c890e49de17e4d/include/ddc/detail/type_seq.hpp#L54-L61

tpadioleau: Ok if you think this adds more burden than it helps reading i will update the implementation

tpadioleau: done in 7759696cd4ff81e672c5774ad9eb524af66391ec

EmilyBourne (approved):

PR #795, opened 2025-03-05
Restrict TaggedVector operators

tpadioleau:


include/ddc/detail/tagged_vector.hpp
```diff
@@ -372,7 +372,9 @@ class TaggedVector : public TaggedVectorConversionOperators<TaggedVector<Element
         return *this;
     }
 
-    template <class OElementType>
+    template <
+            class OElementType,
+            class = std::enable_if_t<std::is_convertible_v<OElementType, ElementType>>>
```
tpadioleau, line 377: We were actually doing this check for the other operators. So in the end, this is just about aligning with the rest.

EmilyBourne (approved):

tpadioleau:
thanks @EmilyBourne

PR #819, opened 2025-04-09
Function to print ChunkSpan for debugging purpose

PaulGannay:


tests/print.cpp
```diff
@@ -0,0 +1,410 @@
...
+        std::stringstream ss;
+        ss << std::scientific << cells_in;
+        EXPECT_EQ(
+                "[[ 4.701857e-01 -8.370129e-01]\n"
+                " [ 4.398320e-01  3.475363e-01]]",
+                ss.str());
+    }
+}
+
+struct OutputChecker
```
PaulGannay, line 99: I tried to create a function that check that the output is correct, they are two problems:
 1 - It is not finished, it still fails to detect every possible wrong output.
 2 - I did it to avoid having to write the output we want in string and then compare to said string, but it became so complicated that it needs it's own set of tests.

I believe a better solution would be to use Gtest regex engine, and use a simpler 'OutputChecker' that only checks that '[' and ']' are correctly balanced.

tpadioleau: I like the tests in `Simple2DChunk`. I think if you can add a 3d ou 4d like this, it would be enough. In my opinion there is no need of generating random numbers.

include/ddc/print.hpp
tpadioleau, line 1: We have a `ddc::detail` namespace, can you put inside the private symbols ?

include/ddc/print.hpp
```diff
@@ -0,0 +1,314 @@
+// Copyright (C) The DDC development team, see COPYRIGHT.md file
+//
+// SPDX-License-Identifier: MIT
+
+#pragma once
+
+#include <sstream>
+
+#ifdef KOKKOS_COMPILER_GNU
```
tpadioleau, line 9:
```suggestion
#if defined(KOKKOS_COMPILER_GNU)
```

include/ddc/print.hpp
```diff
@@ -0,0 +1,314 @@
...
+                    ret = std::max(ret, get_element_width(s(i0)));
+                }
+            }
+        }
+
+        return ret;
+    }
+};
+
+#ifdef KOKKOS_COMPILER_GNU
```
tpadioleau, line 222:
```suggestion
#if defined(KOKKOS_COMPILER_GNU)
```

include/ddc/print.hpp
```diff
@@ -0,0 +1,314 @@
...
+                    ret = std::max(ret, get_element_width(s(i0)));
+                }
+            } else {
+                for (std::size_t i0 = 0; i0 < edgeitems; ++i0) {
+                    ret = std::max(ret, get_element_width(s(i0)));
+                }
+                for (std::size_t i0 = extent - edgeitems; i0 < extent; ++i0) {
+                    ret = std::max(ret, get_element_width(s(i0)));
+                }
+            }
```
tpadioleau, line 215: Can we simplify ?

tpadioleau (changes requested):
Looks good! Quick review relative to the coding style of DDC

include/ddc/print.hpp
```diff
@@ -0,0 +1,314 @@
...
+
+#ifdef KOKKOS_COMPILER_GNU
+
+template <class Type>
+void print_demangled_type_name(std::ostream& os)
+{
+    char* demangled_name;
+    int status;
+
+    demangled_name = abi::__cxa_demangle(typeid(Type).name(), nullptr, 0, &status);
```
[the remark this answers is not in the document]
PaulGannay: It solves a memory leak that could happen in case __cxa_demangle returns an error.

tests/print.cpp
```diff
@@ -0,0 +1,223 @@
...
+    ddc::Chunk cells_in_dev_alloc("cells_in_dev", domain_full, ddc::DeviceAllocator<cell>());
+    ddc::ChunkSpan const cells_in = cells_in_dev_alloc.span_view();
+
+    ddc::parallel_for_each(
+            domain_full,
+            KOKKOS_LAMBDA(ddc::DiscreteElement<> const i) { cells_in(i) = 0.12345; });
+
+    {
+        std::stringstream ss;
+        print_content(ss, cells_in);
```
PaulGannay, line 55: I only test that the content of the chunk is correctly displayed since the display of the metadata is compiler dependant.

include/ddc/print.hpp
tpadioleau: > Can we simplify ?

Misread the code, withdrawing my comment

include/ddc/print.hpp
```diff
@@ -0,0 +1,352 @@
+// Copyright (C) The DDC development team, see COPYRIGHT.md file
+//
+// SPDX-License-Identifier: MIT
+
+#pragma once
+
+#include <sstream>
```
tpadioleau, line 7: I saw these headers
```suggestion
#include <algorithm>
#include <cstddef>
#include <memory>
#include <ostream>
#include <sstream>
#include <type_traits>
#include <utility>
```

tests/print.cpp
```diff
@@ -0,0 +1,268 @@
+// Copyright (C) The DDC development team, see COPYRIGHT.md file
+//
+// SPDX-License-Identifier: MIT
+
+#include <fstream>
+#include <iomanip>
+#include <sstream>
+#include <string>
```
tpadioleau, line 8:
```suggestion
#include <ios>
#include <string>
#include <type_traits>
```

tests/print.cpp
tpadioleau: Fine for me

tests/print.cpp
```diff
@@ -0,0 +1,268 @@
...
+void PrintTestCheckOutput0d()
+{
+    ddc::DiscreteDomain<> const domain_full;
+
+    ddc::Chunk cells_in_dev_alloc("cells_in_dev", domain_full, ddc::DeviceAllocator<cell>());
+    ddc::ChunkSpan const cells_in = cells_in_dev_alloc.span_view();
+
+    ddc::parallel_for_each(
+            domain_full,
+            KOKKOS_LAMBDA(ddc::DiscreteElement<> const i) { cells_in(i) = 0.12345; });
```
tpadioleau, line 50:
```suggestion
    ddc::deepcopy(cells_in, 0.12345);
```

Should work

tests/print.cpp
```diff
@@ -0,0 +1,268 @@
...
+            = ddc::init_trivial_bounded_space(ddc::DiscreteVector<DDim1>(dim1));
+
+    ddc::DiscreteDomain<DDim0, DDim1> const domain_2d(domain_0, domain_1);
+
+    ddc::Chunk cells_in_dev_alloc("cells_in_dev", domain_2d, ddc::DeviceAllocator<cell>());
+    ddc::ChunkSpan const cells_in = cells_in_dev_alloc.span_view();
+
+    ddc::parallel_for_each(
+            domain_2d,
+            KOKKOS_LAMBDA(ddc::DiscreteElement<DDim0, DDim1> const i) { cells_in(i) = 0.12345; });
```
tpadioleau, line 83:
```suggestion
    ddc::deepcopy(cells_in, 0.12345);
```

tests/print.cpp
```diff
@@ -0,0 +1,268 @@
...
+
+    cells_in(ddc::DiscreteElement<DDim0, DDim1>(1, 1)) = -0.12345;
```
tpadioleau, line 85:
```suggestion
    cells_in(domain_2d.front() + ddc::DiscreteVector<DDim0, DDim1>(1, 1)) = -0.12345;
```

Note that it will fail on a GPU build.

tpadioleau (changes requested):

include/ddc/print.hpp
PaulGannay: Done

tests/print.cpp
PaulGannay: This was removed, now the values are set together in the parallel_for

tests/print.cpp
PaulGannay: done for the 2 test were it was usable.

tests/print.cpp
PaulGannay: done for the 2 test were it was usable.

include/ddc/print.hpp
```diff
@@ -0,0 +1,358 @@
...
+        ss << elem;
+        return ss.tellp();
+    }
+
+    template <class T>
+    void display_aligned_element(std::ostream& os, T& elem, std::size_t largest_element)
+    {
+        std::size_t const elem_width = get_element_width(elem);
+
+        for (int i = 0; i < largest_element - elem_width; ++i) {
```
tpadioleau, line 58:
```suggestion
        for (std::size_t i = 0; i < largest_element - elem_width; ++i) {
```

include/ddc/print.hpp
```diff
@@ -0,0 +1,358 @@
...
+    }
+
+    template <class ElementType, class Extents, class Layout, class Accessor, std::size_t... Is>
+    std::ostream& recursive_display(
+            std::ostream& os,
+            Kokkos::mdspan<ElementType, Extents, Layout, Accessor> const& s,
+            int level,
+            std::size_t largest_element,
+            std::size_t beginning,
+            std::size_t end)
```
tpadioleau, line 89: Can you pass a `std::integer_sequence` ? This should allow template deduction simplifying its usage.

include/ddc/print.hpp
```diff
@@ -0,0 +1,358 @@
...
+#if defined(KOKKOS_COMPILER_GNU) || defined(KOKKOS_COMPILER_CLANG)
+template <class Type>
+void print_demangled_type_name(std::ostream& os)
+{
+    int status;
+
+    std::unique_ptr<char, decltype(std::free)*> const demangled_name(
+            abi::__cxa_demangle(typeid(Type).name(), nullptr, nullptr, &status),
+            std::free);
+    if (status != 0) {
```
tpadioleau, line 259: Just to be sure, if `status != 0`, is calling `std::free` on the pointer ok ?

tests/print.cpp
tpadioleau, line 1: Can you remove `typename cell`, I guess it comes from the game of life. Something more neutral would be better, `ElementType`, `T`.

include/ddc/print.hpp
```diff
@@ -0,0 +1,358 @@
...
+
+#if defined(KOKKOS_COMPILER_GNU) || defined(KOKKOS_COMPILER_CLANG)
+#    include <cxxabi.h>
+#endif
+
+namespace ddc {
+namespace detail {
+struct ChunkPrinter
+{
+    static int const threshold = 10;
```
tpadioleau, line 26:
```suggestion
    static constexpr int const s_threshold = 10;
```

include/ddc/print.hpp
```diff
@@ -0,0 +1,358 @@
...
+    // If this ever becomes modifiable by the user, we need to ensure that
+    // edgeitems < (threshold / 2) stays true.
+    static int const edgeitems = 3;
```
tpadioleau, line 29:
```suggestion
    static constexpr int const s_edgeitems = 3;
```

tpadioleau (changes requested):
Almost there :)

include/ddc/print.hpp
```diff
@@ -0,0 +1,358 @@
...
+
+#include "ddc/chunk_span.hpp"
+#include "ddc/discrete_domain.hpp"
+
+#if defined(KOKKOS_COMPILER_GNU) || defined(KOKKOS_COMPILER_CLANG)
+#    include <cxxabi.h>
+#endif
+
+namespace ddc {
+namespace detail {
```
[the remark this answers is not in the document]
PaulGannay: Not everything in the file is inside `ddc::detail`, the public API is only in `ddc`.
Do you want me to merge this line and add a new `namespace ddc` after it ?

include/ddc/print.hpp
PaulGannay: The pointer is supposed to be nullptr in this case, so yes.

tpadioleau: ok thanks!

include/ddc/print.hpp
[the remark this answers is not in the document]
tpadioleau: Good point, I think we can keep that way then.

tests/print.cpp
```diff
@@ -0,0 +1,279 @@
...
+    ddc::Chunk cells_in_dev_alloc("cells_in_dev", domain_2d, ddc::DeviceAllocator<cell>());
+    ddc::ChunkSpan const cells_in = cells_in_dev_alloc.span_view();
+
+    // Fill the array with 0.12345 in the cells that should be visible and -0.12345 in the one that will be eluded
+    // Check that the output is only aligned on 0.12345
+    ddc::parallel_for_each(
+            domain_2d,
+            KOKKOS_LAMBDA(ddc::DiscreteElement<DDim0, DDim1> const i) {
+                if ((i.uid<DDim0>() >= 3 && i.uid<DDim0>() < dim0 - 3)
+                    || (i.uid<DDim1>() >= 3 && i.uid<DDim1>() < dim1 - 3)) {
```
tpadioleau, line 169: We need to rework this part, we cannot rely on `uid`. I suggest to create two `DiscreteDomain` by calling `remove`. It takes two `DiscreteVector` `n0` and `n1` to remove the first `n0` elements and the last `n1` elements.

For example
```cpp
auto subdom_0 = domain_0.remove(ddc::DiscreteVector<Dim0>(3), ddc::DiscreteVector<Dim1>(3));
// ...
if(subdom_0.contains(ddc::DiscreteElement<Dim0>(i)))
// ...
```

PaulGannay: It works, thank you.

tpadioleau (approved):
Ok looks good to me!

tpadioleau:
Can you add your name to the AUTHORS file ?

tpadioleau (approved):

PR #827, opened 2025-04-16
Prefer stride(i) instead of stride_i()

tpadioleau:
I made a quick review, I think we don't have dangerous calls to `stride(i)`. We don't have functions that allow multiple ranks.


yasahi-hpc (approved):
LGTM. 
Regarding internal `Gbtrs`, do you have a plan to suppress this in the future? All the serial solvers are available after `KK 4.7`.

tpadioleau:
> LGTM. Regarding internal `Gbtrs`, do you have a plan to suppress this in the future? All the serial solvers are available after `KK 4.7`.

I think last time we agreed on waiting the release that will have all these solvers. So if 4.7 has them all, that is fine to me.

yasahi-hpc:
Sounds good

PR #829, opened 2025-04-17
Add coverage test

tpadioleau:


tpadioleau:
@EmilyBourne Any idea how to suppress the `tests` directory from the report ? https://app.codecov.io/gh/CExA-project/ddc/pull/829/tree

EmilyBourne:
> @EmilyBourne Any idea how to suppress the `tests` directory from the report ? https://app.codecov.io/gh/CExA-project/ddc/pull/829/tree

The gcov flag --filter should do this. Maybe check your regex?

tpadioleau:
> > @EmilyBourne Any idea how to suppress the `tests` directory from the report ? https://app.codecov.io/gh/CExA-project/ddc/pull/829/tree
> 
> The gcovr flag --filter should do this. Maybe check your regex?

That is also what I thought. I feel the paths are not correct due to my usage of the volume mounting with docker

tpadioleau:
@EmilyBourne It seems that the codecov action was also executing gcov as it could find coverage files in the build. Since I delete the build it seems much better! I guess you did not see this as you called this action from a new job ?

tpadioleau:
@EmilyBourne Also consider `--exclude-throw-branches`, in my case this revealed that a lot of "potentially throwing" code was not tested. We may not be interested in error handling for now. It gets much greener after adding this option.

tpadioleau:
Some observations:
- not tested template code is not counted
- header files not included do not count

This makes the coverage higher than expected.

tpadioleau:
It seems I need to run `gcov` from the container to make it work with Ginkgo that is only available from the container. I guess it is also more coherent with regard to the version of gcc used for compiling.

.github/workflows/tests-ubuntu.yaml
```diff
@@ -116,8 +116,86 @@ jobs:
...
+            /root/.local/bin/gcovr \
+              --exclude-throw-branches \
+              --filter include/ddc \
+              --lcov coverage.lcov \
+              --merge-mode-functions=merge-use-line-0 \
+              build
+          EOF
+
+          docker run \
+            --user 0 \
```
tpadioleau, line 185: @EmilyBourne it ends like this :D

PR #838, opened 2025-04-24
Add a contributing guide

tpadioleau:


CONTRIBUTING.md
```diff
@@ -0,0 +1,76 @@
...
+To format a C++ file, please apply the command
+
+```bash
+clang-format -i the-cpp-file
+```
+
+One can find the formatting style in the file `.clang-format`.
+
+> [!WARNING]
+> The formatting might not give the same result with different versions of a tool.
```
tretre91, line 72: It might be a good idea to specify the versions of clang-format and gersemi that are used, this avoids having to search through the github workflow files

tpadioleau: I agree. The problem is that the version used comes from the CI and I don't know how to keep the documentation synchronized with it ? I could add a link to the page that mentions the version ?

tretre91: Yeah it would be difficult (impossible?) to sync the docs and the workflow files
We can keep it as is then, if someones has format-related CI failures and need the versions they may just ask or search themselves

tpadioleau: Yes we don't provide easy tools to help with coding style so I guess maintainers have to help with this in a PR.

PR #853, opened 2025-05-16
Remove DiscreteDomain undefined behavior

tpadioleau:
This PR removes an undefined behavior when accessing container's data from a `ddc::DiscreteElement` due to a shift of the pointer out of bounds of the memory allocation.

This is achieved by removing the specialization of `ddc::Chunk` and `ddc::ChunkSpan` for `ddc::DiscreteDomain`. The slicing operator taking a `ddc::DiscreteDomain` for `ddc::ChunkSpan` is transferred to the common implementation but enabled only when `SupportType` is a `ddc::DiscreteDomain`.


tretre91 (approved):
Looks good to me (I'm not _that_ familiar with this part of the code, but I see nothing shocking)

tpadioleau:
> Looks good to me (I'm not _that_ familiar with this part of the code, but I see nothing shocking)

Thank you

PR #860, opened 2025-05-23
Add assertions when slicing

tpadioleau:


include/ddc/chunk_span.hpp
```diff
@@ -267,6 +267,7 @@ class ChunkSpan : public ChunkCommon<ElementType, SupportType, LayoutStridedPoli
     KOKKOS_FUNCTION constexpr auto operator[](
             DiscreteElement<QueryDDims...> const& slice_spec) const
     {
+        assert(select<QueryDDims...>(this->m_domain).contains(slice_spec));
```
yasahi-hpc, line 270: Compared to the check below, why empty slice_spec is not allowed?

tests/chunk.cpp
```diff
@@ -520,6 +520,21 @@ TEST(Chunk2DTest, SliceCoordY)
...
+TEST(Chunk2DTest, SliceCoordXOutOfBounds)
+{
+#if !defined(NDEBUG) // The assertion is only checked if NDEBUG isn't defined
+    ChunkXY<double> chunk(dom_x_y);
+
+    char const* const death_msg
+            = R"rgx([Aa]ssert.*select<QueryDDims...>\(this->m_domain\).contains\(slice_spec\))rgx";
+
+    EXPECT_DEATH(chunk[dom_x.front() - 1], death_msg);
+    EXPECT_DEATH(chunk[dom_x.back() + 1], death_msg);
```
yasahi-hpc, line 532: In case of array out of bounds, is there a way to identify which dimension involves out of bounds access?

tests/chunk.cpp
```diff
@@ -572,6 +587,21 @@ TEST(Chunk2DTest, SliceDomainY)
...
+TEST(Chunk2DTest, SliceDomainXOutOfBounds)
+{
+#if !defined(NDEBUG) // The assertion is only checked if NDEBUG isn't defined
+    ChunkXY<double> chunk(dom_x_y.remove(DVectXY(1, 0), DVectXY(1, 0)));
+
+    char const* const death_msg
+            = R"rgx([Aa]ssert.*DiscreteDomain<QueryDDims...>\(this->m_domain\).contains\(odomain.front\(\)\) && DiscreteDomain<QueryDDims...>\(this->m_domain\).contains\(odomain.back\(\)\))rgx";
+
+    EXPECT_DEATH(chunk[dom_x_y.remove_last(DVectXY(1, 0))], death_msg);
+    EXPECT_DEATH(chunk[dom_x_y.remove_first(DVectXY(1, 0))], death_msg);
```
yasahi-hpc, line 599: Could you please illustrate or comment this example in detail?
For me, it is not clear what has been actually removed from each domain.
It would be useful if you describe

1. what is the original scope of `dom_x_y`
2. what is the scope of chunk after remove (l. 593)
3. what is the scope of last-removed `dom_x_y` (l. 598)
4. what is the scope of first-removed dom_x_y (l. 599)

yasahi-hpc (commented):
Some of the modifications are not very clear to me. 
If you can improve the error messages, it would be nicer.
Also comments in death-tests would be helpful.

include/ddc/chunk_span.hpp
yasahi-hpc: OK

tests/chunk.cpp
tpadioleau: thanks!

PR #877, opened 2025-06-22
Increase discrete_space.hpp coverage

tpadioleau:
Fixes red flags in https://app.codecov.io/gh/CExA-project/ddc/blob/main/include%2Fddc%2Fdiscrete_space.hpp


tretre91 (approved):
looks good to me

PR #878, opened 2025-06-23
Add spline compatibility check helper

yasahi-hpc:
This PR aims at a helper to check the consistency of spline builder and evaluator.
For some reason, I need double parentheses to use `ASSERT_FALSE` like 
`ASSERT_FALSE((ddc::is_spline_compatible_v<Builder2D_1, Builder2D_1>));`

### DIFF
--- a/include/ddc/kernels/splines/spline_traits.hpp
+++ b/include/ddc/kernels/splines/spline_traits.hpp
@@ -4,14 +4,7 @@
 
 #pragma once
 
-#include <cassert>
-#include <cstddef>
-#include <memory>
-#include <optional>
-#include <stdexcept>
-#include <tuple>
 #include <type_traits>
-#include <utility>
 
 #include <ddc/ddc.hpp>
 
@@ -21,6 +14,7 @@
 #include "spline_evaluator_2d.hpp"
 
 namespace ddc {
+
 template <class T>
 struct is_spline_builder : std::false_type
 {
@@ -53,7 +47,7 @@ template <class T>
 inline constexpr bool is_spline_builder_v = is_spline_builder<T>::value;
 
 template <class T>
-struct is_spline_builder2D : std::false_type
+struct is_spline_builder2d : std::false_type
 {
 };
 
@@ -69,7 +63,7 @@ template <
         ddc::BoundCond BcLower2,
         ddc::BoundCond BcUpper2,
         ddc::SplineSolver Solver>
-struct is_spline_builder2D<SplineBuilder2D<
+struct is_spline_builder2d<SplineBuilder2D<
         ExecSpace,
         MemorySpace,
         BSpline1,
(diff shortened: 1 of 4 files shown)


yasahi-hpc:
@tpadioleau 

Following changes are needed?

1. Only expose `is_<check>_v `
2. Add docs to `is_<check>_v `
3. Use parentheses to `&&`?

```C++
template <class Builder, class Evaluator>
 struct is_spline_compatible<Builder, Evaluator,
        std::enable_if_t<is_spline_builder_v<Builder> && is_spline_evaluator_v<Evaluator>
                         || is_spline_builder_v<Evaluator> && is_spline_evaluator_v<Builder> >>
```

Should be 

```C++
template <class Builder, class Evaluator>
 struct is_spline_compatible<Builder, Evaluator,
        std::enable_if_t<(is_spline_builder_v<Builder> && is_spline_evaluator_v<Evaluator>)
                         || (is_spline_builder_v<Evaluator> && is_spline_evaluator_v<Builder>) >>
```

include/ddc/kernels/splines/spline_traits.hpp
```diff
@@ -0,0 +1,119 @@
...
+struct is_spline_compatible<Builder, Evaluator,
+       std::enable_if_t<is_spline_builder_v<Builder> && is_spline_evaluator_v<Evaluator>
+                        || is_spline_builder_v<Evaluator> && is_spline_evaluator_v<Builder> >> {
+       static constexpr bool value = std::is_same_v<typename Builder::exec_space, typename Evaluator::exec_space> && 
+                                     std::is_same_v<typename Builder::memory_space, typename Evaluator::memory_space> &&
+                                     std::is_same_v<typename Builder::bsplines_type, typename Evaluator::bsplines_type>;
+};
+
+template <class Builder, class Evaluator>
+struct is_spline_compatible<Builder, Evaluator,
```
EmilyBourne, line 106: I'm not sure this name is very explicit. I would expect it to check if a spline is compatible with a builder or if a spline is compatible with an evaluator, not if a spline builder is compatible with an evaluator

I'm not sure what to suggest instead though
```suggestion
struct is_spline_interpolation_possible<Builder, Evaluator,
```
```suggestion
struct is_interpolation_possible<Builder, Evaluator,
```
?

yasahi-hpc: I do not have a strong preference.
But, I would prefer to include a term `spline` at least and ideally not too long

EmilyBourne: I definitely agree that it should not be too long.

I'm not sure it is important to include `spline` though? If you had a different intermediate representation it is likely that you would still need a similar check for a compatible Builder and Evaluator. That is why I suggested `is_interpolation_possible`

If splines are important then maybe `shares_spline_basis`? But this doesn't cover the Memory/Execution space aspects

yasahi-hpc:
@tpadioleau Can you format this
clangformat 20 is unavailable on my env

tests/splines/spline_traits.cpp
```diff
@@ -0,0 +1,294 @@
+// Copyright (C) The DDC development team, see COPYRIGHT.md file
+//
+// SPDX-License-Identifier: MIT
+
```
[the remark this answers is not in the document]
tpadioleau: thanks

PR #892, opened 2025-07-18
WIP: Use StridedDiscreteDomain with SplineBuilder

EmilyBourne:


tests/splines/batched_spline_builder.cpp
```diff
@@ -257,6 +269,25 @@ void BatchedSplineTest()
...
             std::optional(derivs_rhs.span_cview()));
+
+    {
+        ddc::Chunk coef_2_alloc(dom_spline, ddc::KokkosAllocator<double, MemorySpace>());
+        ddc::ChunkSpan const coef_2 = coef_2_alloc.span_view();
+        spline_builder(
+                coef_2,
+                vals.span_cview(),
+                std::optional(derivs_strided[interpolation_domain.front()].span_cview()),
+                std::optional(derivs_strided[interpolation_domain.back()].span_cview()));
```
tpadioleau, line 280: I think a way to properly do this would be to do a cartesian product between a 1D strided domain and 1D contiguous domain. Once you slice by a DiscreteElement in the strided dimension you would recover the contiguous domain over derivatives.

tpadioleau: We could allow strided domains in the spline operator but it feels artificial, the operator still requires all derivatives, so it would work only when the stride is unitary.

EmilyBourne: > We could allow strided domains in the spline operator but it feels artificial, the operator still requires all derivatives, so it would work only when the stride is unitary.

Yes. If a cross product of strided and unstrided domains was possible then there would be no need to allow strides domains here.

However I wonder if it is worth considering keeping this in mind for the handling of ND splines. If I had to pass `derivs_strided` here instead of `derivs_strided[front]` and `derivs_strided[back]` then it would be much easier to calculate the number of arguments required for the ND case (N! derivatives+ 1 argument for the values)

tpadioleau: > I guess it's not possible right now?

No indeed, but this is something to work on to be able to release v1 in my opinion.

> it would be much easier to calculate the number of arguments required for the ND case (N! derivatives+ 1 argument for the values)

Do you claim that it is possible to pass all derivatives in one chunkspan ? Still we would need to check at runtime that the dimension of derivatives has unitary stride right ?

If we had this product feature, in 2D we could consider passing a single
```cpp
ChunkSpan<double, Product<StridedDiscreteDomain<DDimI1>, DiscreteDomain<Deriv<I1>>>> derivs_dim1;
```
instead of two arrays `derivs_min1` and `derivs_max1`. Do you think we can do better than this with only Chunk/ChunkSpan data structures ? This would also allow to check that the user passes derivatives at the correct location ?

EmilyBourne: The formula for the number of derivatives required is:
$$\sum_{i=1}^{N} \binom{N}{i}$$
if we pass all derivatives of the same type in one chunkspan.
If we pass explicitly as we have done up to now the formula is:
$$\sum_{i=1}^{N} 2^i \binom{N}{i}$$

N | N derivs with strided | N derivs without strided
--|-----------------------|--------------------------
1 | 1 | 2
2 | 3 | 8
3 | 7 | 26
4 | 15 | 80
5 | 31 | 242

tpadioleau: > > Still we would need to check at runtime that the dimension of derivatives has unitary stride right ?
> 
> Yes probably. This could be in an assert I guess?

Yes we can but my point is to say that we need to develop this product feature. We cannot propagate too many unnecessary `StridedDiscreteDomain`, there could be a performance penalty.

tpadioleau:
Closing as it is now out of date.

PR #893, opened 2025-07-22
Fix slicing operator

tpadioleau:


tpadioleau:
> A test should be added. This is important because the test will probably show that this change is not sufficient. In #892 I additionally had to add strided_discrete_domain to the includes

True, I fixed it. It seems that the compiler did not encounter the correct overload of `select` before `operator[]`. I think there are two solutions:
- either using ADL with `using ddc::select` inside the function
- either use the `Rebind` type alias that is now part of the mandatory API of domains.

EmilyBourne (approved):

PR #915, opened 2025-09-14
Use CMake file sets to install headers

tpadioleau:


CMakeLists.txt
```diff
@@ -205,14 +249,49 @@ if("${DDC_BUILD_KERNELS_SPLINES}")
...
+            FILE_SET HEADERS
+                BASE_DIRS include
+                FILES
+                    include/ddc/kernels/splines.hpp
+                    include/ddc/kernels/splines/bsplines_non_uniform.hpp
+                    include/ddc/kernels/splines/bsplines_uniform.hpp
+                    include/ddc/kernels/splines/constant_extrapolation_rule.hpp
+                    include/ddc/kernels/splines/deriv.hpp
+                    include/ddc/kernels/splines/greville_interpolation_points.hpp
+                    include/ddc/kernels/splines/integrals.hpp
```
tretre91, line 269: Should this also include the files in `include/ddc/splines/kokkos-kernels-ext`?

tretre91: Yep, they should be included https://github.com/CExA-project/ddc/actions/runs/17852348534/job/50763540054?pr=915#step:22:99

tpadioleau: done in 587891e1ee43174334b82fbe0458b81474507d14

install_test/main.cpp
```diff
@@ -12,14 +12,20 @@ static_assert(std::is_same_v<ddc::Real, float>);
 
 #if !__has_include(<ddc/kernels/fft.hpp>)
 #    error
+#else
```
science-enthusiast, line 15: These are not critical. We can address them later if needed.

Earlier we just wanted to error out, now we want to include the file if it is available. The following order is simpler to read:

```
#if __has_include(<ddc/kernels/fft.hpp>)
#    include <ddc/kernels/fft.hpp>
#else
#    error
#endif
```
Also, an error message could be added.

PR #944, opened 2025-10-14
Use `StridedDiscreteDomain`s to pass the derivatives to the spline builders

tretre91:
This PR modifies the SplineBuilder interface to pass `StridedDiscreteDomain`s for the derivatives, which allows to pack some of the arguments and reduce the rate at which the number of arguments grows with the number of dimensions.

This is based on #929


tpadioleau:
Closing as it is now out of date.

PR #972, opened 2025-11-18
Enforce that the dimensions passed to discrete element, vector and domain are unique

tretre91:
Currently, there is no check to see if the user passes the same dimension multiple times in the template arguments of `DiscreteElement`, `DiscreteVector` and `[Sparse,Strided]DiscreteDomain`. `DiscreteElement<DDimX, DDimX, DDimX> delem;` compiles without issues.

This PR adds a static assert in these classes to check that the dimensions are unique.

### DIFF
--- a/include/ddc/discrete_domain.hpp
+++ b/include/ddc/discrete_domain.hpp
@@ -64,9 +64,7 @@ class DiscreteDomain
     friend class DiscreteDomain;
 
     static_assert(
-            ((type_seq_size_v<type_seq_remove_t<detail::TypeSeq<DDims...>, detail::TypeSeq<DDims>>>
-              == sizeof...(DDims) - 1)
-             && ...),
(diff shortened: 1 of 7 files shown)


include/ddc/discrete_domain.hpp
```diff
@@ -63,6 +63,12 @@ class DiscreteDomain
     template <class...>
     friend class DiscreteDomain;
 
+    static_assert(
+            ((type_seq_size_v<type_seq_remove_t<detail::TypeSeq<DDims...>, detail::TypeSeq<DDims>>>
+              == sizeof...(DDims) - 1)
+             && ...),
+            "The dimensions of a DiscreteDomain must be unique");
```
tpadioleau: I think this is a good idea, could you make it a helper trait ? like `detail::type_seq_unique_v<detail::TypeSeq<DDims...>>` ?

PR #973, opened 2025-11-19
Add higher order derivatives support to the evaluators

### LINKED ISSUE #926: Add the possibility to compute 2nd derivatives or higher with the spline evaluators
Currently, the `SplineEvaluator` and `SplineEvaluator2D` operators offers methods to compute the derivative along the first dimension, (the second dimension and the cross-derivatives for 2D). However, to compute cross-derivatives involving a change of coordinates, we need the second derivatives in the two dimensions (2D case). 
E.g. for a mapping $\mathcal{F}: (u,v) \mapsto (x,y)$ and a given function $f$, 
```math
\begin{aligned}
\partial_{xy} f & = \partial_{xy}u \partial_u f +  \partial_{xy}v \partial_v f \\
                         & + \partial_x u \partial_y u \partial_{u^2} f  \\
                         & + \partial_x v \partial_y v \partial_{v^2} f  \\
                         & + (\partial_x u \partial_y v + \partial_x v \partial_y u) \partial_{uv} f
\end{aligned}
```
* $\partial_{xy}u, \partial_{xy}v, \partial_x u,  \partial_y u ,  \partial_x v,$ and $\partial_y v$ should be provided by the Jacobian and Hessian of the mapping, 
* $\partial_u f , \partial_v f$ and $\partial_{uv} f$  are already provided by the spline evaluator 2D, 
* only remain $\partial_{u^2} f$ and $\partial_{v^2} f$ . 

I noticed that the computation of $n$-th derivatives is already possible in the definition of the B-splines: https://github.com/CExA-project/ddc/blob/c8aac4840109ceeede757d6f01090a3e30403338/include/ddc/kernels/splines/bsplines_non_uniform.hpp#L551C34-L551C57

I think that in https://github.com/CExA-project/ddc/blob/main/include/ddc/kernels/splines/spline_evaluator_2d.hpp the main changes are to create another tag, e.g.: 
```cpp
    struct eval_deriv2_type
    {
    };
``` 

and later in the code https://github.com/CExA-project/ddc/blob/main/include/ddc/kernels/splines/spline_evaluator_2d.hpp#L1271, add the new option, e.g.: 
```cpp
    template <class EvalType1, class EvalType2, class Layout, class... CoordsDims>
    KOKKOS_INLINE_FUNCTION double eval_no_bc(
            ddc::Coordinate<CoordsDims...> const& coord_eval,
            ddc::ChunkSpan<double const, spline_domain_type, Layout, memory_space> const
                    spline_coef) const
    {
        static_assert(
                std::is_same_v<EvalType1, eval_type> || std::is_same_v<EvalType1, eval_deriv_type>);
        static_assert(
                std::is_same_v<EvalType2, eval_type> || std::is_same_v<EvalType2, eval_deriv_type>);
        ddc::DiscreteElement<bsplines_type1> jmin1;
        ddc::DiscreteElement<bsplines_type2> jmin2;

        std::array<double, bsplines_type1::degree() + 1> vals1_ptr;
        Kokkos::mdspan<double, Kokkos::extents<std::size_t, bsplines_type1::degree() + 1>> const
                vals1(vals1_ptr.data());
        std::array<double, bsplines_type2::degree() + 1> vals2_ptr;
        Kokkos::mdspan<double, Kokkos::extents<std::size_t, bsplines_type2::degree() + 1>> const
                vals2(vals2_ptr.data());

        std::array<double, 3 * (bsplines_1::degree() + 1)> derivs1_ptr;
        ddc::DSpan2D const derivs1(derivs1_ptr.data(), bsplines_1::degree() + 1, 3);
        std::array<double, 3 * (bsplines_2::degree() + 1)> derivs2_ptr;
        ddc::DSpan2D const derivs2(derivs2_ptr.data(), bsplines_2::degree() + 1, 3);

        ddc::Coordinate<continuous_dimension_type1> const coord_eval_interest1(coord_eval);
        ddc::Coordinate<continuous_dimension_type2> const coord_eval_interest2(coord_eval);

        if constexpr (std::is_same_v<EvalType1, eval_type>) {
            jmin1 = ddc::discrete_space<bsplines_type1>().eval_basis(vals1, coord_eval_interest1);
        } else if constexpr (std::is_same_v<EvalType1, eval_deriv_type>) {
            jmin1 = ddc::discrete_space<bsplines_type1>().eval_deriv(vals1, coord_eval_interest1);
        } else if constexpr (std::is_same_v<EvalType1, eval_deriv2_type>) {
            jmin1 = ddc::discrete_space<bsplines_1>()
                            .eval_basis_and_n_derivs(derivs1, coord_eval_interest1, 2);
            std::size_t two = 2;
            for (std::size_t i = 0; i < bsplines_1::degree() + 1; ++i) {
                vals1[i] = derivs1(i, two);
            }
        }
        if constexpr (std::is_same_v<EvalType2, eval_type>) {
            jmin2 = ddc::discrete_space<bsplines_type2>().eval_basis(vals2, coord_eval_interest2);
        } else if constexpr (std::is_same_v<EvalType2, eval_deriv_type>) {
            jmin2 = ddc::discrete_space<bsplines_type2>().eval_deriv(vals2, coord_eval_interest2);
        } else if constexpr (std::is_same_v<EvalType2, eval_deriv2_type>) {
            jmin2 = ddc::discrete_space<bsplines_2>()
                            .eval_basis_and_n_derivs(derivs2, coord_eval_interest2, 2);
            std::size_t two = 2;
            for (std::size_t i = 0; i < bsplines_2::degree() + 1; ++i) {
                vals2[i] = derivs2(i, two);
            }
        }

        double y = 0.0;
        for (std::size_t i = 0; i < bsplines_type1::degree() + 1; ++i) {
            for (std::size_t j = 0; j < bsplines_type2::degree() + 1; ++j) {
                y += spline_coef(
                             ddc::DiscreteElement<
                                     bsplines_type1,
                                     bsplines_type2>(jmin1 + i, jmin2 + j))
                     * vals1[i] * vals2[j];
            }
        }
        return y;
    }
```

The addition of public method such as `deriv_dim_1`, `deriv_dim_2` etc should be pretty similar for the second derivatives by changing the tag `eval_deriv_type'  by `eval_deriv2_type'. Suggestion of name `second_deriv_dim_1`, `second_deriv_dim_2`. But maybe as it seems to be general for the B-splines, it may be possible to generalise it to the evaluator too, so put the order as an input?

tretre91:
This PR adds the ability to compute derivatives higher than the first order using the `SplineEvaluator` classes.
The `deriv`, `deriv_dim_1`, `deriv_1_and_2`, `deriv_1_2_3`, etc. methods are deprecated in favor of a `deriv` method taking as first argument a `DiscreteElement` defined on the dimensions of interest whose components are the desired orders of derivation along each of the dimension.

For example, for a 3D evaluator with `continuous_dimension_type`s `I1`, `I2` and `I3`
```cpp
spline_evaluator.deriv(ddc::DiscreteElement<ddc::Deriv<I1>, ddc::Deriv<I3>>(2, 1), ...)
```
will use the second derivative on the first dimension, the first derivative on the third dimension, and the values (no derivation) on the second dimension because the DiscreteElement does not contain a component in `Deriv<I2>`.

For practical reasons, the 1D evaluator can go up to the first derivative, the 2D evaluator up to the second derivative and the 3D evaluator up to the third derivative.

Note that I still need to create some tests for the higher order derivatives, the current tests only go up to cross-differentiation with the first derivative on all dimensions.

This should resolve #926

### DIFF
--- a/tests/splines/spline_error_bounds.hpp
+++ b/tests/splines/spline_error_bounds.hpp
@@ -6,6 +6,7 @@
 
 #include <algorithm>
 #include <array>
+#include <utility>
 
 #include <ddc/kernels/splines.hpp>
 
--- a/tests/splines/1d_spline_evaluator_derivatives.cpp
+++ b/tests/splines/1d_spline_evaluator_derivatives.cpp
@@ -4,9 +4,6 @@
 
 #include <algorithm>
 #include <cstddef>
-#if defined(BSPLINES_TYPE_UNIFORM)
-#    include <type_traits>
-#endif
 #include <vector>
 
 #include <ddc/ddc.hpp>
--- a/tests/splines/2d_spline_evaluator_derivatives.cpp
+++ b/tests/splines/2d_spline_evaluator_derivatives.cpp
@@ -4,9 +4,6 @@
 
 #include <algorithm>
 #include <cstddef>
-#if defined(BSPLINES_TYPE_UNIFORM)
-#    include <type_traits>
-#endif
 #include <vector>
 
 #include <ddc/ddc.hpp>
--- a/tests/splines/3d_spline_evaluator_derivatives.cpp
+++ b/tests/splines/3d_spline_evaluator_derivatives.cpp
@@ -4,9 +4,6 @@
 
 #include <algorithm>
 #include <cstddef>
-#if defined(BSPLINES_TYPE_UNIFORM)
-#    include <type_traits>
-#endif
 #include <vector>
 
 #include <ddc/ddc.hpp>


include/ddc/kernels/splines/spline_evaluator_2d.hpp
```diff
@@ -1289,15 +1378,54 @@ class SplineEvaluator2D
...
+                        Kokkos::extents<std::size_t, bsplines_type1::degree() + 1, 3>> const
+                        derivs1(derivs1_ptr.data());
+
+                jmin1 = ddc::discrete_space<bsplines_type1>()
+                                .eval_basis_and_n_derivs(derivs1, coord_eval_interest1, order1);
+
+                for (std::size_t i = 0; i < bsplines_type1::degree() + 1; ++i) {
+                    vals1[i] = DDC_MDSPAN_ACCESS_OP(derivs1, i, order1);
+                }
+            }
```
tpadioleau: I would be in favor of only keeping the cases `order1 > 0 && order1 < 3`. The case `order1 == 0`, can be represented by not adding the dimension in the discrete element. I wonder if the case `order1 == 1` is necessary, I guess we can consider it an optimization path ?

@EmilyBourne Any opinion ?

include/ddc/kernels/splines/spline_evaluator.hpp
```diff
@@ -557,27 +656,41 @@ class SplineEvaluator
...
                 vals(vals_ptr.data());
         ddc::Coordinate<continuous_dimension_type> const coord_eval_interest(coord_eval);
-        if constexpr (std::is_same_v<EvalType, eval_type>) {
+
+        if constexpr (sizeof...(DDims) == 0) {
             jmin = ddc::discrete_space<bsplines_type>().eval_basis(vals, coord_eval_interest);
-        } else if constexpr (std::is_same_v<EvalType, eval_deriv_type>) {
-            jmin = ddc::discrete_space<bsplines_type>().eval_deriv(vals, coord_eval_interest);
+        } else {
+            if (order == 0) {
```
tpadioleau: Can you move the definition of `order` in this `else` branch ? This would remove the call to `select_or`.

tests/splines/spline_error_bounds.hpp
tpadioleau: This class seems independent of the naming of dimensions, could you use `std::array` instead of `ddc::DiscreteElement` to represent the order of derivatives ?

tpadioleau (changes requested):
Ok looks pretty good, just a few comments/questions.

tpadioleau (commented):
Seems good to me.

Can you say a word on the tests for the derivatives ? Are they extracted from the `*_spline_builder.cpp` files ?

I think we will need to do a pass to update the included headers. Unless you prefer to do it before merging, this can be done after.

tretre91:
> Can you say a word on the tests for the derivatives ? Are they extracted from the *_spline_builder.cpp files ?

Yes, they are extracted from the `batched[_2d,3d]_spline_builder.cpp` files, I only kept the periodic case and removed everything related to greville or hermite bc since the derivatives don't use the boundary conditions.
I also factorized the tests of the different derivatives into one function (`test_deriv`) so that I can generate all the combinations of derivatives at compile time, note that in the assertion for the norm, I kept the `1e-11 * max_norm_diff` but in my tests it was always smaller than the other value (`error_bounds.error_bound(...)`)
```cpp
EXPECT_LE(
        max_norm_error_diff,
        std::
                max(error_bounds.error_bound(
                            std::array<ddc::DiscreteElementType, 1> {order},
                            {dx<I>(ncells)},
                            {s_degree_x}),
                    1e-11 * max_norm_diff));
```

> I think we will need to do a pass to update the included headers. Unless you prefer to do it before merging, this can be done after.

I'd prefer doing it before merging, I just took a look at the files and afaict only the `<type_traits>` header is not used, did you notice any other ones that should be removed/updated?

tpadioleau:
> I'd prefer doing it before merging, I just took a look at the files and afaict only the `<type_traits>` header is not used, did you notice any other ones that should be removed/updated?

Indeed and `tests/splines/spline_error_bounds.hpp` is missing `<utility>` (for `std::index_sequence`). That is all I could spot.

PR #1002, opened 2025-12-16
Use static_cast

tpadioleau:
- fixes `google-readability-casting` by adding C++ `static_cast`
- fixes `modernize-use-scoped-lock` (due to the upgrade to clang-tidy-21 in 1e862873e851ad41ad9c8029062fae5af551767f)
- ignores a `misc-const-correctness` related to PDI
- fixes `misc-confusable-identifiers` by renameing some variables in `tests/splines/cosine_evaluator.hpp` 
- other `misc-*` did not trigger anything
- triggers the Ubuntu CI when `.clang-tidy` changes

### DIFF
--- a/benchmarks/deepcopy.cpp
+++ b/benchmarks/deepcopy.cpp
@@ -57,8 +57,8 @@ void memcpy_1d(benchmark::State& state)
         std::memcpy(dst_data.data(), src_data.data(), dst_data.size() * sizeof(double));
     }
     state.SetBytesProcessed(
-            static_cast<int64_t>(state.iterations())
-            * static_cast<int64_t>(state.range(0) * sizeof(double)));
+            static_cast<std::int64_t>(state.iterations())
+            * static_cast<std::int64_t>(state.range(0) * sizeof(double)));
 }
 
 void deepcopy_1d(benchmark::State& state)
@@ -72,8 +72,8 @@ void deepcopy_1d(benchmark::State& state)
         ddc::parallel_deepcopy(dst, src);
     }
     state.SetBytesProcessed(
-            static_cast<int64_t>(state.iterations())
-            * static_cast<int64_t>(state.range(0) * sizeof(double)));
+            static_cast<std::int64_t>(state.iterations())
+            * static_cast<std::int64_t>(state.range(0) * sizeof(double)));
 }
 
 void memcpy_2d(benchmark::State& state)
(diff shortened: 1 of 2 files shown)


tests/splines/cosine_evaluator.hpp
```diff
@@ -23,18 +23,18 @@ struct CosineEvaluator
...
+        double m_coef0;
 
-        double m_c1;
+        double m_coef1;
 
     public:
         template <class Domain>
-        explicit Evaluator([[maybe_unused]] Domain domain) : m_c0(1.0)
-                                                           , m_c1(0.0)
+        explicit Evaluator([[maybe_unused]] Domain domain) : m_coef0(1.0)
```
yasahi-hpc:
```suggestion
        explicit Evaluator(Domain /*domain*/) : m_coef0(1.0)
```

I do not remember exactly, but is there a clang-tidy check for unused variable?

benchmarks/deepcopy.cpp
```diff
@@ -56,7 +56,9 @@ void memcpy_1d(benchmark::State& state)
     for (auto _ : state) {
         std::memcpy(dst_data.data(), src_data.data(), dst_data.size() * sizeof(double));
     }
-    state.SetBytesProcessed(int64_t(state.iterations()) * int64_t(state.range(0) * sizeof(double)));
+    state.SetBytesProcessed(
+            static_cast<int64_t>(state.iterations())
+            * static_cast<int64_t>(state.range(0) * sizeof(double)));
```
yasahi-hpc:
```suggestion
            static_cast<std::int64_t>(state.iterations())
            * static_cast<std::int64_t>(state.range(0) * sizeof(double)));
```

I think it is better to use `std::int64_t` for consistency

yasahi-hpc (changes requested):
Mostly looks good.
I am very surprised why `std::int64_t(3)` is not allowed, but I could not find a reasonable workaround.
I personally prefer localized suppression for this, but it is up to you.

benchmarks/splines.cpp
```diff
@@ -229,29 +231,29 @@ void characteristics_advection(benchmark::State& state)
     std::int64_t const non_uniform = 1;
     // Preallocate 12 unitary benchmarks for each combination of cpu/gpu execution space, uniform/non-uniform and spline degree we may want to benchmark (those are determined at compile-time, that's why we need to build explicitly 12 variants of the bench even if we call only one of them)
     std::map<std::array<std::int64_t, 3>, std::function<void(benchmark::State&)>> benchmarks;
-    benchmarks[std::array {host, uniform, std::int64_t(3)}]
+    benchmarks[std::array {host, uniform, static_cast<std::int64_t>(3)}]
```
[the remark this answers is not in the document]
tpadioleau: I found in https://google.github.io/styleguide/cppguide.html#Casting

> Do not use cast formats like (int)x unless the cast is to void. You may use cast formats like T(x) only when T is a class type.

In both cases it is C cast.

yasahi-hpc (approved):
LGTM

PR #1013, opened 2025-12-29
Use a commented name instead of maybe_unused

tpadioleau:


tpadioleau:
@yasahi-hpc unfortunately this clashes with doxygen. Doxygen cannot find the commented parameter. I think it is also related to the fact we only have declarations that directly define functions.

yasahi-hpc:
> @yasahi-hpc unfortunately this clashes with doxygen. Doxygen cannot find the commented parameter. I think it is also related to the fact we only have declarations that directly define functions.

That is unfortunate. 
May be we need to define separately?

I actually found that a commented name is already used in DDC.
Do they cause any problems?

https://github.com/CExA-project/ddc/blob/f5deb6cb75f77e8932155b611e1bed7aa9223799/include/ddc/discrete_domain.hpp#L416-#L420

yasahi-hpc:
I have made a look at the source codes and documentations.
In my opinion, it is better not to document these methods of `DiscreteDomain` since the class is just a specialization for the empty case.

Is it possible to remove the docstrings from the methods in specialized class and explain the behaviors (if the behavior is different for the empty case) in the methods in the general `DiscreteDomain` class.

yasahi-hpc:
This is the `std::array` style. We need a specialization for rank 0, but we do not need separate documentations for that.

tpadioleau:
> I think it is better to remove the case `DiscreteDomain<>` from the documentation. You cannot remove that because of some restriction of doxygen?

It is a different class, it contains different member functions. Of course it is close to the general case but still different. Even if I remove it, it does not solve the problem for other classes like `src/ddc/kernels/splines/constant_extrapolation_rule.hpp` that also has an unused parameter that should be documented.

yasahi-hpc (approved):
Other than newly added comments, looks good to me

tpadioleau:
> LGTM

Thanks I will merge as soon as the CI passes

PR #1022, opened 2026-01-06
Introduce parallel_transform algorithm

tpadioleau:


include/ddc/kernels/fft.hpp
```diff
@@ -248,7 +242,7 @@ void impl(
             norm_coef = (backward_full_norm_coef(DiscreteDomain<DDimOut>(ddom_out)) * ...);
         }
 
-        rescale(exec_space, out, static_cast<real_type_t<Tout>>(norm_coef));
+        ddc::parallel_transform(exec_space, out, ScaleFn<real_type_t<Tout>>(norm_coef));
```
EmilyBourne: What is the benefit of a class over a lambda here?

tpadioleau: I don't trust nvcc on lambdas with multiple variadic template parameters :D

PR #1050, opened 2026-02-05
Implement a target for the main of the tests

tpadioleau:


tests/CMakeLists.txt
```diff
@@ -6,9 +6,13 @@ include(GoogleTest)
 
 add_subdirectory(discrete_space)
 
+add_library(ddc_gtest_main)
```
tretre91: Don't know if it would make much of a difference, but this could be an object library since it's only used to be linked against other targets within the project

tpadioleau: I am not confident in this feature, they seem more complicated to use https://cmake.org/cmake/help/latest/command/add_library.html#object-libraries and I don't know how it behaves with Kokkos / CUDA

PR #1087, opened 2026-03-20
Add save_npy function

tpadioleau:
Add the possibility to save a `Kokkos::mdspan` in a Numpy file format (version 1.0). It is only implemented for layouts left and right, and for some scalar types.

It should be reviewed with this compatibility table in mind https://numpy.org/doc/stable/user/basics.types.html#relationship-between-numpy-data-types-and-c-data-types. Note that I find surprising that fixed width numpy types correspond to non-fixed width types in C.

We could also check  that floating point types are IEEE754 compliant with https://en.cppreference.com/cpp/types/numeric_limits/is_iec559.

### DIFF
--- a/src/ddc/save_npy.cpp
+++ b/src/ddc/save_npy.cpp
@@ -1,0 +1,107 @@
+// Copyright (C) The DDC development team, see COPYRIGHT.md file
+//
+// SPDX-License-Identifier: MIT
+
+#include <bit>
+#include <cstddef>
+#include <cstdint>
+#include <filesystem>
+#include <fstream>
+#include <numeric>
+#include <stdexcept>
+#include <string>
+#include <utility>
+#include <vector>
+
+#include "save_npy.hpp"
+
+namespace ddc::detail {
+
+NpyByteOrder get_byte_order(std::size_t const itemsize) noexcept
+{
+    if (itemsize == 1) {
+        return NpyByteOrder::not_applicable;
+    }
+
+    if (std::endian::native == std::endian::little) {
+        return NpyByteOrder::little_endian;
+    }
+
+    if (std::endian::native == std::endian::big) {
+        return NpyByteOrder::big_endian;
+    }
+
+    return NpyByteOrder::not_applicable;
+}
+
+void write_le(std::ostream& os, std::uint16_t const value_u16)
+{
+    constexpr unsigned int mask = 0xFFU;
+
+    unsigned int const value_u = value_u16;
+
+    std::array<char, 2> bytes;
+    bytes[0] = value_u & mask;
+    bytes[1] = (value_u >> 8U) & mask;
+
+    os.write(bytes.data(), sizeof(value_u16));
+}
+
+std::string NpyDtype::str() const
+{
+    return std::string(1, static_cast<char>(byte_order)) + static_cast<char>(kind)
+           + std::to_string(itemsize);
+}
+
+// See specification at https://numpy.org/neps/nep-0001-npy-format.html#format-specification-version-1-0
+void save_npy(std::ostream& os, NpyArrayView const& view)
+{
+    // Build shape string: (d0, d1, ..., dN,)
+    std::string shape_str = "(";
+    for (std::size_t const ext : view.shape) {
+        shape_str += std::to_string(ext);
+        shape_str += ", ";
+    }
+    shape_str += ")";
+
+    std::string const header_dict
+            = std::string("{'descr': '") + view.dtype.str() + "', 'fortran_order': "
+              + (view.fortran_order ? "True" : "False") + ", 'shape': " + shape_str + ", }";
+
+    // Pad header to a multiple of 16
+    std::size_t const non_padded_header_len = header_dict.size() + 1;
+    // magic(6) + major(1) + minor(1) + header_len(2) + header
+    std::size_t const padding = 16 - (6 + 1 + 1 + 2 + non_padded_header_len) % 16;
+    if (!std::in_range<std::uint16_t>(non_padded_header_len + padding)) {
+        throw std::runtime_error("save_npy: header too large for npy v1.0.");
+    }
+    auto const header_len = static_cast<std::uint16_t>(non_padded_header_len + padding);
+
+    // magic string
+    os.write("\x93NUMPY", 6);
+    // major version
+    os.put(1);
+    // minor version
+    os.put(0);
+    // header length in little-endian
+    write_le(os, header_len);
+    // header + padding + newline
+    os.write(header_dict.data(), header_dict.size());
+    os.write("               ", padding);
+    os.put('\n');
+
+    // Raw data
+    std::size_t const n_elems
+            = std::accumulate(view.shape.begin(), view.shape.end(), 1ULL, std::multiplies<> {});
+    os.write(reinterpret_cast<char const*>(view.data), n_elems * view.dtype.itemsize);
+}
+
+void save_npy(std::filesystem::path const& filename, NpyArrayView const& view)
+{
+    std::ofstream file(filename, std::ios::binary);
+    file.exceptions(std::ios::failbit | std::ios::badbit);
+
+    save_npy(file, view);
+}
+
+} // namespace ddc::detail
--- a/src/ddc/save_npy.cpp
+++ b/src/ddc/save_npy.cpp
@@ -73,7 +73,7 @@ void save_npy(std::ostream& os, NpyArrayView const& view)
     // magic(6) + major(1) + minor(1) + header_len(2) + header
     std::size_t const alignment = 16;
     std::size_t const remainder = (6 + 1 + 1 + 2 + non_padded_header_len) % alignment;
-    std::size_t const padding = remainder == 0 ? 0 : alignment - remainder;
+    std::size_t const padding = (alignment - remainder) % alignment;
     if (!std::in_range<std::uint16_t>(non_padded_header_len + padding)) {
         throw std::runtime_error("save_npy: header too large for npy v1.0.");
     }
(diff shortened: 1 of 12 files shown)


src/ddc/save_npy.cpp
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    return NpyByteOrder::not_applicable;
+}
+
+void write_le(std::ostream& os, std::uint16_t const value_u16)
+{
+    constexpr unsigned int mask = 0xFFU;
+
+    unsigned int const value_u = value_u16;
```
PaulGannay, line 41: Why don't you define mask as an `uint16_t`, and why use the intermediate variable `value_u`?

src/ddc/save_npy.hpp
```diff
@@ -0,0 +1,173 @@
...
+
+    ddc::detail::save_npy(
+            filename,
+            ddc::detail::NpyArrayView {
+                    .data = mds.data_handle(),
+                    .dtype = ddc::detail::convert_to_npy_dtype<std::remove_const_t<T>>(),
+                    .shape = std::move(shape),
+                    .fortran_order = std::is_same_v<Layout, Kokkos::layout_left>,
+            });
+}
```
PaulGannay, line 152: There is a lot of code duplication between the two `save_npy` function.
Maybe the code in common could be put in a new `check_precondition` function, or the code to open the file could be move inside the `.hpp` and this function could call the other directly?

PaulGannay (changes requested):

src/ddc/save_npy.cpp
tpadioleau: Originally it was `std::uint16_t` but in the operation `value_u & mask` it was promoted to an `int` and then the linter complained that we should not use signed integers for binary operations

src/ddc/save_npy.hpp
```diff
@@ -0,0 +1,173 @@
...
+
+    if constexpr (std::is_same_v<T, bool>) {
+        // ── Single-byte / untyped ─────────────────────────────────────────
+        kind = NpyKind::boolean;
+    } else if constexpr (std::is_same_v<T, std::byte>) {
+        // std::byte → raw byte buffer, no arithmetic meaning
+        kind = NpyKind::other;
+    } else if constexpr (std::is_same_v<T, char>) {
+        // char is a distinct type; its signedness is implementation-defined
+        kind = std::is_signed_v<char> ? NpyKind::signed_int : NpyKind::unsigned_int;
```
tretre91: I don't think we need a special case for `char`, it would be handled by the `is_signed<T>/is_unsigned<T>` branches
```suggestion
```

src/ddc/save_npy.cpp
```diff
@@ -0,0 +1,107 @@
...
+    shape_str += ")";
+
+    std::string const header_dict
+            = std::string("{'descr': '") + view.dtype.str() + "', 'fortran_order': "
+              + (view.fortran_order ? "True" : "False") + ", 'shape': " + shape_str + ", }";
+
+    // Pad header to a multiple of 16
+    std::size_t const non_padded_header_len = header_dict.size() + 1;
+    // magic(6) + major(1) + minor(1) + header_len(2) + header
+    std::size_t const padding = 16 - (6 + 1 + 1 + 2 + non_padded_header_len) % 16;
```
tretre91: If the header length is already a multiple of 16, you're going to add unnecessary spaces
```suggestion
    std::size_t const padding = (16 - (6 + 1 + 1 + 2 + non_padded_header_len) % 16) % 16;
```
Alternatively, you could keep this and add a space on line 90 so that the string is 16 characters long

tests/save_npy/generate_npy_fixtures.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+template <typename T>
+void save_array_3d(std::filesystem::path const& path, T value)
+{
+    Kokkos::View<T*, Kokkos::HostSpace> const alloc("", n);
+    Kokkos::deep_copy(alloc, value);
+    Kokkos::mdspan const view(alloc.data(), ns);
+
+    ddc::experimental::save_npy(path, view);
+}
```
tretre91, line 66: Is it intended that this is using a 1D view and mdspan instead of 3D?

src/ddc/save_npy.cpp
```diff
@@ -0,0 +1,107 @@
...
+    std::string shape_str = "(";
+    for (std::size_t const ext : view.shape) {
+        shape_str += std::to_string(ext);
+        shape_str += ", ";
+    }
+    shape_str += ")";
+
+    std::string const header_dict
+            = std::string("{'descr': '") + view.dtype.str() + "', 'fortran_order': "
+              + (view.fortran_order ? "True" : "False") + ", 'shape': " + shape_str + ", }";
```
tretre91, line 69: I don't think it's actually an issue, but according to the spec:
> For repeatability and readability, this dictionary is formatted using pprint.pformat() so the keys are in alphabetic order.

and `pformat` doesn't add trailing commas in tuples and dicts

src/ddc/save_npy.cpp
tpadioleau: Good catch, is the double modulo clear ? I can break the logic in two lines.

src/ddc/save_npy.hpp
tpadioleau: yeah i agree

tests/save_npy/generate_npy_fixtures.cpp
tpadioleau: No... but then why does it pass the test ?

tretre91: Oh I missread the `ns` for `n` in the mdspan constructor, the view is 1D but the mdspan is 3D so it's fine

src/ddc/save_npy.cpp
tretre91: Breaking it into two lines might make it less confusing yes

tests/save_npy/generate_npy_fixtures.cpp
tpadioleau: Well me too :)

src/ddc/save_npy.cpp
tpadioleau:
```
np.save("test.npy", np.ones((2, 3, 4), dtype=np.double))`
```
gives me 
```
{'descr': '<f8', 'fortran_order': False, 'shape': (2, 3, 4), }
```
Indeed no trailing comma for the shape but I see one for the dict. I am not sure we should change then.

tretre91: Interesting, we can keep it like this then, I guess they only care about the keys being in alphabetical order

src/ddc/save_npy.cpp
tpadioleau: done in f16864b65896c138b234b9cae18bb95ecf2ee0ea, let me know if this is not clear

src/ddc/save_npy.hpp
tpadioleau: done in a7bea8053cf9e9d9822ecc75dd3e6d64f4e10f3e

src/ddc/save_npy.cpp
```diff
@@ -0,0 +1,109 @@
...
+    if (std::endian::native == std::endian::little) {
+        return NpyByteOrder::little_endian;
+    }
+
+    if (std::endian::native == std::endian::big) {
+        return NpyByteOrder::big_endian;
+    }
+
+    return NpyByteOrder::not_applicable;
+}
```
tretre91, line 35: The function can be constexpr, the if conditions on the endianness can also be replaced by `if constexpr`s

src/ddc/save_npy.hpp
```diff
@@ -0,0 +1,152 @@
...
+    std::size_t itemsize; // in bytes
+
+    std::string str() const;
+};
+
+template <typename T>
+NpyDtype convert_to_npy_dtype()
+{
+    std::size_t const itemsize = sizeof(T);
+    NpyByteOrder const byte_order = get_byte_order(itemsize);
```
tretre91, line 46: If you make `get_byte_order` constexpr, you can replace this with
```suggestion
    constexpr std::size_t itemsize = sizeof(T);
    constexpr NpyByteOrder byte_order = get_byte_order(itemsize);
```

tretre91 (commented):
not blocking

src/ddc/save_npy.cpp
tpadioleau: I prefer compiling over constexpring :)

src/ddc/save_npy.hpp
```diff
@@ -0,0 +1,152 @@
...
+    NpyKind kind;
+
+    if constexpr (std::is_same_v<T, bool>) {
+        // ── Single-byte / untyped ─────────────────────────────────────────
+        kind = NpyKind::boolean;
+    } else if constexpr (std::is_same_v<T, std::byte>) {
+        // std::byte → raw byte buffer, no arithmetic meaning
+        kind = NpyKind::other;
+    } else if constexpr (
+            std::is_same_v<T, std::complex<float>> || std::is_same_v<T, std::complex<double>>) {
```
tretre91: Should there also be a check for `Kokkos::complex` here?

tpadioleau: Good question, maybe, they are both guaranteed to be of size 2 * sizeof(T) with T=float, double or long double ?

tpadioleau: Done

tretre91 (approved):

tpadioleau:
I should add the two new dependencies in the readme, after merging to avoid running the CI again.

PaulGannay (approved):
LGTM

tpadioleau:
Thank you both, merging as is. We will need to add support for layout stride later as we did for `ddc::print`.

PR #1115, opened 2026-04-11
Rework nvcc warnings

tpadioleau:
Revert #1112 in favor of using the flag `NVCC_PREPEND_FLAGS`, see https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/#nvcc-environment-variables


tpadioleau:
Working, see https://github.com/CExA-project/ddc/actions/runs/24280636571/job/70902010907

PR #1123, opened 2026-04-22
Fix non-uniform interpolation points with uniform splines

EmilyBourne:


tpadioleau (approved):
thanks looks ok to me

tpadioleau (approved):

PR #1135, opened 2026-05-06
Batched parallel copy

tpadioleau:


tretre91 (approved):
LGTM
 I had 2 remarks though:
- Should there also be an overload which doesn't take an execution space, like what is done with `parallel_deepcopy`?
- Also are you fine with the overlap in functionality between `parallel_deepcopy` and `parallel_copy` with ChunkSpans over the same domain? The only use case I can see for `parallel_deepcopy` over `parallel_copy` would be if you want to be absolutely sure that the copy is done over the same domain (and uses a `deep_copy`, even though this should be an implementation detail)

tpadioleau:
> * Should there also be an overload which doesn't take an execution space, like what is done with `parallel_deepcopy`?

Probably, I don't know how to handle this implementation:
- when calling `Kokkos::deep_copy` we cannot pass a string whereas it is possible with `Kokkos::parallel_for`.
- concerning the execution space behavior, both implementation are non-blocking, seems fine. But regarding the case without an execution space, `Kokkos::deep_copy` is fencing every stream if I remember correctly whereas the `Kokkos::parallel_for` implementation will still be non-blocking.

> * Also are you fine with the overlap in functionality between `parallel_deepcopy` and `parallel_copy` with ChunkSpans over the same domain? The only use case I can see for `parallel_deepcopy` over `parallel_copy` would be if you want to be absolutely sure that the copy is done over the same domain (and uses a `deep_copy`, even though this should be an implementation detail)

I am still thinking about this, I wanted `parallel_deepcopy` to be more oriented to transfers host/device. But clearly on a CPU build, the two algorithms are the same when input and output arrays have the same dimensions and layout. I have disabled `parallel_copy` in the case of different memory spaces because it would require an extra allocation to implement all its features.

tpadioleau:
> concerning the execution space behavior, both implementation are non-blocking, seems fine. But regarding the case without an execution space, Kokkos::deep_copy is fencing every stream if I remember correctly whereas the Kokkos::parallel_for implementation will still be non-blocking.

Regarding this point, I think I am in favor of having the semantic of a parallel_for, using `Kokkos::deep_copy` is just an optimization path.

tpadioleau:
> Regarding this point, I think I am in favor of having the semantic of a parallel_for, using Kokkos::deep_copy is just an optimization path.

What about this rule when one calls a DDC algorithm without explicit execution space:
- if the algorithm uses an output array, it will use the singleton of the execution space associated to the memory space of this array, for example `ChunkDst::memory_space::execution_space()`
- else if the algorithm uses an input array, it will use the singleton of the execution space associated to the memory space of this array, for example `ChunkSrc::memory_space::execution_space()`
- use the singleton of the default execution space, for example `Kokkos::DefaultExecutionSpace()`

tretre91:
> > Regarding this point, I think I am in favor of having the semantic of a parallel_for, using Kokkos::deep_copy is just an optimization path.
> 
> What about this rule when one calls a DDC algorithm without explicit execution space:
> 
>     * if the algorithm uses an output array, it will use the singleton of the execution space associated to the memory space of this array, for example `ChunkDst::memory_space::execution_space()`
> 
>     * else if the algorithm uses an input array, it will use the singleton of the execution space associated to the memory space of this array, for example `ChunkSrc::memory_space::execution_space()`
> 
>     * use the singleton of the default execution space, for example `Kokkos::DefaultExecutionSpace()`

Sounds good, if we go this way I think this behaviour should be documented somewhere, for example `CudaHostPinnedSpace` is device accessible but its associated execution space is `HostSpace::execution_space` (same for HIP and SYCL), this could be confusing for someone coming from Kokkos who would expect the algorithm to run on the default execution space

tpadioleau:
> Sounds good, if we go this way I think this behaviour should be documented somewhere, for example `CudaHostPinnedSpace` is device accessible but its associated execution space is `HostSpace::execution_space` (same for HIP and SYCL), this could be confusing for someone coming from Kokkos who would expect the algorithm to run on the default execution space

An alternative is to always pass an execution space, this is the choice of Kokkos experimental std algorithms

tpadioleau:
To illustrate the advantage of using an algorithm instead of a manual copy using `ddc::DiscreteElement` (executed on a Nvidia V100):
```
Kokkos::OpenMP::initialize WARNING: OMP_PROC_BIND environment variable not set
  In general, for best performance with OpenMP 4.0 or better set OMP_PROC_BIND=spread and OMP_PLACES=threads
  For best performance with OpenMP 3.1 set OMP_PROC_BIND=true
  For unit testing set OMP_PROC_BIND=false

2026-05-21T15:57:44+02:00
Running ./build/benchmarks/ddc_benchmark_parallel_copy
Run on (160 X 3900 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x80)
  L1 Instruction 32 KiB (x80)
  L2 Unified 1024 KiB (x80)
  L3 Unified 28160 KiB (x4)
Load Average: 18.93, 17.56, 11.25
***WARNING*** CPU scaling is enabled, the benchmark real time measurements may be noisy and will incur extra overhead.
***WARNING*** ASLR is enabled, the results may have unreproducible noise in them.
------------------------------------------------------------------------------------------------------------
Benchmark                                                  Time             CPU   Iterations UserCounters...
------------------------------------------------------------------------------------------------------------
benchmark_kokkos_deepcopy/2000                         41378 ns        41269 ns        16635 bytes_per_second=722.156Gi/s
benchmark_kokkos_deepcopy/20000                      3147687 ns      3140109 ns          220 bytes_per_second=949.086Gi/s
benchmark_kokkos_manual_copy/2000                     176308 ns       175858 ns         4001 bytes_per_second=169.468Gi/s
benchmark_kokkos_manual_copy/20000                  16767524 ns     16726954 ns           42 bytes_per_second=178.169Gi/s
benchmark_ddc_discrete_domain_manual_copy/2000        176197 ns       175783 ns         3998 bytes_per_second=169.54Gi/s
benchmark_ddc_discrete_domain_manual_copy/20000     16765583 ns     16729646 ns           42 bytes_per_second=178.141Gi/s
benchmark_ddc_strided_domain_manual_copy/2000         380289 ns       379477 ns         1853 bytes_per_second=78.5352Gi/s
benchmark_ddc_strided_domain_manual_copy/20000      48200340 ns     48092248 ns           15 bytes_per_second=61.9691Gi/s
benchmark_ddc_discrete_domain_parallel_copy/2000       41490 ns        41377 ns        17018 bytes_per_second=720.266Gi/s
benchmark_ddc_discrete_domain_parallel_copy/20000    3136688 ns      3128065 ns          219 bytes_per_second=952.74Gi/s
benchmark_ddc_strided_domain_parallel_copy/2000        41096 ns        40986 ns        16925 bytes_per_second=727.137Gi/s
benchmark_ddc_strided_domain_parallel_copy/20000     3132936 ns      3124908 ns          219 bytes_per_second=953.703Gi/s
```

PR #1151, opened 2026-05-23
Pre install googletest

tpadioleau:
This hopefully should fix an issue with the C language enabled in a subdirectory incompatible with the latest spurious update of PDI.


tpadioleau:
It is not enough, other libraries seem to require C language. Closing

PR #1164, opened 2026-06-04
Extend save_npy to layout stride and device memory

### LINKED ISSUE #1163: Support `Kokkos::layout_stride` in `save_npy`
Similar to `ddc::print`, see https://github.com/CExA-project/ddc/blob/fb6cc87eddc71c4ad0b6baeb9cb5a44ff9ad992a/src/ddc/print.hpp#L349-L382

tpadioleau:
`save_npy` now takes a `ChunkSpan` instead of `mdspan`. We now support device memory and layout stride by first making the span contiguous, if necessary, and copying the data to the host, if necessary.

Tests now use device memory and a layout stride test has been added.

The readme is updated to include new dependencies.


tretre91 (approved):
lgtm

PR #1169, opened 2026-06-08
Introduce experimental scan algorithms

tpadioleau:


tpadioleau:
Patch coverage is likely not maximal due to a single pass scan algorithm when executed in serial. It would require an openmp build to get a two pass scan to cover the false branching.

PR #1172, opened 2026-06-10
Coverage without inlining

tpadioleau:


tpadioleau:
does not help, closing

PR #1181, opened 2026-06-24
Fix spline evaluator

EmilyBourne:
Fix spline evaluator call to `operator()(spline_eval, spline_coef)` for ND evaluators (N>1). When the coefficients are not passed they are deduced on the fly. This local coordinate cannot be indexed.

Note that none of these operators seem to appear in the tests.

### DIFF
--- a/src/ddc/kernels/splines/spline_evaluator_2d.hpp
+++ b/src/ddc/kernels/splines/spline_evaluator_2d.hpp
@@ -449,7 +449,7 @@ public:
                         for (auto const i2 : evaluation_domain2) {
                             ddc::Coordinate<continuous_dimension_type1, continuous_dimension_type2>
                                     coord_eval_2D(ddc::coordinate(i1), ddc::coordinate(i2));
-                            spline_eval_2D(i1, i2) = eval(coord_eval_2D(i1, i2), spline_coef_2D);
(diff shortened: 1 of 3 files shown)


tpadioleau:
> Note that none of these operators seem to appear in the tests.

Urgh...

EmilyBourne:
> > Note that none of these operators seem to appear in the tests.
> 
> Urgh...

None of these files even

tpadioleau (approved):
Thank you for this fix

tpadioleau:
> > > Note that none of these operators seem to appear in the tests.
> > 
> > 
> > Urgh...
> 
> None of these files even

Not entirely true, look at https://app.codecov.io/gh/CExA-project/ddc/pull/1181/blob/src/ddc/kernels/splines/spline_evaluator_3d.hpp. We don't have proper testing of these files only but they are used in some tests.

EmilyBourne:
> > > > Note that none of these operators seem to appear in the tests.
> > > 
> > > 
> > > Urgh...
> > 
> > 
> > None of these files even
> 
> Not entirely true, look at https://app.codecov.io/gh/CExA-project/ddc/pull/1181/blob/src/ddc/kernels/splines/spline_evaluator_3d.hpp. We don't have proper testing of these files only but they are used in some tests.

Ah sorry, I grepped `spline_evaluator_3d.hpp` but the includes will just be `<ddc/kernels/splines.hpp>`

PR #1183, opened 2026-06-29
Fix: Safely return max() for missing tags in TypeSeq

Quntized:
If i use 
```bash
    struct DimX
    {
        static constexpr bool PERIODIC = true;
    };
    struct DimY
    {
        static constexpr bool PERIODIC = true;
    };
    struct DimZ {
        static constexpr bool PERIODIC = true;
    };
    struct DimW {};
    using SeqXY = ddc::detail::TypeSeq<DimX,DimY>;
    auto rank_size_3 = ddc::type_seq_rank_v<DimW,SeqXY>;
    std::cout<<rank_size_3<<std::endl;
```

Result i got:

```bash
1
```

### DIFF
--- a/tests/type_seq.cpp
+++ b/tests/type_seq.cpp
@@ -48,6 +48,8 @@ TEST(TypeSeqTest, Rank)
     EXPECT_EQ((ddc::type_seq_rank_v<T0, A>), 0);
     EXPECT_EQ((ddc::type_seq_rank_v<T1, A>), 1);
     EXPECT_EQ((ddc::type_seq_rank_v<T2, A>), 2);
+    constexpr std::size_t max_val = std::numeric_limits<std::size_t>::max();
+    EXPECT_EQ((ddc::type_seq_rank_v<T3, A>), max_val);
 }
 
 TEST(TypeSeqTest, Element)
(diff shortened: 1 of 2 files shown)


Quntized:
@tpadioleau can you check?

tests/type_seq.cpp
```diff
@@ -48,6 +48,8 @@ TEST(TypeSeqTest, Rank)
     EXPECT_EQ((ddc::type_seq_rank_v<T0, A>), 0);
     EXPECT_EQ((ddc::type_seq_rank_v<T1, A>), 1);
     EXPECT_EQ((ddc::type_seq_rank_v<T2, A>), 2);
+    constexpr std::size_t max_val = std::numeric_limits<std::size_t>::max();
+    EXPECT_EQ((ddc::type_seq_rank_v<T3, A>), max_val);
```
tpadioleau:
```suggestion
    EXPECT_EQ((ddc::type_seq_rank_v<T3, A>), (std::numeric_limits<std::size_t>::max()));
```

tpadioleau (changes requested):
I agree this is better, thx

tpadioleau (approved):

tpadioleau:
to be rebased on #1184

tpadioleau (approved):

Quntized:
> to be rebased on #1184

Thanks for the rebase

PR #1207, opened 2026-07-28
Support Kokkos-FFT v2

tpadioleau:


tpadioleau:
Note: We can either
- make the nightly work now
- wait for the latest image to be updated, it will require to wait spack packages v2026.09

tpadioleau:
The nightly is stopped for now.

PR #1210, opened 2026-08-01
Fix 2D extrapolation rule usage in SplineEvaluatorND

### LINKED ISSUE #1208: Cannot use a 2D `ConstantExtrapolationRule` with `SplineEvaluatorND`
The 2D version of the class `ConstantExtrapolationRule`:
https://github.com/CExA-project/ddc/blob/a1caf7e985ff62ff89bb36aa30b677f75b5c2a3a/src/ddc/kernels/splines/constant_extrapolation_rule.hpp#L75-L82
cannot be used with `SplineEvaluatorND` due to the following assertions:
https://github.com/CExA-project/ddc/blob/a1caf7e985ff62ff89bb36aa30b677f75b5c2a3a/src/ddc/kernels/splines/spline_evaluator_nd.hpp#L181-L206
which does not match the usage:
https://github.com/CExA-project/ddc/blob/a1caf7e985ff62ff89bb36aa30b677f75b5c2a3a/src/ddc/kernels/splines/spline_evaluator_nd.hpp#L668-L673

tpadioleau:
This aims to fix both:
- wrong static assertion
- implicit usage of the default constructor of the extrapolation rule (ddc::ConstantExtrapolationRule does not have a default constructor)

It also removes unnecessary variadic templates on the constructor of SplineEvaluatorND.

### DIFF
--- a/src/ddc/kernels/splines/spline_evaluator_nd.hpp
+++ b/src/ddc/kernels/splines/spline_evaluator_nd.hpp
@@ -152,6 +152,17 @@ private:
     cexa::tuple<ddc::type_seq_element_t<s_idx<BSplines>, upper_extrap_rule_ts>...>
             m_upper_extrap_rules;
 
+    /**
+     * @brief Build a SplineEvaluatorND acting on batched_spline_domain.
+     *
+     * @param extrap_rules The extrapolation rules at the lower then upper boundary, for each dimension.
+     */
+    explicit SplineEvaluatorND(cexa::tuple<ExtrapolationRule...> const& extrap_rules)
+        : m_lower_extrap_rules(cexa::get<2 * s_idx<BSplines>>(extrap_rules)...)
+        , m_upper_extrap_rules(cexa::get<2 * s_idx<BSplines> + 1>(extrap_rules)...)
+    {
+    }
+
 public:
     static_assert(
             sizeof...(BSplines) == dimension,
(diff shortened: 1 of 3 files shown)


tpadioleau:
@tretre91 could you review on monday ?

tretre91 (approved):
LGTM

tpadioleau:
We are loosing coverage on SplineEvaluator2D, is it a problem ? Maybe it could be time to rebase our 2D and 3D implementations on the ND implementation ?

tretre91:
> We are loosing coverage on SplineEvaluator2D, is it a problem ? Maybe it could be time to rebase our 2D and 3D implementations on the ND implementation ?

I would be fine if we just duplicated this test, but going forward we should consider using only the ND one.
I'm curious to know how much of a compile time increase we would get, if any

tpadioleau:
> > We are loosing coverage on SplineEvaluator2D, is it a problem ? Maybe it could be time to rebase our 2D and 3D implementations on the ND implementation ?
> 
> I would be fine if we just duplicated this test, but going forward we should consider using only the ND one.

Ok for me, do you feel like handling it ? If so you can open a PR on my branch.

> I'm curious to know how much of a compile time increase we would get, if any

I expect a speed up in the CI

tretre91:
> Ok for me, do you feel like handling it ? If so you can open a PR on my branch.

Sure, I will do it tomorrow

tretre91:
> > Ok for me, do you feel like handling it ? If so you can open a PR on my branch.
> 
> Sure, I will do it tomorrow

Opened #1213

tretre91 (approved):

PR #1227, opened 2026-08-27
Remove `Kokkos::Experimental` symbols

### LINKED ISSUE #1226: Do not use symbols from `Kokkos::Experimental `
Kokkos does not make any promise to symbols coming from `Kokkos::Experimental`, I suggest to remove all of them to avoid API breaking in a minor version. So far we have:
- `Kokkos::Experimental::layout_left_padded`
- `Kokkos::Experimental::layout_right_padded`
- `Kokkos::Experimental::count`
- `Kokkos::Experimental::begin`
- `Kokkos::Experimental::end`

tpadioleau:
As described in #1226 this PR removes all symbols from `Kokkos::Experimental`. It will help keeping a lower minimal version requirement of Kokkos in the future. Nevertheless DDC versions <0.16 will likely be incompatible with Kokkos >=5.4 (due padded layouts being out of the `Kokkos::Experimental` namespace).

### DIFF
--- a/src/ddc/chunk_span.hpp
+++ b/src/ddc/chunk_span.hpp
@@ -282,13 +282,22 @@ public:
         Slicer<to_type_seq_t<SupportType>> const slicer;
         auto subview = slicer(this->allocation_mdspan(), slice_spec);
         using layout_type = decltype(subview)::layout_type;
+        using mapping_type = decltype(subview)::mapping_type;
         using extents_type = decltype(subview)::extents_type;
         using OutTypeSeqDDims
                 = type_seq_remove_t<to_type_seq_t<SupportType>, TypeSeq<QueryDDims...>>;
         using OutDDom = detail::Rebind<SupportType, OutTypeSeqDDims>::type;
         if constexpr (
-                std::is_same_v<layout_type, Kokkos::Experimental::layout_left_padded<>>
-                || std::is_same_v<layout_type, Kokkos::Experimental::layout_right_padded<>>) {
+                std::is_same_v<layout_type, Kokkos::layout_left>
+                || std::is_same_v<layout_type, Kokkos::layout_right>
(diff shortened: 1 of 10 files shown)


src/ddc/chunk_span.hpp
```diff
@@ -282,13 +282,22 @@ class ChunkSpan : public ChunkCommon<ElementType, SupportType, LayoutStridedPoli
...
+                std::is_same_v<layout_type, Kokkos::layout_left>
+                || std::is_same_v<layout_type, Kokkos::layout_right>
+                || std::is_same_v<layout_type, Kokkos::layout_stride>) {
+            return ChunkSpan<
+                    ElementType,
+                    OutDDom,
+                    layout_type,
+                    memory_space>(subview, OutDDom(this->m_domain));
+        } else {
+            static_assert(mapping_type::is_always_strided() && mapping_type::is_always_unique());
```
tretre91, line 300: Why is this check needed? Looking at the mdspan source code, `is_always_strided` and `is_always_unique` are true for all 5 standard layouts

tretre91: Never mind, I read the associated issue, this is a requirement, not a way to only match `layout_[right|left]_padded`

tretre91 (approved):

src/ddc/chunk_span.hpp
tpadioleau: Maybe an overkill, writing this code I was wondering if starting with a mapping with some `always_*` (like `always_strided`) properties are preserved when applying submdspan ? Is it specified that the returned layout must preserve the same properties as the input layout ?

tpadioleau: Actually that is not possible, there is the property `is_always_exhaustive` that is `true` for `layout_left|right` but `false` for `layout_stride`.

PR #1232, opened 2026-09-08
Allow different precisions for spline calculations splines

### LINKED ISSUE #999: Allow different precisions for spline calculations
Currently spline calculations are hard-coded to use double precision. It would be nice to make this a template parameter so lower precision can be tested

EmilyBourne:
Allow different precisions for spline calculations splines. Fixes #999 

The tolerances used in the tests are chosen manually for double precision. They therefore do not pass currently when using single precision

### DIFF
--- a/src/ddc/kernels/splines/math_tools.hpp
+++ b/src/ddc/kernels/splines/math_tools.hpp
@@ -9,6 +9,8 @@
 
 #include <Kokkos_Core.hpp>
 
+#include <ddc/real_type.hpp>
+
 namespace ddc::detail {
 
 template <typename T>
@@ -59,18 +61,18 @@ KOKKOS_INLINE_FUNCTION T modulo(T x, T y)
     return x - y * Kokkos::floor(double(x) / y);
 }
 
-KOKKOS_INLINE_FUNCTION double ipow(double a, std::size_t i)
+KOKKOS_INLINE_FUNCTION Real ipow(Real a, std::size_t i)
 {
-    double r(1.0);
+    Real r(1.0);
     for (std::size_t j(0); j < i; ++j) {
         r *= a;
     }
     return r;
 }
 
-KOKKOS_INLINE_FUNCTION double ipow(double a, int i)
+KOKKOS_INLINE_FUNCTION Real ipow(Real a, int i)
 {
-    double r(1.0);
+    Real r(1.0);
     if (i > 0) {
         for (int j(0); j < i; ++j) {
             r *= a;
--- a/src/ddc/kernels/splines/math_tools.hpp
+++ b/src/ddc/kernels/splines/math_tools.hpp
@@ -7,9 +7,9 @@
 #include <array>
 #include <cstddef>
 
-#include <Kokkos_Core.hpp>
-
 #include <ddc/real_type.hpp>
+
+#include <Kokkos_Core.hpp>
 
 namespace ddc::detail {
 
--- a/src/ddc/kernels/splines/null_extrapolation_rule.hpp
+++ b/src/ddc/kernels/splines/null_extrapolation_rule.hpp
@@ -6,6 +6,8 @@
 
 #include <Kokkos_Macros.hpp>
 
+#include <ddc/real_type.hpp>
+
 namespace ddc {
 
 /**
@@ -16,10 +18,10 @@ struct NullExtrapolationRule
     /**
      * @brief Evaluates the spline at a coordinate outside of the domain.
      *
-     * @return A double with the value of the function outside the domain (here, 0.).
+     * @return A Real with the value of the function outside the domain (here, 0.).
      */
     template <class CoordType, class ChunkSpan>
-    KOKKOS_FUNCTION double operator()(CoordType, ChunkSpan) const
+    KOKKOS_FUNCTION Real operator()(CoordType, ChunkSpan) const
     {
         return 0.0;
     }
--- a/src/ddc/kernels/splines/null_extrapolation_rule.hpp
+++ b/src/ddc/kernels/splines/null_extrapolation_rule.hpp
@@ -4,9 +4,9 @@
 
 #pragma once
 
-#include <Kokkos_Macros.hpp>
-
 #include <ddc/real_type.hpp>
+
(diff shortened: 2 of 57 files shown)


src/ddc/kernels/splines/view.hpp
```diff
@@ -51,16 +53,16 @@ using View1D = ViewND<1, ElementType>;
 template <class ElementType>
 using View2D = ViewND<2, ElementType>;
 
-using DSpan1D = ddc::Span1D<double>;
+using DSpan1D = ddc::Span1D<Real>;
 
-using DSpan2D = ddc::Span2D<double>;
+using DSpan2D = ddc::Span2D<Real>;
```
EmilyBourne, line 58: These are the only aliases defined in this file that are actually used in the code. Should this file be removed in a future PR?

EmilyBourne: The objects in this file are also missing documentation, but this is a problem that predates this PR.

tpadioleau: Yeah could be indeed

src/ddc/kernels/splines/math_tools.hpp
```diff
@@ -7,6 +7,8 @@
 #include <array>
 #include <cstddef>
 
+#include <ddc/real_type.hpp>
+
```
tpadioleau, line 11:
```suggestion
#include <ddc/ddc.hpp>

```

src/ddc/kernels/splines/null_extrapolation_rule.hpp
```diff
@@ -4,6 +4,8 @@
 
 #pragma once
 
+#include <ddc/real_type.hpp>
```
tpadioleau, line 7:
```suggestion
#include <ddc/ddc.hpp>
```

src/ddc/kernels/splines/periodic_extrapolation_rule.hpp
```diff
@@ -4,18 +4,38 @@
 
 #pragma once
 
+#include <ddc/real_type.hpp>
```
tpadioleau, line 7:
```suggestion
#include <ddc/ddc.hpp>
```

src/ddc/kernels/splines/splines_linear_problem.hpp
```diff
@@ -7,6 +7,8 @@
 #include <cstddef>
 #include <iosfwd>
 
+#include <ddc/real_type.hpp>
```
tpadioleau, line 10:
```suggestion
#include <ddc/ddc.hpp>
```

src/ddc/kernels/splines/view.hpp
```diff
@@ -6,6 +6,8 @@
 
 #include <cstddef>
 
+#include <ddc/real_type.hpp>
```
tpadioleau, line 9:
```suggestion
#include <ddc/ddc.hpp>
```

src/ddc/kernels/splines/math_tools.hpp
```diff
@@ -59,18 +61,18 @@ KOKKOS_INLINE_FUNCTION T modulo(T x, T y)
     return x - y * Kokkos::floor(double(x) / y);
```
tpadioleau, line 61: Maybe also ?
```suggestion
    return x - y * Kokkos::floor(Real(x) / y);
```

tpadioleau (changes requested):
Looks globally ok to me, thanks for this contribution.

If I run
```console
grep -Rn "double" src/
src/ddc/save_npy.hpp:59:            std::is_same_v<T, std::complex<float>> || std::is_same_v<T, std::complex<double>>
src/ddc/save_npy.hpp:61:            || std::is_same_v<T, Kokkos::complex<double>>) {
src/ddc/save_npy.hpp:68:                !std::is_same_v<T, long double>,
src/ddc/save_npy.hpp:69:                "long double is platform-specific (80/96/128-bit); cast to double first.");
src/ddc/detail/kokkos.hpp:109:/// Only dynamic dimensions is supported for now i.e. `double[4]*` is not yet covered.
src/ddc/detail/kokkos.hpp:123:/// Only dynamic dimensions is supported for now i.e. `double[4]*` is not yet covered.
src/ddc/kokkos_allocator.hpp:22:                    <= std::max(alignof(std::max_align_t), alignof(Kokkos::complex<long double>)),
src/ddc/kernels/splines/math_tools.hpp:61:    return x - y * Kokkos::floor(double(x) / y);
src/ddc/kernels/fft.hpp:155:            std::is_same_v<real_type_t<Tin>, float> || std::is_same_v<real_type_t<Tin>, double>,
src/ddc/kernels/fft.hpp:156:            "Base type of Tin (and Tout) must be float or double.");
src/ddc/kernels/fft.hpp:159:            "Types Tin and Tout must be based on same type (float or double)");
src/ddc/kernels/fft.hpp:270:    double const lx = ddc::rlength(x_mesh);
src/ddc/kernels/fft.hpp:329: * @tparam Tin The type of the input elements (float, Kokkos::complex<float>, double or Kokkos::complex<double>).
src/ddc/kernels/fft.hpp:330: * @tparam Tout The type of the output elements (Kokkos::complex<float> or Kokkos::complex<double>).
src/ddc/kernels/fft.hpp:382: * @tparam Tin The type of the input elements (Kokkos::complex<float> or Kokkos::complex<double>).
src/ddc/kernels/fft.hpp:383: * @tparam Tout The type of the output elements (float, Kokkos::complex<float>, double or Kokkos::complex<double>).
src/ddc/real_type.hpp:13:using Real = double;
```

It seems to only remain one `double` in `math_tools.hpp` and also one in `fft.hpp` that should be `Real`. I can open a new PR for the fft. It remains an open question to me how to do proper mixing of different precisions as it could happen in the fft.

src/ddc/kernels/splines/math_tools.hpp
EmilyBourne: I wondered about this when I came across it. The only place it seems to be used is with `T = int` so I think casting to `Real` makes sense

tpadioleau (approved):
Ok for me!

PR #1233, opened 2026-09-09
699 create bsplines on a subset of a periodic domain

### LINKED ISSUE #699: Create bsplines on a subset of a periodic domain
B-splines are created as periodic or non-periodic objects, but the periodicity is collected from the dimension. This means that it is not possible to create a B-spline on a sub-section of the domain without creating a new dimension. It would be simpler if the periodicity was a template parameter of the bsplines (which could be the periodicity of the dimension by default).

I.e. I would like to be able to do:
```cpp
class DimPeriodic {
    static constexpr bool PERIODIC = true;
};

struct GridP : ddc::NonUniformPointSampling<DimPeriodic>
{
};

struct BSplinesP : UniformBSplines<DimPeriodic, 3>
{
};

void main() {
    ddc::Coordinate<DimPeriodic> p_min(0.0);
    ddc::Coordinate<DimPeriodic> p_max(2.0 * M_PI);
    std::size_t n_p(20);
    std::vector<ddc::Coordinate<DimPeriodic>> p_grid(n_p);
    p_grid[0] = p_min;
    for(std::size_t i(0); i < n_p-1; ++i)
        p_grid[i] = p_min + (p_max - p_min) / (n_p - 1);
    init_discrete_space<GridP>(p_grid);
    ddc::Coordinate<DimPeriodic> p_mid(M_PI);
    init_discrete_space<BSplinesP>(p_min, p_mid);
    ...
}
```

tpadioleau:


src/ddc/kernels/splines/constant_extrapolation_rule.hpp
tpadioleau: What should we do about this line ? I don't have yet access to the BSplines.

EmilyBourne: Why are the coordinates needed in the non-periodic case? Couldn't they be collected with `ddc::discrete_space<BSplinesNI>().rmin()` ?

tpadioleau: They are used to do a projection of the given coordinate in the interval `[min, max]`. I guess we could retrieve them through the global variable.

tpadioleau: I am trying to find usage examples but I feel it is not used anywhere ?

EmilyBourne: I don't know if there is a test with 2 non-periodic dimensions but I think this is a logical case to want to keep

tpadioleau: I agree, we deprecate the constructor with three parameters and use `rmin` and `rmax` ?

src/ddc/kernels/splines/bsplines_non_uniform.hpp
```diff
@@ -41,7 +41,7 @@ struct NonUniformBsplinesKnots : NonUniformPointSampling<typename T::continuous_
  * @tparam CDim The tag identifying the continuous dimension on which the support of the B-spline functions are defined.
  * @tparam D The degree of the B-splines.
  */
-template <class CDim, std::size_t D>
+template <class CDim, std::size_t D, bool Periodic = CDim::PERIODIC>
```
EmilyBourne, line 42: The only remaining places where `PERIODIC` is used in DDC are here and here:
https://github.com/CExA-project/ddc/blob/b55e285af639466f24a2a2af0edf029cf22f7c4a/src/ddc/kernels/splines/periodic_extrapolation_rule.hpp#L15

Is it worth keeping this concept? I don't think the assertion in `PeriodicExtrapolationRule` is necessary as there is always assertions on its usage, e.g.:
https://github.com/CExA-project/ddc/blob/b55e285af639466f24a2a2af0edf029cf22f7c4a/src/ddc/kernels/splines/spline_evaluator.hpp#L107-L115

The only advantage I can think of, is that while this property remains required, adding the template parameter remains backwards compatible.

tpadioleau: I can remove ` static_assert(DimI::PERIODIC, "PeriodicExtrapolationRule requires periodic dimension"); ` but I would still need a value for the extrapolation rules of the evaluator even if the BSplines is periodic. Do you suggest to just pass something like `void` ?

EmilyBourne: > but I would still need a value for the extrapolation rules of the evaluator even if the BSplines is periodic

I don't understand what you mean by this

tpadioleau: Ok maybe I did not undersand what you mean by
> Is it worth keeping this concept?

I though you were suggesting to remove the class `PeriodicExtrapolationRule`. So I was wondering what value we should put to have a "periodic behavior" in the template parameters of the evaluator in periodic BSplines.

EmilyBourne: No I was suggesting that you remove the need to write periodicity into the dimension tags:
```cpp
class X {
    static constexpr bool PERIODIC = false; // Is this line still useful
};
```

tpadioleau: I see, I am not sure how to replace such code in Gyselalib++ though:
```cpp
template <class IDim>
constexpr std::enable_if_t<!IDim::continuous_dimension_type::PERIODIC, double>
total_interval_length(IdxRange<IDim> const& idx_range)
{
    return std::fabs(ddc::rlength(idx_range));
}
```

EmilyBourne: In this case the constraint would be on the Gyselalib++ side. We have other requirements that come from Gyselalib++ but are not imposed by DDC (e.g. `IS_COVARIANT`).

My comment here is from a purely DDC point of view imagining other projects that may use the library

tpadioleau: Indeed, users can extend the constraints they put on the dimensions.

EmilyBourne (approved):
LGTM

PR #1234, opened 2026-09-10
Deprecate aliases in `view.hpp`

EmilyBourne:
As discussed in #1232 `view.hpp` contains lots of unused, undocumented aliases. This PR therefore deprecates the file and updates the few places where its aliases (only `DSpan1D` and `DSpan2D`) are used.


tpadioleau (dismissed):
Thanks looks good to me

tpadioleau (approved):

tpadioleau (approved):

PR #1236, opened 2026-09-16
Generalise ConstantExtrapolationRule implementation to ND

EmilyBourne:
Generalise `ConstantExtrapolationRule` implementation to ND. The implementation is heavily inspired by `SplineEvaluatorND::eval_no_bc`.


src/ddc/kernels/splines/constant_extrapolation_rule.hpp
```diff
@@ -137,60 +79,87 @@ struct ConstantExtrapolationRule<DimI, DimNI>
...
+        using BSplinesNI = ddc::type_seq_find_cdim_t<QDim, TypeSeqBSplines>;
+        if constexpr (BSplinesNI::is_periodic()) {
+            return ddc::Coordinate<QDim>(coord_extrap);
+        } else {
+            return Kokkos::
+                    clamp(ddc::Coordinate<QDim>(coord_extrap),
+                          ddc::discrete_space<BSplinesNI>().rmin(),
+                          ddc::discrete_space<BSplinesNI>().rmax());
+        }
+    }
```
tpadioleau:
```suggestion
    template <class BSplinesNI, class QDim>
    KOKKOS_INLINE_FUNCTION ddc::Coordinate<QDim> get_eval_pos(
            ddc::Coordinate<QDim> coord_extrap) const
    {
        static_assert(ddc::in_tags_v<QDim, ddc::detail::TypeSeq<DimNI...>>);
        if constexpr (BSplinesNI::is_periodic()) {
            return ddc::Coordinate<QDim>(coord_extrap);
        } else {
            return Kokkos::
                    clamp(ddc::Coordinate<QDim>(coord_extrap),
                          ddc::discrete_space<BSplinesNI>().rmin(),
                          ddc::discrete_space<BSplinesNI>().rmax());
        }
    }
```

That would avoid to define `type_seq_find_cdim_t` which I think could not be part of the other type_seq meta functions as it is too specific to "continuous dimensions", TypeSeq should rather be for any type

src/ddc/kernels/splines/constant_extrapolation_rule.hpp
```diff
@@ -137,60 +79,84 @@ struct ConstantExtrapolationRule<DimI, DimNI>
...
+                                    vals)[idx[ddc::type_seq_rank_v<BSplines, TypeSeqBSplines>]]
+                            * ...);
+                });
 
         return y;
     }
+
+private:
+    template <class TypeSeqBSplines, class QDim>
+    KOKKOS_INLINE_FUNCTION ddc::Coordinate<QDim> get_eval_pos(
```
tpadioleau:
```suggestion
    KOKKOS_FUNCTION ddc::Coordinate<QDim> get_eval_pos(
```

Being pedantic :)

src/ddc/kernels/splines/constant_extrapolation_rule.hpp
```diff
@@ -137,60 +79,84 @@ struct ConstantExtrapolationRule<DimI, DimNI>
...
+        } else {
+            return Kokkos::
+                    clamp(ddc::Coordinate<QDim>(coord_extrap),
+                          ddc::discrete_space<BSplinesNI>().rmin(),
+                          ddc::discrete_space<BSplinesNI>().rmax());
+        }
+    }
+
+    template <std::size_t N, class Functor, class... Is>
+    KOKKOS_INLINE_FUNCTION static void for_each(
```
tpadioleau:
```suggestion
    KOKKOS_FUNCTION static void for_each(
```

src/ddc/kernels/splines/constant_extrapolation_rule.hpp
```diff
@@ -137,60 +79,84 @@ struct ConstantExtrapolationRule<DimI, DimNI>
...
-                            clamp(ddc::Coordinate<DimNI>(coord_extrap),
-                                  ddc::discrete_space<bsplines_ni_type>().rmin(),
-                                  ddc::discrete_space<bsplines_ni_type>().rmax()));
-        }
+        static_assert(in_tags_v<DimI, to_type_seq_t<CoordType>>);
+        static_assert((in_tags_v<DimNI, to_type_seq_t<CoordType>> && ...));
+        using TypeSeqBSplines = ddc::detail::TypeSeq<BSplines...>;
+
+        ddc::Coordinate<DimI, DimNI...> const
+        coord_eval(m_eval_pos, get_eval_pos<TypeSeqBSplines>(ddc::select<DimNI>(coord_extrap))...);
```
tpadioleau:
```suggestion
        coord_eval(m_eval_pos, get_eval_pos<BSplines>(ddc::select<BSplines::continuous_dimension_type>(coord_extrap))...);
```
Should work with the change below ?

tpadioleau (changes requested):
Thanks for the generalization. If possible I would prefer we don't provide `type_seq_find_cdim_t`, it is not as generic as the other helpers as it relies on `continuous_dimension_type`.

If I am not mistaken I think we could actually get rid of the template `DimNI...` because it was only necessary for the deprecated constructor.

src/ddc/kernels/splines/constant_extrapolation_rule.hpp
EmilyBourne: I inlined here to match the implementation here:
https://github.com/CExA-project/ddc/blob/f22ae0470fae902a4e12e98b8e4907578ab7d5e4/src/ddc/kernels/splines/spline_evaluator_nd.hpp#L742

Why don't you want to inline this?

tpadioleau: Functions defined inside the body of the class are actually `inline` by default. I must have missed it in the evaluator.

EmilyBourne: Done: 5caf00c2

src/ddc/kernels/splines/constant_extrapolation_rule.hpp
EmilyBourne: Done: 5caf00c2

tpadioleau (approved):
Awesome, generalisation and simplification! thanks!

PR #1238, opened 2026-09-17
Add helper to combine tagged types

### LINKED ISSUE #864: Provide a method to combine types such as DiscreteDomains
It is quite common to want to combine 2 existing `DiscreteDomain` types into an extended `DiscreteDomain`. E.g. from `DiscreteDomain<GridBatch1, GridBatch2>` and `DiscreteDomain<OperatorDim1, OperatorDim2>` I would like to obtain a `DiscreteDomain<GridBatch1, GridBatch2, OperatorDim1, OperatorDim2>`. Currently to achieve this we need to transit via `TypeSeq` objects:
```cpp
using BatchedDomain = ddc::detail::convert_type_seq_to_discrete_domain_t<type_seq_cat_t<ddc::to_type_seq_t<BatchDomain>, ddc::to_type_seq_t<OperatorDomain>>>;
```

It would be nice to have an operator which can do this. This could be implemented as above or more generally 

<details>

<summary>Example implementation</summary>

```cpp
template <class... Containers>
struct Combine;

template <class Container>
struct Combine<Container>
{
    using type = Container;
};

template <
        template <typename...>
        class Container,
        class... Tags,
        class... OTags,
        class... TailContainers>
struct Combine<Container<Tags...>, Container<OTags...>, TailContainers...>
{
    using type = Combine<Container<Tags..., OTags...>, TailContainers...>::type;
};

/**
 * @brief A helper structure to determine the type when combining tags across containers.
 *
 * E.g. combine_t<Idx<Tag1>, Idx<Tag2>> == Idx<Tag1, Tag2>
 *
 * @tparam Containers The containers that should be combined. They should differ only in the tags.
 */
template <class... Containers>
using combine_t = typename detail::Combine<Containers...>::type;
```

</details>

The advantage of the general method is that it can also be used for `DiscreteElement` and `DiscreteVector`. It is rare to need to combine these without also needing to combine a `DiscreteDomain` (from which the element and vector can be extracted) but this does happen.

tpadioleau:


tests/CMakeLists.txt
```diff
@@ -49,6 +49,7 @@ add_executable(
     save_npy.cpp
     sparse_discrete_domain.cpp
     strided_discrete_domain.cpp
+    tagged_type_traits.cpp
```
EmilyBourne, line 52: Did you forget to commit this file?

tpadioleau: Yup :)

EmilyBourne (dismissed):

EmilyBourne (approved):

